PR #115, opened 2017-07-18
CSS Example Display Tweaks

stephaniehobson:
- Match media query width to where kuma flips display
  - approx 738px
- Selected example looks more editable, selected has arrow
- Output: space example from edges, prevent overflow.
- Vertically centre reset button
- Equalize space between example-choices
- Tweaks to header
    - less contrast
    - language tweak (only one examples can be selected at once)
    - remove unused top and left declarations
    - fix typo
- Extend grey background to cover entire page

### DIFF
--- a/css/editable-css.css
+++ b/css/editable-css.css
@@ -162,8 +162,11 @@ pre[class*='language-'] {
     outline: 0 solid transparent;
 }
 
-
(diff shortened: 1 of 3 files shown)


schalkneethling:
r+

PR #121, opened 2017-07-18
Adds status border colors back to CSS examples

schalkneethling:
Not sure if these were intentionally left of in the latest revision but, I do feel they still offer some additional benefit. @stephaniehobson

### DIFF
--- a/css/editable-css.css
+++ b/css/editable-css.css
@@ -89,6 +89,7 @@ pre[class*='language-'] {
 
 .live .example-choice {
     background-color: #e0e0e0;
+    border-left: 10px solid #9d9d9c;
 }
 
 .example-choice-list .example-choice:first-child {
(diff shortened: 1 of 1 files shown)


stephaniehobson:
They were intentionally left off because the arrow symbol changes to communicate the invalid status and the border interferes with the "editable" looking form field. If the arrow symbol is not enough of a signal I think we should explore another way to communicate status.

schalkneethling:
@stephaniehobson Thought as much, but wanted to be sure ;) I used examples with it in the Google doc. I reckon we can add another scenario to demonstrate how things would be without. I will update the doc, and close this.

stephaniehobson:
Maybe it would be good to have a section in the (really great) overview document you put together mentioning stuff we specifically would like to improve.

PR #225, opened 2017-08-22
added ArrayBuffer examples

maboa:
I added some ArrayBuffer examples.

Notes:

1. Sticking to one console.log per example for now. Happy to change though.

2. I still need to add attributes in the code tags (ie class, data-line and data-char) I based my examples on array-concat.html (which doesn't have them) so just noticed that now.

3. I included a constructor example - not sure if this is needed (also difficult to know what to console.log without using byteLength which was used in the next example).

4. Wasn't quite sure what to do for get ArrayBuffer[@@species] so left out for now.

5. ArrayBuffer.transfer() not currently working. Not sure if it is supported.


live-examples/js-examples/arraybuffer-slice.html
```diff
@@ -0,0 +1,21 @@
...
+<code id="static-js">// create an ArrayBuffer with a size in bytes
+  var buffer = new ArrayBuffer(16);
+
+  var int32View = new Int32Array(buffer); // create the view
+
+  for (var i = 0; i < int32View.length; i++) {
+    int32View[i] = i;
+  }
+
+  // 0,1,2,3
```
wbamberg: This comment was unclear to me

live-examples/js-examples/arraybuffer-slice.html
```diff
@@ -0,0 +1,21 @@
+<pre>
+<code id="static-js">// create an ArrayBuffer with a size in bytes
+  var buffer = new ArrayBuffer(16);
+
+  var int32View = new Int32Array(buffer); // create the view
+
+  for (var i = 0; i < int32View.length; i++) {
+    int32View[i] = i;
+  }
```
wbamberg: I don't have any bright ideas here really, but is there a more concise/simple way we can initialise this array? It feels to me like this is quite extraneous to the example...

live-examples/js-examples/arraybuffer-constructor.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">// create an ArrayBuffer with a size in bytes (in this case 8)
+var buffer = new ArrayBuffer(8);
+
+console.log(buffer);
+
+// expected output: [object ArrayBuffer]
+</code>
```
wbamberg: I think this example is worth having. It shows us a valid argument to pass. Logging byteLength would be fine here too I think.

schalkneethling:
> I still need to add attributes in the code tags (ie class, data-line and data-char) I based my examples on array-concat.html (which doesn't have them) so just noticed that now.

@maboa This is no longer needed. This was used before when we were using CodeMirror. I need to update the README to reflect this.

site.json
```diff
@@ -178,6 +178,41 @@
...
+            "type": "js"
+        },
+        "arraybufferSlice": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/arraybuffer-slice.html",
+            "fileName": "arraybuffer-slice.html",
+            "title": "JavaScript Demo: ArrayBuffer.slice()",
+            "type": "js"
+        },
+        "arraybufferTransfer": {
```
wbamberg: I think you need to remove this entry, otherwise `npm run build` throws an error.

wbamberg:
Closing this, because these examples are in https://github.com/mdn/interactive-examples/pull/237.

PR #237, opened 2017-08-25
Add dataview examples

maboa:
DataView examples ready for review :)

Note: the set and get examples for a specific type are identical.


site.json
```diff
@@ -177,6 +177,174 @@
...
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/dataview-getint16.html",
+            "fileName": "dataview-getint16.html",
+            "title": "JavaScript Demo: DataView.getInt16()",
+            "type": "js"
+        },
+        "dataviewGetInt32": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/dataview-getint32.html",
+            "fileName": "dataview-getint16.html",
```
wbamberg: This should be `dataview-getint32.html` I think

site.json
```diff
@@ -177,6 +177,174 @@
...
+            "exampleCode": "live-examples/js-examples/dataview-setfloat32.html",
+            "fileName": "dataview-setfloat32.html",
+            "title": "JavaScript Demo: DataView.setFloat32()",
+            "type": "js"
+        },
+        "dataviewSetFloat64": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/dataview-setfloat64.html",
+            "fileName": "dataview-setfloat64.html",
+            "title": "JavaScript Demo: DataView.setFloat65()",
```
wbamberg: 64 :)

site.json
```diff
@@ -177,6 +177,174 @@
...
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/dataview-setint16.html",
+            "fileName": "dataview-setint16.html",
+            "title": "JavaScript Demo: DataView.setInt16()",
+            "type": "js"
+        },
+        "dataviewSetInt32": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/dataview-setint32.html",
+            "fileName": "dataview-setint16.html",
```
wbamberg: Should be `"dataview-setint32.html"` I think

maboa:
Oops. Will fix :) BTW I didn't mean to pile these on top of the arraybuffer changes - not sure what I did wrong.

live-examples/js-examples/arraybuffer-slice.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">// create an ArrayBuffer with a size in bytes
```
schalkneethling: Missing ` class="language-js"`

live-examples/js-examples/arraybuffer-slice.html
```diff
@@ -0,0 +1,14 @@
...
+var buffer = new ArrayBuffer(16);
+
+var int32View = new Int32Array(buffer); // create the view
+int32View[2] = 42;
+
+var lastHalfView = new Int32Array(buffer.slice(8,16));
+
+console.log(lastHalfView[0]);
+
+// expected output: 42
```
schalkneethling: Nit: Some examples have an empty line between the `console` statement and the expected output line, and some do not. For consistency, and taking up less vertical space, lets go with, no empty line. Thoughts?

maboa: Sure. That's fine. Will rectify now :)

maboa:
Ensuring consistency around spacing.

maboa:
I'm going to make the ArrayBuffer changes on this branch. Apologies for the confusion.

maboa:
ArrayBuffer changes which makes spacing consistent with dataview examples. Also added class="language-js" to <code> tag as requested.

I'm wondering about creating a linter to catch style issues. Any thoughts on that?

schalkneethling:
@maboa 

> I'm wondering about creating a linter to catch style issues. Any thoughts on that?

Yes, we definitely want to do that. It is one of the first tasks on my last after the beta release.

live-examples/js-examples/arraybuffer-slice.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js" class="language-js">// create an ArrayBuffer with a size in bytes
+var buffer = new ArrayBuffer(16);
+
+var int32View = new Int32Array(buffer); // create the view
+int32View[2] = 42;
+var lastHalfView = new Int32Array(buffer.slice(8,16));
```
chrisdavidmills: Because the 2nd parameter is optional, this would also work if you did just slice(8), no? This could therefore be a little misleading. I'd include two examples, one with and one without the second parameter, to make it clear that it is optional, and show what it does.

live-examples/js-examples/dataview-byteoffset.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js" class="language-js">//create an ArrayBuffer
+var buffer = new ArrayBuffer(16);
+
+// Create a couple of views
+var view1 = new DataView(buffer);
+var view2 = new DataView(buffer,12,4); //from byte 12 for the next 4 bytes
+
+console.log(view1.byteOffset + view2.byteOffset); // 0 + 12
```
chrisdavidmills: Adding offsets together seems like a slightly bizarre thing to do to me - would you ever really do that? I'd just show the two offsets console.log'ed separately - this would indicate a comparison of the offsets of two different views, which would make more sense.

maboa: Hi Chris :) It is odd - agreed. But we're working on the assumption that we can only have one console.log for now (we've encountered issues with multiple console logs).

That said I think there is a case for simplifying here and just having view2 and its offset.

live-examples/js-examples/arraybuffer-slice.html
maboa: Fair point. OK to remove the second parameter in this case I think. Again we are working with the restriction of having a single console.log and one interactive example per page, so not sure we can include two examples.

live-examples/js-examples/dataview-byteoffset.html
chrisdavidmills: OK, fair enough — I wasn't aware of that limitation. Your simplification sounds good.

live-examples/js-examples/arraybuffer-slice.html
chrisdavidmills: I think having an example including the second parameter that actually needs it would be better. Maybe do a slice to return the middle 4 bytes?

maboa: Yes - makes sense.

live-examples/js-examples/dataview-byteoffset.html
maboa: OK will make that change 👍

live-examples/js-examples/arraybuffer-slice.html
maboa: BTW The limitation I'm coming up against is that it's difficult to console.log the whole array as that involves either looping (multiple console.logs) or some extra code to somehow 'flatten' first.

live-examples/js-examples/dataview-byteoffset.html
schalkneethling: @maboa That limitation is now removed with https://github.com/mdn/interactive-examples/pull/245 :)

maboa:
Thanks Schalk!

Issue #253, opened 2017-09-01
Add error messge for JS editor


atopal:
Currently the JS editor fails silently and simply doesn't print any output on errors. This was an issue during user testing and it's a major focus of feedback from the A/B test survey. Showing the error itself would be ideal, but we should show at least a general error message instead of failing silently, especially since beginners are very likely to make mistakes they don't understand.

PR #260, opened 2017-09-01
Write error message to the output

wbamberg:
I think this should be a fix for https://github.com/mdn/interactive-examples/issues/253.

### DIFF
--- a/js/editable-js.js
+++ b/js/editable-js.js
@@ -24,11 +24,7 @@
      * @param {string} body - The live example string to parse into a Function
      */
     function executeInteractiveExample(body) {
-        try {
-            new Function(body)();
-        } catch (e) {
-            return 'Error: ' + e.message;
-        }
+      new Function(body)();
     }
 
     function initInteractiveEditor() {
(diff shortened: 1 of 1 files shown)


js/editable-js.js
```diff
@@ -61,7 +57,13 @@
         var output = document.querySelector('#output code');
 
         output.classList.add('fade-in');
-        executeInteractiveExample(exampleCode);
+
+        try {
+          executeInteractiveExample(exampleCode);
```
schalkneethling: As this function no consist of just one line, I reckon we can get rid of it and just do `new Function(exampleCode)();` right inside the `try` block.

js/editable-js.js
```diff
@@ -61,7 +57,13 @@
...
+        }
+        catch (e) {
```
schalkneethling: Two nits: Ensure `catch` is on the same line as the closing `}`, and use `event` instead of `e` for consistency.

wbamberg:
PTAL :).

PR #263, opened 2017-09-04
added SharedArrayBuffer examples

maboa:
No conflicts this time. Hopefully I'm doing it right? :)

On the slice example I wondered whether to populate the entire view but I think we'd need a loop to do this. Open to suggestions though.

### DIFF
--- a/live-examples/js-examples/sharedarraybuffer-slice.html
+++ b/live-examples/js-examples/sharedarraybuffer-slice.html
@@ -1,0 +1,13 @@
+<pre>
+<code id="static-js" class="language-js">// create a SharedArrayBuffer with a size in bytes
+var buffer = new SharedArrayBuffer(16);
+var int32View = new Int32Array(buffer); // create the view
+// produces Int32Array [0, 0, 0, 0]
+
+int32View[1] = 42;
+var sliced = new Int32Array(buffer.slice(4,12));
+
+console.log(sliced);
+// expected output: 42,0
+</code>
+</pre>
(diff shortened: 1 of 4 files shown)


schalkneethling:
r+ from me but, I am leaving it open so @wbamberg can provide feedback on your question.

> On the slice example I wondered whether to populate the entire view but I think we'd need a loop to do this. Open to suggestions though.

Have you tried using a loop with a `console.log` inside? It should technically work and spit out each iteration.

maboa:
Thanks Schalk. I think the logging of output is OK. I was just wondering whether there was an easy way to set the `int32View` to say `[1, 2, 3, 4]` rather than leaving as `[0, 0, 0, 0]`. I don't think it's a big deal though.

BTW in the Promise Examples I'm using a console.log inside a loop at times 👍

live-examples/js-examples/sharedarraybuffer-slice.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js" class="language-js">// create a SharedArrayBuffer with a size in bytes
+var buffer = new SharedArrayBuffer(16);
+var int32View = new Int32Array(buffer); // create the view
+// produces Int32Array [0, 0, 0, 0]
+
+int32View[1] = 42;
+var sliced = new Int32Array(buffer.slice(4,12));
```
wbamberg: Nit: consider a comment for the arguments:

`// slice from index 4 to index 12`

PR #267, opened 2017-09-05
promise examples

maboa:
Promised promise examples as promised :)

### DIFF
--- a/live-examples/js-examples/promise-constructor.html
+++ b/live-examples/js-examples/promise-constructor.html
@@ -1,0 +1,9 @@
+<pre>
+<code id="static-js" class="language-js">var p = new Promise(function(resolve, reject) {
+  setTimeout(resolve, 100, 'foo');
+});
+
+console.log(p);
+// expected output: [object promise]
+</code>
+</pre>
--- a/live-examples/js-examples/promise-constructor.html
+++ b/live-examples/js-examples/promise-constructor.html
@@ -1,9 +1,9 @@
 <pre>
-<code id="static-js" class="language-js">var p = new Promise(function(resolve, reject) {
+<code id="static-js" class="language-js">var promise1 = new Promise(function(resolve, reject) {
   setTimeout(resolve, 100, 'foo');
 });
 
-console.log(p);
-// expected output: [object promise]
+console.log(promise1);
+// expected output: [object Promise]
 </code>
 </pre>
--- a/live-examples/js-examples/promise-race.html
+++ b/live-examples/js-examples/promise-race.html
@@ -1,0 +1,16 @@
+<pre>
+<code id="static-js" class="language-js">var p1 = new Promise(function(resolve, reject) {
+    setTimeout(resolve, 500, 'one');
+});
+
+var p2 = new Promise(function(resolve, reject) {
+    setTimeout(resolve, 100, 'two');
+});
+
+Promise.race([p1, p2]).then(function(value) {
+  console.log(value);
+  // Both resolve, but p2 is faster
+});
+// expected output: two
+</code>
+</pre>
--- a/live-examples/js-examples/promise-race.html
+++ b/live-examples/js-examples/promise-race.html
@@ -1,15 +1,15 @@
 <pre>
-<code id="static-js" class="language-js">var p1 = new Promise(function(resolve, reject) {
+<code id="static-js" class="language-js">var promise1 = new Promise(function(resolve, reject) {
     setTimeout(resolve, 500, 'one');
 });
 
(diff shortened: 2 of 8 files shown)


live-examples/js-examples/promise-constructor.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js" class="language-js">var p = new Promise(function(resolve, reject) {
```
schalkneethling: Nit: Not sure what the norm is with regards to examples on MDN but, I always feel like we should use variable names one would use in general. So here `var p` seems to not fit the bill. I would probably have done `var promise = ...` Thoughts @maboa @wbamberg

live-examples/js-examples/promise-race.html
```diff
@@ -0,0 +1,16 @@
+<pre>
+<code id="static-js" class="language-js">var p1 = new Promise(function(resolve, reject) {
+    setTimeout(resolve, 500, 'one');
+});
+
+var p2 = new Promise(function(resolve, reject) {
```
schalkneethling: Nit: Also here, perhaps use more descriptive variable names?

site.json
```diff
@@ -2,7 +2,7 @@
     "bundles": {
         "codeMirror": {
             "javascript": ["js/libs/codemirror.js", "js/libs/javascript.js"],
-            "css": ["css/libs/codemirror.css", "css/editor-libs/codemirror-override.css"],
+            "css": ["css/libs/codemirror.css"],
```
schalkneethling: Whoops, I believe this is "your master was behind mdn master when you created the feature branch" but, we do not want to remove that second stylesheet.

live-examples/js-examples/promise-constructor.html
maboa: Yes - that sounds reasonable Schalk. I always worry that it becomes slightly confusing / less readable having both promise and Promise floating about and for some reason I'm no longer a fan of the myPromise = new Promise ... type convention.

live-examples/js-examples/promise-race.html
maboa: ok

site.json
maboa: Good spot. I'll add it back in.

live-examples/js-examples/promise-constructor.html
wbamberg: Never ask a writer for their opinions on naming :).

I agree that `p` is not good but I don't like `promise` either. Since this is a timer, why not `timer`?

live-examples/js-examples/promise-reject.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js" class="language-js">Promise.reject(new Error('fail')).then(function(error) {
+  // not called
+}, function(error) {
+  console.log(error);
+});
+// expected output: Error: fail
+</code>
+</pre>
```
wbamberg: I found this example pretty hard to understand. Do you think something like this might be clearer:

```
function resolved(result) {
  console.log("Resolved");
}

function rejected(result) {
  console.log("Rejected");
}

Promise.reject('fail').then(resolved, rejected);

//expected output: Rejected
```

...is this correct though?

maboa: It is correct. I think. I shamelessly stole my example from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject where they describe encouraging the use of the Error object. The part I find confusing with reject (taking your code as an example) is that I'm not sure what we need to pass to `Promise.reject` to get `function resolved` to be called. ?

wbamberg: > I'm not sure what we need to pass to Promise.reject to get function resolved to be called?

As far as I know: there is nothing you can pass. If you call `reject()`, the promise is always rejected.

Good point about Error. You could still do this:

```
function resolved(result) {
  console.log("Resolved");
}

function rejected(result) {
  console.log(result);
}

Promise.reject(new Error('fail')).then(resolved, rejected);
//expected output: "Error: fail"
```

...but if you don't think this is clearer, I'm happy to accept the existing version too.

maboa: It seems odd that you can supply a function that will never be called. :shrug: I think your last example is a good compromise.

live-examples/js-examples/promise-constructor.html
maboa: In the end I went with promise1, promise2 etc. But I'm happy to change.

wbamberg: I'm fine with that :).

PR #259, opened 2017-09-01
typed-array examples

maboa:
I'm not entirely sure why this can't be automatically merged - sorry.

Let me know if this a problem and/or I can fix.

PR #268, opened 2017-09-05
Typedarray examples

maboa:
changes as per review


maboa:
Hold on. I messed up here. Let me try and fix. Sorry.

schalkneethling:
> This was meant to be ONE commit added to #259 :/ However it should contain all the necessary changes. ie it's PR 259 made good.

Ah, so you are saying that #259 is a duplicate of this one then?

live-examples/js-examples/typedarray-indexof.html
```diff
@@ -0,0 +1,8 @@
+<pre>
+<code id="static-js" class="language-js">var uint8 = new Uint8Array([10, 20, 30, 40, 50]);
+
+// check from position 3 onwards
+console.log(uint8.indexOf(50, 3));
+// expected output: 4
+</code>
+</pre>
```
wbamberg: Nit: Maybe also have an example that's not found?

```
var uint8 = new Uint8Array([10, 20, 30, 40, 50]);

// check from position 3 onwards
console.log(uint8.indexOf(50));
// expected output: 4

// check from position 3 onwards
console.log(uint8.indexOf(51));
// expected output: -1
```

live-examples/js-examples/typedarray-lastindexof.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js" class="language-js">var uint8 = new Uint8Array([10, 20, 50, 50, 50]);
+
+console.log(uint8.lastIndexOf(50, 4));  // check the first 4 elements
```
wbamberg: I think this is checking all 5 elements isn't it? (because it's zero-based, so "4" is actually the last item).

So (I think) that parameter is redundant here.

Again maybe consider using 2 logs here, one omitting the second argument, and one including it (but making it less than array.length):

```
var uint8 = new Uint8Array([10, 20, 50, 50, 50]);

// check all elements
console.log(uint8.lastIndexOf(50, 4));
// expected output: 4

// check the first 4 elements
console.log(uint8.lastIndexOf(50, 3));
// expected output: 3
```

live-examples/js-examples/typedarray-name.html
```diff
@@ -0,0 +1,5 @@
+<pre>
+<code id="static-js" class="language-js">console.log(Uint8Array.name);
+// expected output: Uint8Array
+</code>
+</pre>
```
wbamberg: Nit: Maybe worth having another of these, with a different type?

```
console.log(Uint8Array.name);
// expected output: Uint8Array

console.log(Float32Array.name);
// expected output: Float32Array
```

live-examples/js-examples/typedarray-set.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js" class="language-js">// create a SharedArrayBuffer with a size in bytes
+var buffer = new ArrayBuffer(8);
+var uint8 = new Uint8Array(buffer);
+
+uint8.set([1, 2, 3], 3);
```
wbamberg: Nit: This could perhaps benefit from a comment like "Copy the values given, starting at index 3", or something better than that.

live-examples/js-examples/typedarray-some.html
```diff
@@ -0,0 +1,12 @@
...
+  return element < 0;
+}
+
+var int8 = new Int8Array([-10, 20, -30, 40, -50]);
+
+// test every element - at least one must pass
+console.log(int8.some(isNegative));
+// expected output: true
+</code>
+</pre>
```
wbamberg: Nit: Again maybe worth considering an example that will return false:

```
function isNegative(element, index, array) {
  return element < 0;
}

var int8 = new Int8Array([-10, 20, -30, 40, -50]);
var positives = new Int8Array([10, 20, 30, 40, 50]);

console.log(int8.some(isNegative));
// expected output: true
console.log(positives.some(isNegative));
// expected output: false
```

... but be careful to keep to 12 lines :). I would omit the comment at line 8.

live-examples/js-examples/typedarray-sort.html
```diff
@@ -0,0 +1,8 @@
+<pre>
+<code id="static-js" class="language-js">var uint8 = new Uint8Array([40, 10, 50, 20, 30]);
+uint8.sort();
+
+console.log(uint8);
+// expected output: 10
```
wbamberg: should be:

`// expected output: 10,20,30,40,50`

live-examples/js-examples/typedarray-tostring.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js" class="language-js">var uint8 = new Uint8Array([10, 20, 30, 40, 50]);
+
+console.log(uint8.toString());
+// expected output: 10,20,30,40,50
+</code>
```
wbamberg: Nit: It's unfortunate that this looks the same as `console.log(uint8);`. Is it worth calling something stringy on it, and logging that?

```
var uint8 = new Uint8Array([10, 20, 30, 40, 50]);

var uint8String = uint8.toString();

console.log(uint8String.startsWith("10"));
// expected output: true
```

live-examples/js-examples/typedarray-buffer.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js" class="language-js">// create an ArrayBuffer with a size in bytes
```
schalkneethling: @maboa while you are still working on this one, feel free to remove the `class="language-js"` from these examples. I have cleaned up the rest. Thanks!

maboa:
Re https://github.com/mdn/interactive-examples/pull/268#issuecomment-327205229

@schalkneethling sorry I didn't answer your question above. This was a duplicate PR yes, but all good now :) Apologies for any confusion - I hit the wrong button. :/

PR #271, opened 2017-09-07
global properties

maboa:
I'm undecided whether the examples for eval, infinity, null etc are the best approach. Pushed here for discussion.

### DIFF
--- a/live-examples/js-examples/globalprops-decodeuri.html
+++ b/live-examples/js-examples/globalprops-decodeuri.html
@@ -1,0 +1,14 @@
+<pre>
+<code id="static-js">var uri = "https://mozilla.org/?x=шеллы";
+var encoded = encodeURI(uri);
+console.log(encoded);
+// expected output: https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+
+try {
+  console.log(decodeURI(encoded));
+} catch(e) { // catches a malformed URI
+  console.error(e);
+}
+// expected output: https://mozilla.org/?x=шеллы
+</code>
+</pre>
--- a/live-examples/js-examples/globalprops-decodeuri.html
+++ b/live-examples/js-examples/globalprops-decodeuri.html
@@ -6,10 +6,9 @@ console.log(encoded);
 
 try {
   console.log(decodeURI(encoded));
+  // expected output: https://mozilla.org/?x=шеллы
 } catch(e) { // catches a malformed URI
   console.error(e);
 }
-
-// expected output: https://mozilla.org/?x=шеллы
 </code>
 </pre>
--- a/live-examples/js-examples/globalprops-decodeuri.html
+++ b/live-examples/js-examples/globalprops-decodeuri.html
@@ -9,6 +9,7 @@ try {
 } catch(e) { // catches a malformed URI
   console.error(e);
 }
+
 // expected output: https://mozilla.org/?x=шеллы
 </code>
 </pre>
--- a/live-examples/js-examples/globalprops-decodeuricomponent.html
+++ b/live-examples/js-examples/globalprops-decodeuricomponent.html
@@ -1,0 +1,13 @@
+<pre>
+<code id="static-js">var uriComp = "?x=шеллы";
+var encoded = encodeURIComponent(uriComp);
+console.log(encoded);
+// expected output: %3Fx%3D%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+
+console.log(decodeURIComponent(encoded));
+// expected output: ?x=шеллы
+
+console.log(decodeURI(encoded));
+// expected output: %3Fx%3Dшеллы
+</code>
+</pre>
--- a/live-examples/js-examples/globalprops-decodeuricomponent.html
+++ b/live-examples/js-examples/globalprops-decodeuricomponent.html
@@ -1,13 +1,13 @@
 <pre>
-<code id="static-js">var uriComp = "?x=шеллы";
-var encoded = encodeURIComponent(uriComp);
-console.log(encoded);
-// expected output: %3Fx%3D%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+<code id="static-js">function containsEncodedComponents(x) {
+  // ie ?,=,&,/ etc
+	return (decodeURI(x) !== decodeURIComponent(x));
+}
 
-console.log(decodeURIComponent(encoded));
-// expected output: ?x=шеллы
+console.log(containsEncodedComponents("%3Fx%3Dtest")); // ?x=test
+// expected output: true
 
-console.log(decodeURI(encoded));
-// expected output: %3Fx%3Dшеллы
+console.log(containsEncodedComponents("%D1%88%D0%B5%D0%BB%D0%BB%D1%8B")); // шеллы
+// expected output: false
 </code>
 </pre>
--- a/live-examples/js-examples/globalprops-decodeuricomponent.html
+++ b/live-examples/js-examples/globalprops-decodeuricomponent.html
@@ -1,7 +1,7 @@
 <pre>
 <code id="static-js">function containsEncodedComponents(x) {
   // ie ?,=,&,/ etc
-	return (decodeURI(x) !== decodeURIComponent(x));
+  return (decodeURI(x) !== decodeURIComponent(x));
 }
 
 console.log(containsEncodedComponents("%3Fx%3Dtest")); // ?x=test
--- a/live-examples/js-examples/globalprops-encodeuri.html
+++ b/live-examples/js-examples/globalprops-encodeuri.html
@@ -1,0 +1,15 @@
+<pre>
+<code id="static-js">var uri = "https://mozilla.org/?x=шеллы";
+var encoded = encodeURI(uri);
+console.log(encoded);
+// expected output: https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+
+try {
+  console.log(decodeURI(encoded));
+} catch(e) { // catches a malformed URI
+  console.error(e);
+}
+
+// expected output: https://mozilla.org/?x=шеллы
+</code>
+</pre>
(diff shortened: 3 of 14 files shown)


live-examples/js-examples/globalprops-decodeuri.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">var uri = "https://mozilla.org/?x=шеллы";
+var encoded = encodeURI(uri);
+console.log(encoded);
+// expected output: https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+
+try {
+  console.log(decodeURI(encoded));
+} catch(e) { // catches a malformed URI
+  console.error(e);
```
schalkneethling: Just a note here. We are not capturing `console.error` so, if someone plays around with this to test the `try/catch` functionality, they will not see the `console.error` unless they have their dev tools open :-/

Perhaps this is something we need to also address? @wbamberg

live-examples/js-examples/globalprops-eval.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">function equivalence(x,y){
+  if (eval(x) === eval(y)) return "Strong";
+  if (eval(x.toString()) === eval(y.toString())) return "Weak";
```
schalkneethling: In terms of code style, I am wondering if we should follow the project wide style for `if` conditionals i.e. The above should be written as:

```
if (eval(x) === eval(y)) {
    return "Strong";
}

if (eval(x.toString()) === eval(y.toString())) {
    return "Weak";
}
```

live-examples/js-examples/globalprops-decodeuri.html
maboa: Good to know. Happy to adjust if needed.

live-examples/js-examples/globalprops-eval.html
maboa: I agree that it's probably better form to write them the way you're doing, the issue is space. Actually now I think of it in more recent examples (ie numbers) I've been doing stuff like:

```
if (isNaN(x)) { return 0 }
```
So we should stick a style and I should stick with it :)

live-examples/js-examples/globalprops-decodeuri.html
wbamberg: I think we can omit support for most of the [console API](https://developer.mozilla.org/en-US/docs/Web/API/console) (and some would be difficult).

But it would be nice to have `console.error()`, and it doesn't seem like it would be too difficult.

live-examples/js-examples/globalprops-eval.html
schalkneethling: I completely agree. I also entirely get the number of lines problem so, it is a balance we need to find. I am leaning towards to first example of the second but, would like to know @wbamberg thoughts on this.

maboa:
Thanks @wbamberg and @schalkneethling. It may be that I am overcomplicating matters. In most cases I'm attempting to illustrate a couple of the issues mentioned in the page that the example will be embedded in. Perhaps the solution is to illustrate only one point, just enough to give a flavour. 

Also by simplifying we should be able to save a bit of space and pad out our braces in a more readable style :)

live-examples/js-examples/globalprops-decodeuri.html
schalkneethling: `console.error` support has landed on master ~ https://github.com/mdn/interactive-examples/pull/279

maboa: woohoo!

live-examples/js-examples/globalprops-eval.html
wbamberg: I think we should have:

```
if (eval(x) === eval(y)) {
    return "Strong";
}
```

I feel pretty strongly that we should not sacrifice readability for compactness.

What would be lovely would be to have:
(1) a style guide for JS code samples, not just for this but for all samples on MDN
(2) lint rules to enforce it

The nearest thing I have found is https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Code_samples#JavaScript but it's pretty limited.

wbamberg: @schalkneethling , does webdev have guidelines we can steal?

schalkneethling: There is this, but I am pretty sure they are outdated :( https://mozweb.readthedocs.io/en/latest/reference/js-style.html

schalkneethling:
> What would be lovely would be to have:
> (1) a style guide for JS code samples, not just for this but for all samples on MDN
> (2) lint rules to enforce it

A big Yes to both of these.

maboa:
It's tricky to find the middle ground between a series of console.logs and over-contrived examples. Hopefully I have found it, keeping to a consistent style to boot :) Happy to discuss synchronously.

Near the top of my list of 'lovely to have' features would be `console.log` outputting quotes around strings. Often we want to show the difference between strings and numbers.

wbamberg:
> Near the top of my list of 'lovely to have' features would be console.log outputting quotes around strings. Often we want to show the difference between strings and numbers.

I just opened https://github.com/mdn/interactive-examples/pull/280. Not sure if it's the right approach, but it seems to work better than nothing.

live-examples/js-examples/globalprops-decodeuricomponent.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">function containsEncodedComponents(x) {
+  // ie ?,=,&,/ etc
+	return (decodeURI(x) !== decodeURIComponent(x));
```
wbamberg: I think this is tab-indented, which it should not be.

(for indentation it seems you generally use 2 spaces, right? Again this is a place where having a standard and linting would be great. Looking at existing JS code samples on MDN, they seem to use 2. So 2 sounds good to me.)

live-examples/js-examples/globalprops-encodeuri.html
```diff
@@ -0,0 +1,15 @@
...
+console.log(encoded);
+// expected output: https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B
+
+try {
+  console.log(decodeURI(encoded));
+} catch(e) { // catches a malformed URI
+  console.error(e);
+}
+
+// expected output: https://mozilla.org/?x=шеллы
```
wbamberg: same comment about this line, I would have expected it at 9.

live-examples/js-examples/globalprops-infinity.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var maxNumber = Math.pow(10, 1000); // max positive number
+
+if (maxNumber === Infinity) {
+	console.log("Let's call it Infinity!");
+	// expected output: Let's call it Infinity!
+}
```
wbamberg: Tab indentation in these 2 lines.

live-examples/js-examples/globalprops-null.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">function getVowels(str) {
+  var m = str.match(/[aeiou]/gi);
+  if (m === null) {
```
wbamberg: Would you do this, or would you just do `(!m)`? I'm not sure if this matters...

live-examples/js-examples/globalprops-parsefloat.html
```diff
@@ -0,0 +1,15 @@
...
+console.log(circumference(4.567));
+// expected output: 28.695307297889173
+
+console.log(circumference("4.567abcdefgh"));
+// expected output: 28.695307297889173
+
+console.log(circumference("abcdefgh"));
+//expected output: NaN
+</code>
+</pre>
```
wbamberg: really nice example :)

live-examples/js-examples/globalprops-nan.html
```diff
@@ -0,0 +1,15 @@
...
+  return x;
+}
+
+console.log(sanitise("1"));
+// expected output: 1
+
+console.log(sanitise("NotANumber"));
+// expected output: NaN
+</code>
+</pre>
```
wbamberg: really nice example :)

live-examples/js-examples/globalprops-undefined.html
```diff
@@ -0,0 +1,14 @@
...
+  if (t === undefined) {
+     return "Undefined value!";
+  }
+  return t;
+}
+
+var x;
+
+console.log(test(x));
+// expected output: 0
```
wbamberg: Should be:

`// expected output: ""Undefined value!"`

live-examples/js-examples/globalprops-decodeuri.html
```diff
@@ -0,0 +1,15 @@
...
+}
+
+// expected output: https://mozilla.org/?x=шеллы
```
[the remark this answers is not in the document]
maboa: Yes. Good call.

live-examples/js-examples/globalprops-decodeuricomponent.html
maboa: Well spotted :) Yes 2 spaces - I recently turned invisible characters on in my editor (Atom) as copying and pasting from the interactive editor was causing issues. Unfortunately this one slipped through the net. I suspect there might be others out there - but we can remedy with a PR dedicated to formatting.

live-examples/js-examples/globalprops-encodeuri.html
maboa: Yup.

live-examples/js-examples/globalprops-infinity.html
maboa: D'oh! :)

live-examples/js-examples/globalprops-null.html
maboa: I was struggling to find an example that would return null. Some people are in favour of very expressive code so things like `m == false` as opposed to `!m` and I can see that point of view.

live-examples/js-examples/globalprops-parsefloat.html
maboa: Thanks :) I tried to simplify after your previous review - hopefully this works.

live-examples/js-examples/globalprops-undefined.html
maboa: :+1:

PR #280, opened 2017-09-14
Format logged objects to indicate their type

wbamberg:
Here's a simple attempt to format logged objects to indicate their type:

* wrap strings in quotes
* wrap arrays in square brackets

### DIFF
--- a/js/editor-libs/console.js
+++ b/js/editor-libs/console.js
@@ -17,6 +17,24 @@
     };
 
     /**
+    * Formats output to indicate its type:
+    * - quotes around strings
+    * - square brackets around arrays
+    * @param {any} input - The output to log.
+    */
+    function indicateType(input) {
+        switch (typeof(input)) {
+            case "string":
+                input = `"${input}"`;
+            case "object":
+                if (Array.isArray(input)) {
+                    input = `[${input}]`;
+                }
+        }
+        return input;
+    }
+
+    /**
      * Writes the provided content to the editor’s output area
      * @param {String} content - The content to write to output
      */
@@ -34,7 +52,7 @@
 
     // eslint-disable-next-line no-console
     console.log = function(loggedItem) {
-        writeOutput(loggedItem);
+        writeOutput(indicateType(loggedItem));
         // do not swallow console.log
         originalConsoleLogger.apply(console, arguments);
     };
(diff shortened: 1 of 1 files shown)


wbamberg:
This is pretty basic actually. It doesn't understand objects in arrays:

```
console.log(["1", 1]
> [1,1]
```
Maybe we should add that?

js/editor-libs/console.js
```diff
@@ -17,6 +17,22 @@
...
 
     /**
+    * Formats output to indicate its type:
+    * - quotes around strings
+    * - square brackets around arrays
+    * @param {any} input - The output to log.
+    */
+    function indicateType(input) {
+        if (typeof(input) === "string") {
+            return `"${input}"`;
```
schalkneethling: Unfortunately template literals are not supported in IE[1] and we still need to support at least v11 :(
[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

js/editor-libs/console.js
```diff
@@ -17,6 +17,22 @@
     };
 
     /**
+    * Formats output to indicate its type:
+    * - quotes around strings
+    * - square brackets around arrays
+    * @param {any} input - The output to log.
+    */
```
schalkneethling: Nit: Indent lines 20-24 by one space

js/editor-libs/console.js
```diff
@@ -17,6 +17,22 @@
...
+    function indicateType(input) {
```
schalkneethling: Nit: Wondering if `formatOutput` might be a more apt name?

js/editor-libs/console.js
```diff
@@ -17,6 +17,22 @@
     };
 
     /**
+    * Formats output to indicate its type:
+    * - quotes around strings
+    * - square brackets around arrays
+    * @param {any} input - The output to log.
```
schalkneethling: Nit: Add `@returns Formatted output as a string`

js/editor-libs/console.js
wbamberg: *sobs*

wbamberg:
Thanks for the review @schalkneethling ! I've addressed the issues.

PR #282, opened 2017-09-18
Date - JS Interactive Examples

maboa:
Note that there is a browser inconsistency for Date.toLocaleTimeString (try Chrome, Safari and Firefox). We can of course remove that example.

Apologies for the merge conflict, let me know if this a problem.

### DIFF
--- a/live-examples/js-examples/date-getdate.html
+++ b/live-examples/js-examples/date-getdate.html
@@ -1,8 +1,8 @@
 <pre>
 <code id="static-js">var birthday = new Date('August 19, 1975 23:15:30');
-var day = birthday.getDate();
+var date1 = birthday.getDate();
 
-console.log(day);
+console.log(date1);
 // expected output: 19
 </code>
 </pre>
--- a/live-examples/js-examples/date-getdate.html
+++ b/live-examples/js-examples/date-getdate.html
@@ -1,0 +1,8 @@
+<pre>
+<code id="static-js">var birthday = new Date('August 19, 1975 23:15:30');
+var day = birthday.getDate();
+
+console.log(day);
+// expected output: 19
+</code>
+</pre>
--- a/live-examples/js-examples/date-getday.html
+++ b/live-examples/js-examples/date-getday.html
@@ -1,0 +1,9 @@
+<pre>
+<code id="static-js">var birthday = new Date('Tue, August 19, 1975 23:15:30');
+var day = birthday.getDay();
+// Sunday - Saturday : 0 - 6
+
+console.log(day);
+// expected output: 2
+</code>
+</pre>
--- a/live-examples/js-examples/date-getday.html
+++ b/live-examples/js-examples/date-getday.html
@@ -1,9 +1,9 @@
 <pre>
-<code id="static-js">var birthday = new Date('Tue, August 19, 1975 23:15:30');
-var day = birthday.getDay();
+<code id="static-js">var birthday = new Date('August 19, 1975 23:15:30');
+var day1 = birthday.getDay();
 // Sunday - Saturday : 0 - 6
 
-console.log(day);
+console.log(day1);
 // expected output: 2
 </code>
 </pre>
--- a/live-examples/js-examples/date-getmilliseconds.html
+++ b/live-examples/js-examples/date-getmilliseconds.html
@@ -1,5 +1,6 @@
 <pre>
-<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18.123');
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18');
+moonLanding.setMilliseconds(123);
 
 console.log(moonLanding.getMilliseconds());
 // expected output: 123
--- a/live-examples/js-examples/date-getmilliseconds.html
+++ b/live-examples/js-examples/date-getmilliseconds.html
@@ -1,0 +1,7 @@
+<pre>
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18.123');
+
+console.log(moonLanding.getMilliseconds());
+// expected output: 123
+</code>
+</pre>
--- a/live-examples/js-examples/date-gettime.html
+++ b/live-examples/js-examples/date-gettime.html
@@ -1,0 +1,8 @@
+<pre>
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18');
+
+// milliseconds since Jan 1, 1970, 00:00:00.000
+console.log(moonLanding.getTime());
+// expected output: -14261982000
+</code>
+</pre>
--- a/live-examples/js-examples/date-gettime.html
+++ b/live-examples/js-examples/date-gettime.html
@@ -1,8 +1,8 @@
 <pre>
-<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18');
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18 GMT+00:00');
 
-// milliseconds since Jan 1, 1970, 00:00:00.000
+// milliseconds since Jan 1, 1970, 00:00:00.000 GMT
 console.log(moonLanding.getTime());
-// expected output: -14261982000
+// expected output: -14254782000
 </code>
 </pre>
--- a/live-examples/js-examples/date-gettimezoneoffset.html
+++ b/live-examples/js-examples/date-gettimezoneoffset.html
@@ -1,0 +1,12 @@
+<pre>
+<code id="static-js">var date1 = new Date('Tue, August 19, 1975 23:15:30 GMT+07:00');
+var date2 = new Date('Tue, August 19, 1975 23:15:30 GMT-02:00');
+
+console.log(date1.getTimezoneOffset());
+// expected output: your local timezone offset in minutes
+// (eg -120). NOT the timezone offset of the date object.
+
+console.log(date1.getTimezoneOffset === date2.getTimezoneOffset);
+// expected output: true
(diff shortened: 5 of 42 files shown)


live-examples/js-examples/date-gettimezoneoffset.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var date1 = new Date('Tue, August 19, 1975 23:15:30 GMT+07:00');
+var date2 = new Date('Tue, August 19, 1975 23:15:30 GMT-02:00');
+
+console.log(date1.getTimezoneOffset());
+// expected output: your local timezone offset in minutes
+// (eg -120). NOT the timezone offset of the date object.
+
+console.log(date1.getTimezoneOffset === date2.getTimezoneOffset);
```
wbamberg: Should this be:

`console.log(date1.getTimezoneOffset() === date2.getTimezoneOffset());`

live-examples/js-examples/date-getutcdate.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">var date1 = new Date('Tue, August 19, 1975 23:15:30 GMT+11:00');
+var date2 = new Date('Tue, August 19, 1975 23:15:30 GMT-11:00');
+
+console.log(date1.getUTCDate());
+// expected output: 19
+
+console.log(date2.getUTCDate());
+// expected output: 20
+</code>
```
wbamberg: Nice example :).

live-examples/js-examples/date-getutcmonth.html
```diff
@@ -0,0 +1,13 @@
...
+<code id="static-js">var date1 = new Date('December 31, 1975, 23:15:30.123 GMT+11:00');
+var date2 = new Date('December 31, 1975, 23:15:30.123 GMT-11:00');
+
+// December
+console.log(date1.getUTCMonth());
+// expected output: 11
+
+// January
+console.log(date2.getUTCMonth());
+// expected output: 0
```
wbamberg: I get `NaN` for  both of these.

live-examples/js-examples/date-parse.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">var unixTimeZero = Date.parse('Thu, 01 Jan 1970 00:00:00 GMT');
+var javaScriptRelease = Date.parse('04 Dec 1995 00:12:00 GMT');
+var millisElapsed = javaScriptRelease - unixTimeZero;
+
+console.log(millisElapsed);
+//expected output: 818035920000
```
wbamberg: I think it's a bit strange doing arithmetic here when `unixTimeZero` === 0, so `javaScriptRelease` === `millisElapsed`.

I know it's boring, but I think it would be fine to just log `unixTimeZero` and  `javaScriptRelease` here.

live-examples/js-examples/date-setmilliseconds.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var event = new Date('Tue, August 19, 1975 23:15:30.123');
+
+console.log(event.getMilliseconds());
+// expected output: 123
+
+event.setMilliseconds(456);
+
+console.log(event.getMilliseconds());
+// expected output: 456
```
wbamberg: I get `NaN` for these.

live-examples/js-examples/date-tolocaletimestring.html
```diff
@@ -0,0 +1,15 @@
...
+console.log(event.toLocaleTimeString('en-US'));
+// expected output: 11:15:30 PM
+
+console.log(event.toLocaleTimeString('it-IT'));
+// expected output: 23:15:30
+
+console.log(event.toLocaleTimeString('ar-EG'));
+// expected output: ١١:١٥:٣٠ م
+
+// NOTE: THERE SEEMS TO BE SOME INCONSISTENCY BETWEEN BROWSERS HERE!
```
wbamberg: I don't think we should include this note. If it works across most browsers, we should just ship with the example and rely on the browser compat data to tell people when it doesn't work. If there's clear inconsistency between browsers, we might want to do something like:

```
// expected output (Chrome): XYZ
// expect output (Firefox): ABC
```

live-examples/js-examples/date-valueof.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">var utcDate1 = new Date(Date.UTC(96, 1, 2, 3, 4, 5));
+var utcDate2 = new Date(Date.UTC(0, 0, 0, 0, 0, 0));
+
+console.log(utcDate1.toUTCString());
+// expected output: Fri, 02 Feb 1996 03:04:05 GMT
+
+console.log(utcDate2.toUTCString());
+// expected output: Sun, 31 Dec 1899 00:00:00 GMT
```
wbamberg: This isn't an example of `valueOf`.

live-examples/js-examples/date-getdate.html
```diff
@@ -0,0 +1,8 @@
+<pre>
+<code id="static-js">var birthday = new Date('August 19, 1975 23:15:30');
+var day = birthday.getDate();
```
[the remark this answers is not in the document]
maboa: Yes. Although to differentiate from `Date`, maybe something like `date1` is better.

live-examples/js-examples/date-getday.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">var birthday = new Date('Tue, August 19, 1975 23:15:30');
```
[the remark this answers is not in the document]
maboa: Yes - I guess it won't allow you to specify a non-valid date. We should omit.

live-examples/js-examples/date-getmilliseconds.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18.123');
+
+console.log(moonLanding.getMilliseconds());
+// expected output: 123
```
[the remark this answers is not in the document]
maboa: Aha - I think I forgot to test this x-browser. Looks like you can't specify milliseconds like this in Firefox or Safari. Works in Chrome for me though. I'll figure out a new example.

live-examples/js-examples/date-gettime.html
```diff
@@ -0,0 +1,8 @@
+<pre>
+<code id="static-js">var moonLanding = new Date('July 20, 69 00:20:18');
+
+// milliseconds since Jan 1, 1970, 00:00:00.000
+console.log(moonLanding.getTime());
+// expected output: -14261982000
```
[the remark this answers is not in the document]
maboa: Ah OK - probably because Unix time is GMT. Will need to use UTC time.

live-examples/js-examples/date-gettimezoneoffset.html
maboa: Yup.

live-examples/js-examples/date-getutcdate.html
maboa: :)

live-examples/js-examples/date-getutcmonth.html
maboa: Same issue with the millisecond definition in Firefox and Safari. Will fix.

live-examples/js-examples/date-parse.html
maboa: OK

live-examples/js-examples/date-setmilliseconds.html
maboa: Yes - same issues in Firefox and Safari regarding specifying a date with milliseconds.

live-examples/js-examples/date-tolocaletimestring.html
maboa: OK

live-examples/js-examples/date-valueof.html
maboa: Whoops. Looks like the same example twice.

live-examples/js-examples/date-tolocaletimestring.html
maboa: @wbamberg any chance you could check that you are seeing the same issue? I think Chrome returns in UTC. For me it logs 10:15:30 whereas Firefox logs 11:15:30.

maboa: Sorry - just saw your reply further up. Even setting the time as GMT, Chrome is always an hour behind :/

wbamberg: Does it make a difference if you do something like:

`console.log(event.toLocaleTimeString('en-US'), {timeZone: 'UTC'});`

? Otherwise, I'd be tempted to say something hand-wavy like: "the exact output is dependent on your timezone and local configuration".

live-examples/js-examples/date-setmilliseconds.html
wbamberg: This one still isn't fixed.

live-examples/js-examples/date-tolocaletimestring.html
wbamberg: I wondered if you'd tried the above?

live-examples/js-examples/date-valueof.html
wbamberg: This still needs to be fixed.

maboa:
Apologies @wbamberg - not sure how these slipped through the net. Adjusted and pushed now.

PR #286, opened 2017-09-21
JS expression and operator examples

maboa:
Not sure about the file naming conventions for these examples.

Note that where I couldn't think of an alternative some examples are effectively a series of console logs.

Conceivably I could make the arithmetic example simpler - I'm maybe trying to kill too many birds there.

### DIFF
--- a/live-examples/js-examples/expressions-commaoperators.html
+++ b/live-examples/js-examples/expressions-commaoperators.html
@@ -1,0 +1,15 @@
+<pre>
+<code id="static-js">var x = 1;
+
+console.log(x++, x);
+// expected output: 1
+
+console.log(x);
+// expected output: 2
+
+x = (y = 2, z = 3);
+
+console.log(x);
+// expected output: 3
+</code>
+</pre>
--- a/live-examples/js-examples/expressions-commaoperators.html
+++ b/live-examples/js-examples/expressions-commaoperators.html
@@ -1,13 +1,12 @@
 <pre>
 <code id="static-js">var x = 1;
 
-console.log(x++, x);
-// expected output: 1
+x = (x++, x);
 
 console.log(x);
 // expected output: 2
 
-x = (y = 2, z = 3);
+x = (2, 3);
 
 console.log(x);
 // expected output: 3
--- a/live-examples/js-examples/expressions-functionasteriskexpression.html
+++ b/live-examples/js-examples/expressions-functionasteriskexpression.html
@@ -5,9 +5,12 @@
   yield 'c';
 }
 
-const [...values] = foo();
+var str = "";
+for (let val of foo()) {
+  str = str + val;
+}
 
-console.log(values);
-// expected output: [a,b,c]
+console.log(str);
+// expected output: "abc"
 </code>
 </pre>
--- a/live-examples/js-examples/expressions-functionasteriskexpression.html
+++ b/live-examples/js-examples/expressions-functionasteriskexpression.html
@@ -1,0 +1,13 @@
+<pre>
+<code id="static-js">function* foo() {
+  yield 'a';
+  yield 'b';
+  yield 'c';
+}
+
+const [...values] = foo();
+
+console.log(values);
+// expected output: [a,b,c]
+</code>
+</pre>
--- a/live-examples/js-examples/expressions-functionexpression.html
+++ b/live-examples/js-examples/expressions-functionexpression.html
@@ -1,0 +1,14 @@
+<pre>
+<code id="static-js">try {
+  notHoisted();
+}
+catch(e) {
+  console.log(e);
+  // expected output: TypeError: notHoisted is not a function
+}
+
+var notHoisted = function() {
+  console.log("bar");
+};
+</code>
+</pre>
--- a/live-examples/js-examples/expressions-functionexpression.html
+++ b/live-examples/js-examples/expressions-functionexpression.html
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">function getRectArea2(width, height) {
+<code id="static-js">function getRectArea(width, height) {
   return width * height;
 }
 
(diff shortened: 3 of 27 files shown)


live-examples/js-examples/expressions-commaoperators.html
```diff
@@ -0,0 +1,15 @@
+<pre>
+<code id="static-js">var x = 1;
+
+console.log(x++, x);
+// expected output: 1
```
wbamberg: I think this is problematic because in a real console (and this one, if we implement it) that won't be the output. Perhaps we can just omit this bit?

live-examples/js-examples/expressions-commaoperators.html
```diff
@@ -0,0 +1,15 @@
...
+
+console.log(x);
+// expected output: 2
+
+x = (y = 2, z = 3);
```
wbamberg: Nit: I'm not sure why this can't just be `x=(2, 3)`.

live-examples/js-examples/expressions-functionasteriskexpression.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">function* foo() {
+  yield 'a';
+  yield 'b';
+  yield 'c';
+}
+
+const [...values] = foo();
```
wbamberg: Nit: To me this looks a bit complex. Maybe something like:

```
function* foo() {
  yield 'a';
  yield 'b';
  yield 'c';
}

for (let val of foo()) {
  console.log(val);
}

// expected output: "a","b","c"
```

would be easier to grasp?

live-examples/js-examples/expressions-functionexpression.html
```diff
@@ -0,0 +1,14 @@
...
+  notHoisted();
+}
+catch(e) {
+  console.log(e);
+  // expected output: TypeError: notHoisted is not a function
+}
+
+var notHoisted = function() {
+  console.log("bar");
+};
```
wbamberg: I feel as if this example is confusing - yes, not-hoistedness is an important aspect of function expressions, but it's not the `#1` most basic thing about them, and it's odd for the first simplest example to be one that doesn't work.

site.json
```diff
@@ -394,6 +394,188 @@
...
+            "fileName": "expressions-inoperator.html",
+            "title": "JavaScript Demo: Expressions - In Operator",
+            "type": "js"
+        },
+        "expressionsInstanceof": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/expressions-instanceof.html",
+            "fileName": "expressions-inoperator.html",
+            "title": "JavaScript Demo: Expressions - In Operator",
+            "type": "js"
```
wbamberg: in several places here, `in` should be `instanceof`

live-examples/js-examples/expressions-logicaloperator.html
```diff
@@ -0,0 +1,15 @@
...
+  }
+  return 0;
+}
+
+console.log(addIfBothPositive(2, 3));
+// expected output: 5
+
+console.log(addIfBothPositive(-2, 3));
+// expected output: 0
+</code>
```
wbamberg: Nit: It's totally up to you, but it seems a shame to demonstrate only one of the operators. Maybe worth considering something like:

```
var a = 3;
var b = -2;

console.log(a > 0 && b > 0);
// expected output: false

console.log(a > 0 || b > 0);
// expected output: true

console.log(!(a > 0 || b > 0));
// expected output: false
```

live-examples/js-examples/expressions-voidoperator.html
```diff
@@ -0,0 +1,15 @@
...
+}();
+
+try {
+  test();
+}
+catch(e) {
+  console.log(e);
+  // expected output: ReferenceError: test is not defined
+}
+</code>
```
wbamberg: Weird, but I guess this is a weird bit of JS :).

live-examples/js-examples/expressions-commaoperators.html
maboa: Oh interesting - good spot.

live-examples/js-examples/expressions-commaoperators.html
maboa: I guess it was to show that you can do things within the brackets. Have revised so that we have something like this:

```
var x = 1;

x = (x++, x);

console.log(x);
// expected output: 2

x = (2, 3);

console.log(x);
// expected output: 3
```

live-examples/js-examples/expressions-functionasteriskexpression.html
maboa: Yes we could do this. Note that I've been avoiding console.logs within loops up to now. Maybe I shouldn't? It's just the output will be on separate lines. ie

```
> "a"
> "b"
> "c"
```

and I'm not sure how to represent that neatly.

live-examples/js-examples/expressions-functionexpression.html
maboa: Hmm. Yes I think you're right.

site.json
maboa: thanks

live-examples/js-examples/expressions-logicaloperator.html
maboa: sure!

live-examples/js-examples/expressions-voidoperator.html
maboa: There is so much weird JS - but we're getting there :)

I think the point was to demo that you could only call as a self-invoking function.

live-examples/js-examples/expressions-functionasteriskexpression.html
wbamberg: > So I'd probably be tempted to do :
...

Yes, I think this would be OK.

I do take your point that " It's just the output will be on separate lines...and I'm not sure how to represent that neatly." But I think, realistically that if we say:

`// expected output: "a","b","c"`

and they get:

```
> "a"
> "b"
> "c"
```
...then most people will understand this. Maybe not, though :).

PR #288, opened 2017-09-27
JS Statements and Declarations Examples

maboa:
There's a bit of overlap with the Expressions Examples but I think that's ok.

The Async example is a little large - any ideas on how to reduce this is welcome :)

### DIFF
--- a/live-examples/js-examples/statement-async.html
+++ b/live-examples/js-examples/statement-async.html
@@ -1,0 +1,21 @@
+<pre>
+<code id="static-js">function resolveAfter2Seconds(x) {
+  return new Promise(resolve => {
+    console.log("waiting...");
+    setTimeout(() => {
+      resolve(x);
+    }, 2000);
+  });
+}
+
+async function add1(x) {
+  var a = await resolveAfter2Seconds(20);
+  return(a + x);
+}
+
+add1(10).then(v => {
+  console.log(v);
+  // expected output: 30
+});
+</code>
+</pre>
--- a/live-examples/js-examples/statement-async.html
+++ b/live-examples/js-examples/statement-async.html
@@ -1,21 +1,20 @@
 <pre>
 <code id="static-js">function resolveAfter2Seconds(x) {
   return new Promise(resolve => {
-    console.log("waiting...");
     setTimeout(() => {
       resolve(x);
     }, 2000);
   });
 }
 
-async function add1(x) {
-  var a = await resolveAfter2Seconds(20);
-  return(a + x);
+async function add1() {
+  console.log('calling');
+  var a = await resolveAfter2Seconds('resolved');
+
+  console.log(a);
+  // expected output: "resolved"
 }
 
-add1(10).then(v => {
-  console.log(v);
-  // expected output: 30
-});
+add1();
 </code>
 </pre>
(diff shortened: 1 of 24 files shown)


live-examples/js-examples/statement-async.html
```diff
@@ -0,0 +1,21 @@
...
+  var a = await resolveAfter2Seconds(20);
+  return(a + x);
+}
+
+add1(10).then(v => {
+  console.log(v);
+  // expected output: 30
+});
+</code>
+</pre>
```
wbamberg: I might be missing something, but I thought something like this might be simpler:

```
function resolveAfter2Seconds(x) {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve(x);
    }, 2000);
  });
}

async function add1() {
  console.log("calling");
  var a = await resolveAfter2Seconds(20);
  console.log(a);
  console.log("finished");
}

add1();
```
It's still pretty long, but to me it illustrates more clearly the way you can use async/await to write linear async code. Or am I missing something?

live-examples/js-examples/statement-block.html
```diff
@@ -0,0 +1,14 @@
...
+
+if (x === 1) {
+  var x = 2;
+}
+
+console.log(x);
+// expected output: 2
+
+// Note using let or const instead of var
+// would result in 1 being outputted.
```
wbamberg: Hmm, consider illustrating that?

```
var x = 1;
let y = 1;

if (true) {
  var x = 2;
  let y = 2;
}

console.log(x);
// expected output: 2

console.log(y);
// expected output: 1
```

live-examples/js-examples/statement-async.html
maboa: Sure! Simple is good! Although the example size will grow somewhat once we put the `// expected output:` comments in. I do wonder whether we need `// expected output:` comments for all `console.log`s.

live-examples/js-examples/statement-block.html
maboa: Nice example. Will definitely switch! :)

live-examples/js-examples/statement-async.html
wbamberg: Yes, agreed: as we discussed:

* I don't think always we need `// expected output:`, especially in cases like this where the expected output is very obvious.

* we could omit one of the logs, by doing something like:

```
function resolveAfter2Seconds() {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve("finished");
    }, 2000);
  });
}

async function callAsync() {
  console.log("calling");
  var result = await resolveAfter2Seconds();
  console.log(result);
}

callAsync();
```

live-examples/js-examples/statement-async.html
```diff
@@ -0,0 +1,20 @@
...
+async function add1() {
+  console.log('calling');
+  var a = await resolveAfter2Seconds('resolved');
+
+  console.log(a);
+  // expected output: "resolved"
+}
+
+add1();
+</code>
```
wbamberg: I think given that we're overlong already we should omit the blank line at 13. Also `add1` is now a weird name, and something like `result` might be better than `a` for the return value.

Also I wondered if the example would be a bit less weird if instead of taking `"resolved"` as an argument, it takes no arguments and just returns `"resolved"`:

```
function resolveAfter2Seconds() {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve('resolved');
    }, 2000);
  });
}

async function asyncCall() {
  console.log('calling');
  var result = await resolveAfter2Seconds();
  console.log(result);
  // expected output: "resolved"
}

asyncCall();
```

maboa: Makes sense. That's done now :)

PR #290, opened 2017-09-28
Boolean, Function(s) and JSON JS Examples

maboa:
Bundled these into one PR :)

### DIFF
--- a/live-examples/js-examples/functions-restparameters.html
+++ b/live-examples/js-examples/functions-restparameters.html
@@ -1,15 +1,14 @@
 <pre>
-<code id="static-js">function func1(...[a, b, c]) {
-  return a + b + c;
+<code id="static-js">function sum(...theArgs) {
+  return theArgs.reduce((previous, current) => {
+    return previous + current;
+  });
 }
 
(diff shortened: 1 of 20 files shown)


live-examples/js-examples/functions-restparameters.html
```diff
@@ -0,0 +1,15 @@
...
+console.log(func1(1));
+// expected output: NaN
+
+console.log(func1(1, 2, 3));
+// expected output: 6
+
+console.log(func1(1, 2, 3, 4));
+// expected output: 6
+</code>
+</pre>
```
wbamberg: Just a suggestion, but I wonder if it would be simpler to have an example that doesn't also introduce destructuring at the same time?

e.g.
```

function sum(...theArgs) {
  return theArgs.reduce((previous, current) => {
    return previous + current;
  });
}

console.log(sum(1, 2, 3));
// expected output: 6

console.log(sum(1, 2, 3, 4));
// expected output: 10
```

maboa: Yes - I think it would. Good call!

PR #291, opened 2017-09-29
console.log support for arrays of arrays and arrays of strings

maboa:
This is a tentative pull request for the enhancement of the console log types.

Building on the work by @wbamberg this change should add support for arrays of arrays and arrays of strings. As this method takes advantage of recursion it *should* work to any level of 'nesting'.

Note: we still don't have support for `Int32Array` et al as `Array.isArray()` returns `false` in those cases. Not sure what the best way to test for these cases are, but would be nice to, especially for the `SharedBufferArray` examples.

PS I added a space after commas and made it single quotes around strings, which I think is inline with our preference and browser console output.

### DIFF
--- a/js/editor-libs/console.js
+++ b/js/editor-libs/console.js
@@ -29,7 +29,7 @@
         var output = '';
         for (var i = 0; i < input.length; i++) {
             if (typeof(input[i]) === "string") {
-                output += "'" + input[i] + "'";
+                output += '"' + input[i] + '"';
             } else if (Array.isArray(input[i])) {
                 output += '[';
                 output += formatArray(input[i]);
@@ -56,7 +56,7 @@
      */
      function formatOutput(input) {
          if (typeof(input) === "string") {
-             return "'" + input + "'";
+             return '"' + input + '"';
          } else if (Array.isArray(input)) {
            // check the contents of the array
            return '[' + formatArray(input) + ']';
@@ -64,7 +64,7 @@
              return input;
          }
      }
-    
+
     /**
      * Writes the provided content to the editor’s output area
      * @param {String} content - The content to write to output
--- a/js/editor-libs/console.js
+++ b/js/editor-libs/console.js
@@ -17,22 +17,54 @@
     };
 
     /**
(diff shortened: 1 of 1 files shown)


maboa:
I made string output single quote but looking at various browser consoles - I'm seeing double quotes around strings. Maybe we should attempt to be as consistent as possible with browser consoles.

wbamberg:
Now you mention it, I seem to remember that that was probably why the first implementation used double quotes. Thanks for the PR, I am buried in Firefox 57 for the doc updates for the next few days but will look soon, unless someone else gets there first :).

js/editor-libs/console.js
```diff
@@ -17,23 +17,55 @@
     };
 
     /**
-     * Formats output to indicate its type:
-     * - quotes around strings
+     * Formats arrays:
+     * - quotes around strings in arrays
      * - square brackets around arrays
+     * - adds commas appropraitely (with spacing)
```
schalkneethling: Nit: "appropraitely" > "appropriately"

js/editor-libs/console.js
```diff
@@ -17,23 +17,55 @@
...
-    function formatOutput(input) {
-        if (typeof(input) === "string") {
-            return '"' + input + '"';
-        } else if (Array.isArray(input)) {
-          return '[' + input + ']';
-        } else {
-            return input;
+    function formatArray(input) {
+        var output = '';
+        for (var i = 0; i < input.length; i++) {
```
schalkneethling: Cache the length of the array:

## Suggestion:

```
for (var i = 0, l = input.length; i < l; i++) {
```

js/editor-libs/console.js
```diff
@@ -17,23 +17,55 @@
...
+            if (typeof(input[i]) === "string") {
+                output += '"' + input[i] + '"';
+            } else if (Array.isArray(input[i])) {
+                output += '[';
+                output += formatArray(input[i]);
+                output += ']';
+            } else {
+              output += input[i];
```
schalkneethling: Nit: indentation

PR #294, opened 2017-10-02
Classes and Object JS examples

maboa:


live-examples/js-examples/classes-extends.html
```diff
@@ -0,0 +1,16 @@
+<pre>
+<code id="static-js">class formatDate extends Date {
+  constructor(dateStr) {
+    super(dateStr);
+  }
+
+  getFormattedDate() {
+    var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
```
wbamberg: Very nice example, but this line is too long, we should split it (I know this will make the example even longer, but I'd prefer that).

```
    var months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
                  'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
```

live-examples/js-examples/classes-extends.html
```diff
@@ -0,0 +1,16 @@
...
+    return this.getDate() + '-' + months[this.getMonth()] + '-' + this.getFullYear();
```
wbamberg: Let's use a template literal here :):

```
    return `${this.getDate()}-${months[this.getMonth()]}-${this.getFullYear()}`;
```

live-examples/js-examples/object-prototype-tolocalestring.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var date1 = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));
+
+console.log(date1.toLocaleString('ar-EG'));
+// expected output: "٢٠‏/١٢‏/٢٠١٢ ٤:٠٠:٠٠ ص"
+
+var obj1 = new Object();
+
+console.log(obj1.toLocaleString('ar-EG'));
+// expected output: "[object Object]"
```
wbamberg: I don't understand what this example is showing...

live-examples/js-examples/classes-extends.html
maboa: Sure. Makes sense!

live-examples/js-examples/classes-extends.html
maboa: Sounds good. Let's push things forward :)

live-examples/js-examples/object-prototype-tolocalestring.html
maboa: Hmmm. Yes - I see your point. I wanted to show how toLocaleString could be applied to objects, keeping the object intact, but I guess it's a bit nebulous.

PR #297, opened 2017-10-06
copy edit of the readme file

chrisdavidmills:


README.md
```diff
@@ -231,13 +233,11 @@ console.log(result);
 </pre>
 ```
 
-That is all for the live example piece. All that remains is to tell the page generator about our new example. To do this, open up `site.json` at the root of the project folder. Under `pages`, find an exising entry with the `type` of `js`.
-
-Copy and paste an example.
+That is all for the live example piece. All that remains is to tell the page generator about our new example. To do this, open up `site.json` at the root of the project folder. Under `pages`, find an existing entry with a `type` of `js`.
 
-> The file is ordered by `type`, and then alphabetically for each `type`.
+Copy and paste the example then update it to apply to your new example, noting that pages are grouped by `type`, and then alphabetically for each `type`.
```
schalkneethling: Should this "Copy and paste the example then update i" be "Copy and paste the example, then update i"

README.md
```diff
@@ -124,9 +126,9 @@ your CSS goes here</code></pre>
 </div>
 ```
 
-This is base starting point for all CSS examples. Your next step is to fill in the example element. For `border-raius` it makes sense to have a simple div element with a solid background color. For this example then, the already present `example-element` div will do, however we do need to give it some basic styling, and perhaps add the text "Style Me".
+This is the base starting point for all CSS examples. Your next step is to fill in the example element. For `border-radius` it makes sense to have a simple `&lt;div&gt;` element with a solid background color. The already present `example-element` `&lt;div&gt;` will do, however we do need to give it some basic styling, and perhaps add the text "Style Me".
```
schalkneethling: Here you do not need to escape the < and >. Because it is inside back ticks, it will be processed as code. If escaped as above, the output will be `&lt;div&gt;` instead of `<div>`

README.md
```diff
@@ -124,9 +126,9 @@ your CSS goes here</code></pre>
...
 
-First then, inside the `example-element` div add the text "Style Me".
+First then, inside the `example-element` `&lt;div&gt;' add the text "Style Me", like so:
```
schalkneethling: Same here.

README.md
```diff
@@ -146,7 +148,7 @@ Next, create a new CSS file inside `live-examples\css-examples\css\`. Call this
 }
 ```
 
-Next we need to add some different examples of using `border-radius`. In the `pre` tab from before, you will see there is a nested `example-choice` element. For each example you wish to add, you will add one of these with the CSS style rule to applied, as is content, for example:
+Next we need to add some different examples of using `border-radius`. In the `&lt;pre&gt;` element we first saw above, you will see there is a nested `&lt;div&gt;` with a class of `example-choice`. For each example you wish to add, you will add one of these with the CSS declaration to apply as its content, for example:
```
schalkneethling: Here as well.

PR #300, opened 2017-10-11
JS Map and WeakMap Prototype Examples

maboa:
Just a couple of quick notes:

1. I wasn't sure what to call Map.prototype[@@toStringTag] and Map.prototype[@@iterator]() in `site.json`

2. I left out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/@@species


live-examples/js-examples/weakmap-prototype-has.html
```diff
@@ -0,0 +1,14 @@
...
+
+weakmap1.set(object1, 42);
+
+console.log(weakmap1.get(object1));
+// expected output: 42
+
+console.log(weakmap1.get(object2));
+// expected output: undefined
+</code>
+</pre>
```
wbamberg: This doesn't seem to be an example of `has`

live-examples/js-examples/weakmap-prototype-set.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">var weakmap1 = new WeakMap();
+var object1 = new Object();
+var object2 = new Object();
+
+weakmap1.set(object1, 'foo').set(object2, 'bar');
```
wbamberg: I think I would prefer:

```
weakmap1.set(object1, 'foo');
weakmap1.set(object2, 'bar');
```

live-examples/js-examples/map-prototype-@@iterator.html
```diff
@@ -0,0 +1,15 @@
...
+
+map1.set('0', 'foo');
+map1.set(1, 'bar');
+
+var iterator1 = map1[Symbol.iterator]();
+
+console.log(iterator1.next().value);
+// expected output: ["0", "foo"]
+
+console.log(iterator1.next().value);
```
wbamberg: I made that same comment for the Set examples: consider using for...of instead of next().value?

```
var map1 = new Map();

map1.set('0', 'foo');
map1.set(1, 'bar');

var iterator1 = map1[Symbol.iterator]();

for (let item of iterator1) {
  console.log(item);
}
```

live-examples/js-examples/weakmap-prototype-has.html
maboa: Yep - silly mistake my end. Duplicated previous example, wrote over it but didn't save for some reason. Should have picked this up in review. Apologies.

live-examples/js-examples/weakmap-prototype-set.html
maboa: Fair enough.

live-examples/js-examples/map-prototype-@@iterator.html
maboa: OK. Cool.

live-examples/js-examples/map-prototype-set.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">var map1 = new Map();
+map1.set('bar', 'foo');
+
+console.log(map1.get('bar'));
+// expected output: "foo"
+
+console.log(map1.get('baz'));
```
wbamberg: This is not an example for `set` (sorry, I must have missed this last time).

live-examples/js-examples/map-prototype-set.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">var map1 = new Map();
+map1.set('bar', 'foo');
```
maboa: Sorry @wbamberg - I'm not following - the second line sets. The gets show what  to expect when you query a set map and not.

wbamberg: Sorry Mark, you are right of course.

PR #301, opened 2017-10-12
JS Set and WeakSet Prototype Examples

maboa:
Similar but different to Map and WeakMap :)

To think about:

Should we instantiate an object with `object1 = new Object()` or `object1 = {}` ?

### DIFF
--- a/live-examples/js-examples/set-prototype-add.html
+++ b/live-examples/js-examples/set-prototype-add.html
@@ -1,15 +1,14 @@
 <pre>
 <code id="static-js">var set1 = new Set();
-var object1 = new Object();
 
 set1.add(42);
-set1.add(object1).add('forty two');
-// chainable
+set1.add(42);
+set1.add(13);
 
-console.log(set1);
-// expected output: [object Set]
-
-console.log(set1.size);
-// expected output: 3
+for (let item of set1) {
+  console.log(item);
+  // expected output: 42
+  // expected output: 13
+}
 </code>
 </pre>
--- a/live-examples/js-examples/set-prototype-add.html
+++ b/live-examples/js-examples/set-prototype-add.html
@@ -1,0 +1,15 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+var object1 = new Object();
+
+set1.add(42);
+set1.add(object1).add('forty two');
+// chainable
+
+console.log(set1);
+// expected output: [object Set]
+
+console.log(set1.size);
+// expected output: 3
+</code>
+</pre>
--- a/live-examples/js-examples/set-prototype-delete.html
+++ b/live-examples/js-examples/set-prototype-delete.html
@@ -1,0 +1,15 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+set1.add({x: 10, y: 20}).add({x: 20, y: 30});
+
+// Delete any point with `x > 10`.
+set1.forEach(function(point){
+  if(point.x > 10){
+    set1.delete(point)
+  }
+})
+
+console.log(set1.size);
+// expected output: 1
+</code>
+</pre>
--- a/live-examples/js-examples/set-prototype-delete.html
+++ b/live-examples/js-examples/set-prototype-delete.html
@@ -4,10 +4,10 @@ set1.add({x: 10, y: 20}).add({x: 20, y: 30});
 
 // Delete any point with `x > 10`.
 set1.forEach(function(point){
-  if(point.x > 10){
-    set1.delete(point)
+  if (point.x > 10) {
+    set1.delete(point);
   }
-})
+});
 
 console.log(set1.size);
 // expected output: 1
--- a/live-examples/js-examples/set-prototype-entries.html
+++ b/live-examples/js-examples/set-prototype-entries.html
@@ -5,10 +5,10 @@ set1.add('forty two');
 
 var iterator1 = set1.entries();
 
-console.log(iterator1.next().value);
-// expected output: [42, 42]
-
-console.log(iterator1.next().value);
-// expected output: ["forty two", "forty two"]
+for (let entry of iterator1) {
+  console.log(entry);
+  // expected output: [42, 42]
+  // expected output: ["forty two", "forty two"]
+}
 </code>
 </pre>
--- a/live-examples/js-examples/set-prototype-entries.html
+++ b/live-examples/js-examples/set-prototype-entries.html
@@ -1,0 +1,14 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+set1.add(42);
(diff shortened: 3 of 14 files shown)


live-examples/js-examples/set-prototype-delete.html
```diff
@@ -0,0 +1,15 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+set1.add({x: 10, y: 20}).add({x: 20, y: 30});
+
+// Delete any point with `x > 10`.
+set1.forEach(function(point){
+  if(point.x > 10){
```
wbamberg: spacing: -> `if (point.x > 10) { `

live-examples/js-examples/set-prototype-delete.html
```diff
@@ -0,0 +1,15 @@
...
+    set1.delete(point)
```
wbamberg: missing semicolon

live-examples/js-examples/set-prototype-entries.html
```diff
@@ -0,0 +1,14 @@
...
+set1.add('forty two');
+
+var iterator1 = set1.entries();
+
+console.log(iterator1.next().value);
+// expected output: [42, 42]
+
+console.log(iterator1.next().value);
+// expected output: ["forty two", "forty two"]
+</code>
```
wbamberg: what a mad API!

wbamberg: This is strictly optional of course, but using `for...of` seems more like a normal use than `iterator.next().value`:

```
var set1 = new Set();
set1.add(42);
set1.add('forty two');

var iterator1 = set1.entries();

for (let entry of iterator1) {
  console.log(entry);
}
```

live-examples/js-examples/set-prototype-foreach.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">function logSetElements(value1, value2, set) {
+  console.log('s[' + value1 + '] = ' + value2);
```
wbamberg: the `Map` version of this used a template literal and I got all excited. Should we use it for both? (these are technically ES 2015, right?)

live-examples/js-examples/set-prototype-size.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+var object1 = new Object();
+
+set1.add(42);
+set1.add('forty two');
+set1.add(object1);
```
wbamberg: Again it might be worth adding something twice here:
```
set1.add('forty two'); 
```

live-examples/js-examples/set-prototype-values.html
```diff
@@ -0,0 +1,14 @@
...
+<code id="static-js">var set1 = new Set();
+set1.add(42);
+set1.add('forty two');
+
+var iterator1 = set1.values();
+
+console.log(iterator1.next().value);
+// expected output: 42
+
+console.log(iterator1.next().value);
```
wbamberg: See above, consider using `for...of`?

live-examples/js-examples/weakset-prototype-add.html
```diff
@@ -0,0 +1,13 @@
...
+<code id="static-js">var weakset1 = new WeakSet();
+var object1 = new Object();
+
+weakset1.add(object1);
+console.log(weakset1.has(object1));
+// would normally return true
+
+console.log(weakset1.add(1));
+// expected output: "Error: Invalid value used in weak set" in Chrome
+// expected output: "Error: WeakSet value must be an object, got the number 1" in Firefox
```
wbamberg: This is tricky, isn't it, because we don't get the first console.log(), which is what soeone might expect. Do you think it's worth doing something like:

```
var weakset1 = new WeakSet();
var object1 = new Object();

weakset1.add(object1);
console.log(weakset1.has(object1));
// expected output: true

try {
  weakset1.add(1);
} catch(error) {
  console.log(error);
  // Error message
}
```
? I'm not sure tbh.

wbamberg:
> Should we instantiate an object with object1 = new Object() or object1 = {} ?

As we discussed, I prefer `object1 = {}`. But I'm not militant about it

live-examples/js-examples/set-prototype-add.html
```diff
@@ -0,0 +1,15 @@
+<pre>
+<code id="static-js">var set1 = new Set();
+var object1 = new Object();
+
+set1.add(42);
+set1.add(object1).add('forty two');
+// chainable
+
+console.log(set1);
+// expected output: [object Set]
```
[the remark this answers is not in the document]
maboa: I suppose (since we had space) it was to show that a `Set` was an `object`. But on reflection it does feel like it's not really required. I'll remove it.

live-examples/js-examples/set-prototype-add.html
```diff
@@ -0,0 +1,15 @@
...
+
+console.log(set1.size);
+// expected output: 3
+</code>
+</pre>
```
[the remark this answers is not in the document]
maboa: yup. good call.

live-examples/js-examples/set-prototype-delete.html
maboa: on it

live-examples/js-examples/set-prototype-delete.html
maboa: 👍

live-examples/js-examples/set-prototype-entries.html
maboa: ok

live-examples/js-examples/set-prototype-foreach.html
maboa: yes ... yes I think we should :)

live-examples/js-examples/set-prototype-size.html
maboa: ok

live-examples/js-examples/set-prototype-values.html
maboa: ok

live-examples/js-examples/set-prototype-entries.html
maboa: I suppose the advantage of using `iterator.next().value` is that it makes the `console.log` output slightly clearer, maybe. Bur `for...of` is neater, so swings and roundabouts :)

live-examples/js-examples/weakset-prototype-add.html
maboa: I think it makes sense to try and catch errors. Will amend.

wbamberg:
Looks great, thanks @maboa !

PR #306, opened 2017-10-16
JS Symbol Examples

maboa:
Note - I couldn't think (or find) any suitable examples for:

- Symbol.prototype.valueOf()
- Symbol.prototype[@@toPrimitive]

### DIFF
--- a/live-examples/js-examples/symbol-keyfor.html
+++ b/live-examples/js-examples/symbol-keyfor.html
@@ -1,0 +1,15 @@
+<pre>
+<code id="static-js">var globalSym = Symbol.for('foo'); // global symbol
+
+console.log(Symbol.keyFor(globalSym));
+// expected output: "foo"
+
+var localSym = Symbol(); // local symbol
+
+console.log(Symbol.keyFor(localSym));
+// expected output: undefined
+
+console.log(Symbol.keyFor(Symbol.iterator));
+// expected output: undefined
+</code>
+</pre>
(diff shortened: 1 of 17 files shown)


live-examples/js-examples/symbol-keyfor.html
```diff
@@ -0,0 +1,15 @@
...
+// expected output: "foo"
+
+var localSym = Symbol(); // local symbol
+
+console.log(Symbol.keyFor(localSym));
+// expected output: undefined
+
+console.log(Symbol.keyFor(Symbol.iterator));
+// expected output: undefined
+</code>
```
wbamberg: I get a weird error running this in the editor:

Firefox:

    Error: t is undefined

Chrome:

    Error: Cannot read property 'constructor' of undefined
(I think I have updated the editor code to the laster master.)

maboa: Confirmed - specifically it seems that these lines are causing the issue:

```
var localSym = Symbol(); // local symbol

console.log(Symbol.keyFor(localSym));
// expected output: undefined

console.log(Symbol.keyFor(Symbol.iterator));
// expected output: undefined
```

Note that this wasn't an issue in the `symbol-example` branch, so it looks like it's related to some code we pushed between this and a more recent master. Will investigate.

maboa: https://github.com/mdn/interactive-examples/pull/321 should fix :)

PR #315, opened 2017-10-26
JS RegExp examples (with caveats)

maboa:
I don't necessarily think we should merge this, but I put it out there for discussion.

The issues are:

1. `regexp-prototype-source.html` - specifically:

```
console.log(new RegExp('\n').source === '\\n');
// expected output: true
```

Which I believe should return `true` (source : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source)

It returns `true` in Firefox and Safari but `false` in Chrome and Opera .

2. `regexp-prototype-tostring.html` - (possibly related to 1.) specifically:

```
console.log(new RegExp('\n', 'g').toString());
// expected output: "/\n/g"
```

This produces the expected output in Firefox and Safari but in Chrome and Opera produces:

```
"/
/g"
```

### DIFF
--- a/live-examples/js-examples/regexp-prototype-source.html
+++ b/live-examples/js-examples/regexp-prototype-source.html
@@ -2,7 +2,7 @@
 <code id="static-js">var regex1 = /fooBar/ig;
 
 console.log(regex1.source);
-// expected output: "foobar"
+// expected output: "fooBar"
 
 console.log(new RegExp().source);
 // expected output: "(?:)"
--- a/live-examples/js-examples/regexp-prototype-source.html
+++ b/live-examples/js-examples/regexp-prototype-source.html
@@ -1,0 +1,14 @@
+<pre>
+<code id="static-js">var regex1 = /fooBar/ig;
+
+console.log(regex1.source);
+// expected output: "foobar"
+
+console.log(new RegExp().source);
+// expected output: "(?:)"
+
+console.log(new RegExp('\n').source === '\\n');
+// expected output: true (starting with ES5)
+// (due to escaping)
+</code>
+</pre>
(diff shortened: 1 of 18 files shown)


wbamberg:
I can reproduce the behavior you're describing with the Chrome console.

I think perhaps this is a genuine incompatibility. In the [browser compat table for this API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source#Browser_compatibility) there's a row labeled "Escaping" that presumably describes support for this [feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/source#Empty_regular_expressions_and_escaping).  It gets a "?" for Chrome and Opera, I wonder if that should be a "No".

@Elchi3 , do you know anything about this one?

live-examples/js-examples/regexp-prototype-source.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">var regex1 = /fooBar/ig;
+
+console.log(regex1.source);
+// expected output: "foobar"
```
wbamberg: "fooBar"

live-examples/js-examples/regexp-prototype-tostring.html
```diff
@@ -0,0 +1,17 @@
+<pre>
+<code id="static-js">console.log(new RegExp('a+b+c'));
+// expected output: a+b+c
```
wbamberg: I'm getting: `/a+b+c/` here (on Firefox)

live-examples/js-examples/regexp-prototype-source.html
maboa: Yes.

live-examples/js-examples/regexp-prototype-tostring.html
maboa: Yup.

PR #321, opened 2017-11-06
Fix for console.log(undefined)

maboa:
This was picked up by @wbamberg while testing the `symbol` examples : https://github.com/mdn/interactive-examples/pull/306#discussion_r148442680

Looks like we needed to check for `undefined` and let it fall through.


wbamberg:
This looks good to me, but @schalkneethling , would you please take a look as well?

js/editor-libs/console-utils.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
             }
         }
 
-        if (objectName === 'Symbol') {
+        if (objectName === 'Symbol' && (typeof input != 'undefined')) {
```
schalkneethling: Because we are using `'use strict';` so we can safely check against `undefined`. Also use `!==`

```
if (objectName === 'Symbol' && (typeof input !== undefined)) {
```

js/editor-libs/console-utils.js
```diff
@@ -74,7 +74,9 @@ module.exports = {
      */
     formatOutput: function(input) {
         'use strict';
-        if (typeof input === 'string') {
+        if (typeof input === 'undefined') {
```
schalkneethling: Same here, we can simply use `if (typeof input === undefined) {`

js/editor-libs/console-utils.js
maboa: Ok - good points :)

js/editor-libs/console-utils.js
maboa: Sure.

maboa:
Aha - just tested this - all good on the `!==` but putting `undefined` in quotes causes a regression to the previous error. @schalkneethling

schalkneethling:
> but putting undefined in quotes causes a regression 

@maboa You mean taking away the quotes? Yeah, what I neglected to change in the code is to remove the `typeof`, as that will convert it to a string again.

```
if (input === undefined) {
```

maboa:
OK thanks @schalkneethling - all clear now :) I kinda had it like that originally but then looked around the web a bit to see what folk recommended and the `typeof` method came up. But I guess `use strict` means we can just use a straight comparison with confidence.

PR #335, opened 2017-11-16
Set different cache times for different assets to fix #334

### LINKED ISSUE #334: Set different cache times for different assets
This continues the work started in #311 

Currently, all files served from https://interactive-examples.mdn.mozilla.net have a max age of 900 seconds (15 minutes). This includes both the embedded samples HTML and the static assets shared among samples.

Users will re-request any file after 15 minutes. If the file hasn't changed, CloudFront should send a ``304 Not Modified``, which is smaller and faster than sending the whole file. This also means it will take as much at 15 minutes for users to see changes to samples, asset, etc.

This is probably OK. However, tools like https://www.webpagetest.org will advise us that some assets, such as [codemirror.js](https://interactive-examples.mdn.mozilla.net/js/codemirror-5-31-0.js), should have longer cache times. 1 week is usually the recommended minimum, but web page test will accept down to 1 hour.

We can use very long cache times for files that never change. If the file does change, we can change the name, so that it is a different file. Some asset pipelines allow you to embed a hash of the file in the name to get this effect, and others embed a version in the asset name.

Different cache times for different assets will complicate publishing to S3 (we can't just use ``aws s3 sync``), so I'd like to make sure we know what cache times we want for each asset before making the change.

Here's my proposal:

* /pages - 1 minute (optimize for time to see new version)
* /live-examples - 1 minute (optimize for time to see new version)
* /media - 1 week (developer will need to change name to get new version). Includes:
  - media/clippy.svg
  - media/dino.svg
  - media/example.svg
  - media/firefox.png
  - media/fonts/FiraSans-Bold.woff2
  - media/fonts/FiraSans-ExtraBold.woff2
  - media/fonts/FiraSans-Italic.woff2
  - media/fonts/FiraSans-Light.woff2
  - media/fonts/FiraSans-Regular.woff2
  - media/fonts/FiraSans-SemiBold.woff2
  - media/fonts/FiraSans-SemiBoldItalic.woff2
  - media/warning.svg
* /css/codemirror-5-31-0.css - 1 week
* /css/css-examples-libs-1-8-4.css - 1 week
* /css/editor-css.css - 1 minute
* /css/editor-js.css - 1 minute
* /js/codemirror-5-31-0.js - 1 week
* /js/css-examples-libs-1-8-4.js - 1 week
* /js/editor-css.js - 1 minute
* /js/editor-js.js - 1 minute

With this proposal:
* /pages, /live-examples, and /media will be synced
* /css and /js will be transferred file-by-file, and the sync script will need to be updated when files are added or removed.

If you anticipate often adding or removing assets in /css and /js, then you'll need to split them into /css and /js (1-minute cache) and /css-lib and /js-lib (1-week cache). Different folders will allow us to use sync to upload them.

jgmize:
This is based on the @jwhitlock's proposal in #334. This is WIP, has not yet been tested, and is intended to further the discussion with a concrete example implementation.

### DIFF
--- a/bin/s3-sync.sh
+++ b/bin/s3-sync.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+if [[ -z "$1" ]]; then
+    echo "Bucket name required"
+    exit 1
+fi
+
+SHORT_CACHE="--cache-control max-age=60" # 1 minute
+LONG_CACHE="--cache-control max-age=604800" # 1 week
+ARGS="--acl public-read --delete --profile mdninteractive"
+cd ../docs
+
+for path in "pages live-examples"; do
+    aws s3 sync $path s3://$1 ${SHORT_CACHE} ${ARGS}
+done
+
+for path in "css js"; do
+    aws s3 sync $path s3://$1 --exclude '*' --include 'editor-*' ${SHORT_CACHE} ${ARGS}
+    aws s3 sync $path s3://$1 --exclude 'editor-*' ${LONG_CACHE} ${ARGS}
+done
+
+aws s3 sync media s3://$1 ${LONG_CACHE} ${ARGS}
--- a/bin/s3-sync.sh
+++ b/bin/s3-sync.sh
@@ -5,8 +5,8 @@ if [[ -z "$1" ]]; then
     exit 1
 fi
 
-SHORT_CACHE="--cache-control max-age=60" # 1 minute
-LONG_CACHE="--cache-control max-age=604800" # 1 week
+SHORT_CACHE="--cache-control max-age=1800" # 30 minutes
+LONG_CACHE="--cache-control max-age=2629800" # 1 month
 ARGS="--acl public-read --delete --profile mdninteractive"
 cd ../docs
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ def buildSite() {
 def syncS3(String bucket) {
     stage ('s3 sync') {
         try {
-          sh "cd docs && aws s3 sync . s3://" + bucket +" --acl public-read --delete --profile mdninteractive --cache-control max-age=900"
+          sh "bin/s3-sync.sh " + bucket
         } catch(err) {
           sh "bin/irc-notify.sh --stage 's3 sync " + env.BRANCH_NAME + "' --status 'failed'"
           throw err


bin/s3-sync.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+if [[ -z "$1" ]]; then
+    echo "Bucket name required"
+    exit 1
+fi
+
+SHORT_CACHE="--cache-control max-age=60" # 1 minute
```
jgmize: I used the 1 minute value suggested by @jwhitlock in #335 but I would prefer a longer time, say 5-15 minutes.

bin/s3-sync.sh
```diff
@@ -0,0 +1,22 @@
...
+LONG_CACHE="--cache-control max-age=604800" # 1 week
+ARGS="--acl public-read --delete --profile mdninteractive"
+cd ../docs
+
+for path in "pages live-examples"; do
+    aws s3 sync $path s3://$1 ${SHORT_CACHE} ${ARGS}
+done
+
+for path in "css js"; do
+    aws s3 sync $path s3://$1 --exclude '*' --include 'editor-*' ${SHORT_CACHE} ${ARGS}
```
jgmize: include and exclude syntax based on http://docs.aws.amazon.com/cli/latest/userguide/using-s3-commands.html

schalkneethling:
@jgmize Thank you for starting this. Some suggestions.

The following bundles are library based so, they can have a much longer cache.

codemirror-5-31-0.js
codemirror-5-31-0.css
css-examples-libs-1-8-4.js
css-examples-libs-1-8-4.css

These are our own bundles and I would say a cache setting of 30min would be totally fine:

editor-css.css
editor-css.js
editor-js.css
editor-js.js

The image assets and fonts here https://github.com/mdn/interactive-examples/tree/master/media are unlikely to change much but, there might very be new additions so, here we can also use a much longer cache ttl.

bin/s3-sync.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+if [[ -z "$1" ]]; then
+    echo "Bucket name required"
+    exit 1
+fi
+
+SHORT_CACHE="--cache-control max-age=60" # 1 minute
+LONG_CACHE="--cache-control max-age=604800" # 1 week
```
jgmize: @schalkneethling how long would you like this to be?

schalkneethling: I would feel comfortable with a month. If there is a new release of one of the dependency libs, the version number on the file will change and bust the cache right?

For our own files, 30 minutes would be ideal if possible, but could be longer if it makes more sense. Thanks!

jwhitlock:
Removing myself from review. The main thing to review is if the cache times are correct for the app, and @schalkneethling is the right person for that. I won't know if the script works until it is run, and I think @jgmize should do that.

schalkneethling:
@jgmize If my comment here[1] makes sense and is implemented with the PR, then I reckon we are ready to merge and give it a test run. Thanks!

[1] https://github.com/mdn/interactive-examples/pull/335#discussion_r152544506

PR #338, opened 2017-11-17
More concise syntax for linear-gradient

wbamberg:
Short lines are better, especially for the CSS examples, so this change makes the syntax for linear-gradient more concise by using the default axis.


live-examples/css-examples/background-image.html
```diff
@@ -24,7 +24,7 @@
 
 <div class="example-choice" initial-choice="true">
 <pre>
-<code id="example_four" class="language-css">background-image: linear-gradient(to top, rgba(255,255,0,0.5), rgba(0,0,255,0.5)),
+<code id="example_four" class="language-css">background-image: linear-gradient(rgba(0,0,255,0.5), rgba(255,255,0,0.5)),
```
schalkneethling: Nit: use spaces after commas in parameter list.

schalkneethling:
One small nit, and one conflict. Once that is sorted, this is r+ 🍄

wbamberg:
@schalkneethling , should be fixed now.

PR #339, opened 2017-11-19
Update background-size example

wbamberg:
Here are some updates to the background-size example.

The original example is too tall to fit without scrolling, so I've reduced the vertical height to prevent that:

* removed comments: I think the code is pretty self-explanatory, and given how precious vertical space is, we should avoid comments unless they are really needed

* removed one of the examples, which covered quite similar ground to another one

I've also chosen a different image: I chose this for a few reasons:

* the aspect seems like it will work better to show the difference between keywords like `cover` and `contain`

* the shape is very simple and recognizable, so distortions when you set width and height are more obvious

I'm happy to talk about these changes though if you don't agree.


wbamberg:
> Works for me. I am wondering though if perhaps we want to add all the images used in examples inside a examples folder inside media to keep em separated ;p Also, its worth running these images through either ImageOptim or http://tinypng.com

I should have made these changes. Good call on moving the images, it did occur to me that this would get messy, but I didn't follow the thought through.

PR #340, opened 2017-11-21
Tweaks to the filter example

wbamberg:
* use the new Firefox logo
* reduce the number of examples to 6, to avoid scrollbars
* some minor tweaks to the styles


schalkneethling:
Once the conflict is resolved, this is a r+ 🦊

PR #347, opened 2017-11-22
Added an example for object-fit

wbamberg:


live-examples/css-examples/css/object-fit.css
```diff
@@ -0,0 +1,4 @@
+#example-element {
+  height: 100%;
+  width: 100%;
```
schalkneethling: Nit: Consistent spacing of 4 spaces in CSS

wbamberg:
>  Also, running the image through ImageOptim saves us ~20kb.

I tried using ImageOptim, and it didn't save me anything (I already ran the image through tinypng.com). Is there some non-default setting you are using to get that 20kb?

schalkneethling:
> I tried using ImageOptim, and it didn't save me anything (I already ran the image through tinypng.com). Is there some non-default setting you are using to get that 20kb?

I do not believe so but hey, let's automate it. Will add a step to the build process to optimize images.

PR #333, opened 2017-11-15
Contribute.md Guide for JS Examples

maboa:
This is the first pass of the contributing guide. I imagine it will require a few iterations of reviewing and modification :)

PR #350, opened 2017-11-22
Array examples

maboa:
I've modified these original examples quite heavily. Things have certainly evolved significantly since we started :)

That said, I think the original examples were very good just a different style and I think it's worth looking at the originals to see if we have deviated too far.

Stuff that has changed includes:

- naming of arrays now reflects the object ie `array1` etc instead of the content ie `fruit`
- fewer comments describing what we are doing (hoping that the examples are written a way  explain themselves)
- more examples where we have room (as opposed to one example)
- use of fewer lines by using`console.log`to manipulate + output ie `console.log(array1.concat(array2));` rather than ...
```
var result = array1.concat(array2);
console.log(result);
```
Where space allows, we may choose to revert to the original method of doing things? 

I think these are good test cases and so totally open to candid feedback about how in any way perhaps the originals were better. This might also affect the styleguide (https://github.com/mdn/interactive-examples/pull/333,) so good to know before we do the final pass.


live-examples/js-examples/array-includes.html
```diff
@@ -1,12 +1,15 @@
...
+
+var array2 = ['cat', 'dog', 'bat'];
 
-// log the result
-console.log(result);
+console.log(array2.includes('cat'));
 // expected output: true
+
+console.log(array2.includes('at'));
+// expected output: false
```
wbamberg: in a case like this, the numeric naming you've chosen here seems better (at least, it's hard to think of a descriptive name for the first array, and given a numeric name for the first, it seems better to have a numeric name for the second).

live-examples/js-examples/array-join.html
```diff
@@ -1,12 +1,13 @@
 <pre>
-<code id="static-js">// create an array
-var elements = ['Fire', 'Wind', 'Rain'];
+<code id="static-js">var array1 = ['Fire', 'Wind', 'Rain'];
```
wbamberg: This naming seems inconsistent with the "beasts" name above. I think I like the "beasts" choice for indexOf. Not sure what is a good descriptive name for this array though.

live-examples/js-examples/array-lastindexof.html
```diff
@@ -1,12 +1,10 @@
 <pre>
-<code id="static-js">// create an array
-var animals = ['Dodo', 'Tiger', 'Penguin', 'Dodo'];
+<code id="static-js">var array1 = ['Dodo', 'Tiger', 'Penguin', 'Dodo'];
```
wbamberg: Same here.

live-examples/js-examples/array-slice.html
```diff
@@ -1,12 +1,13 @@
...
-// call slice(), passing start and end indexes
-var sliced = a.slice(2, 4);
+console.log(array1.slice(2, 4));
+// expected output: Array ["camel", "duck"]
 
-// log the result
-console.log(sliced);
-// expected output: camel,duck
+console.log(array1.slice(1, 5));
+// expected output: Array ["bison", "camel", "duck", "elephant"]
```
wbamberg: I think it might be worth replacing one of these two examples with an example that just takes one argument.

live-examples/js-examples/array-join.html
maboa: I thought it was referring to elements of nature - as in the classical Greek elements :)

maboa: After our conversation yesterday, I'm happy making some of the naming content-descriptive. So we can use beasts and even elements. After all they are not very abstract.

live-examples/js-examples/array-slice.html
maboa: Makes sense.

PR #351, opened 2017-11-23
Added an example for "position"

wbamberg:
Here's an updated example for position.

There are a couple of things to say about this.

* I have the position properties `top` and `left` on a single line. I'm not terribly proud of this, but it means we can fit 4 examples in. I think it's acceptable here, because (1) the main property (`position`) is still on its own line, (2) the two properties are closely related, and (3) the two properties are short and have very simple syntax, so it's still quite readable.

* I've omitted `fixed`, partly because I could only fit 4 choices, but mostly because it doesn't work very well in this environment.

* it will break if the output window gets too narrow, but I think it works at the min-width for side-by-side display, so I'm hoping that would be OK.


wbamberg:
One thing I was wondering: would it be good to number the boxes? I kind of like the minimal style, but it might be useful.

live-examples/css-examples/css/position.css
```diff
@@ -1,75 +1,34 @@
...
+    top: 0;
+    left: 100px;
+    background-color: transparent;
+    border: 3px dotted red;
 }
 
-.sticky-example dd + dd {
-    border-top: 1px solid #ccc;
+#example-element-container p:first-of-type {
+  margin-top: 80px;
```
schalkneethling: Nit: 4 spaces

schalkneethling:
Love the animation effect here <3 

I don't think we need to number them, the color makes it clear which element is effected. With these changes, there are a couple additional cleanup that is needed. In `package.json` you need to update the entry for `position` to the following:

```
"position": {
            "baseTmpl": "tmpl/live-css-tmpl.html",
            "cssExampleSrc":
                "../../live-examples/css-examples/css/position.css",
            "exampleCode": "live-examples/css-examples/position.html",
            "fileName": "position.html",
            "title": "CSS Demo: Position",
            "type": "css"
        },
```

You can then also remove the folder at `live-examples/css-examples/js`

wbamberg:
Thanks @schalkneethling!

Issue #356, opened 2017-11-27
Reflect.deleteProperty produces unexpected error


maboa:
I'm seeing different errors in the interactive editor in Chrome and Firefox. Pretty sure this used to work.

```
var arr = [1, 2, 3, 4, 5];
Reflect.deleteProperty(arr, '3'); // true
console.log(arr); // [1, 2, 3, , 5]
```

On Chrome we get : `Error: Cannot read property 'constructor' of undefined`

and on Firefox we get : `Error: t is undefined`

This appears to produce the expected behaviour in browser consoles.

maboa:
We also see what appears to be the same issue, with the following example:

```
const object1 = {
  property1: 42
};

const proto1 = Reflect.getPrototypeOf(object1);

console.log(proto1);
// expected output: [object Object]

console.log(Reflect.getPrototypeOf(proto1));
// expected output: null
```

maboa:
Confirmed that this error was not present in earlier versions of the editor. Digging in ...

schalkneethling:
Tested the live examples in Firefox and Chrome and they produce the expected output now. Closing.

PR #357, opened 2017-11-28
Final JS Examples Pass

maboa:
I've now been through the all JS examples and modified code so that they adhere to the current 'guidelines'. These guidelines had evolved over time.

The `console.log` output has also become more detailed over time so `// expected output` was also brought up to date in line with that.

Note - where examples used the elements introduced post-ES6 implementation, I took the opportunity to use ES6 syntax.

In very few cases I took the opportunity to make the examples better - usually adding to them.

I've also removed some comments where I thought they were not needed - hopefully the examples will explain themselves! :)

Note: a couple of examples produce errors due to what appears to be a bug in the editor, these are detailed here https://github.com/mdn/interactive-examples/issues/356 - I'm pretty sure these worked in earlier versions of the editor, so I'm guessing we introduced the error. Happy to take a look if folk think that would be useful.


live-examples/js-examples/expressions-voidoperator.html
```diff
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">void function test() {
-  console.log("boo!");
+  console.log('boo!'');
```
wbamberg: -> console.log('boo!')

live-examples/js-examples/globalprops-nan.html
```diff
@@ -6,10 +6,10 @@
   return x;
 }
 
-console.log(sanitise("1"));
+console.log(sanitise('1'));
 // expected output: 1
```
wbamberg: This actually logs "1"

wbamberg: This actually logs "1" (this comment is for line 10 but GitHub)

live-examples/js-examples/symbol-iterator.html
```diff
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">var iterable1 = new Object();
+<code id="static-js">const iterable1 = new Object();
```
wbamberg: "expected output" in this one should be `Array [1, 2, 3]`

live-examples/js-examples/weakset-prototype-add.html
```diff
@@ -1,6 +1,6 @@
 <pre>
-<code id="static-js">var weakset1 = new WeakSet();
-var object1 = {};
+<code id="static-js">const weakset1 = new WeakSet();
+const object1 = {};
```
wbamberg: Very nitpicky, but in Firefox the output is `TypeError: WeakSet value must be an object, got the number 1`

(that is, TypeError not Error)

maboa:
OK thanks for the thorough review. I've updated in line with your comments :)

PR #358, opened 2017-11-30
Fixed mistake in array-reduce comment

galman33:
The first element of the array was missing in the comment of the array reduce example.


schalkneethling:
Thank you for your contribution @galman33

PR #359, opened 2017-12-01
Reflect issue

maboa:
Hoping this is a fix for https://github.com/mdn/interactive-examples/issues/356 (Note I have not tested ALL the examples :) )


js/editor-libs/console-utils.js
```diff
@@ -19,7 +19,9 @@ module.exports = {
                 output += this.formatArray(input[i]);
                 output += ']';
             } else {
-                output += this.formatObject(input[i]);
+              if (input[i] !== undefined) {
```
schalkneethling: Should probably just turn this into an `if else {` instead of the nested `if` inside the `else`

schalkneethling:
Closing for now as this has been stagnant for some time. Definitely needs to revisit this once we have better general test coverage.

PR #369, opened 2017-12-13
Adding an interactive example for the opacity CSS property

SphinxKnight:
Adding an interactive example for the [opacity](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity) CSS property

### DIFF
--- a/live-examples/css-examples/css/opacity.css
+++ b/live-examples/css-examples/css/opacity.css
@@ -1,8 +1,4 @@
-#example-element-container {
-    border: 2px solid;
-    border-image: linear-gradient(90deg, white, turquoise) 1;
-}
-
 #example-element {
     background-color: turquoise;
+    padding: 1em;
 }
--- a/live-examples/css-examples/css/opacity.css
+++ b/live-examples/css-examples/css/opacity.css
@@ -1,0 +1,8 @@
+#example-element-container {
+    border: 2px solid;
+    border-image: linear-gradient(90deg, white, turquoise) 1;
+}
+
+#example-element {
+    background-color: turquoise;
+}
--- a/live-examples/css-examples/opacity.html
+++ b/live-examples/css-examples/opacity.html
@@ -1,6 +1,6 @@
-<section id="example-choice-list" class="example-choice-list large" data-property="border-radius">
+<section id="example-choice-list" class="example-choice-list large" data-property="opacity">
 
-    <div class="example-choice" initial-choice="true">
+    <div class="example-choice">
         <pre><code id="example_one" class="language-css">opacity: 0;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_one">
             <span class="visually-hidden">Copy to Clipboard</span>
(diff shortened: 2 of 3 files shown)


live-examples/css-examples/opacity.html
```diff
@@ -0,0 +1,32 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="border-radius">
+
```
wbamberg: "border-radius" should be "opacity"? @schalkneethling , do we use these data attributes?

live-examples/css-examples/opacity.html
```diff
@@ -0,0 +1,32 @@
...
+    <div class="example-choice" initial-choice="true">
+        <pre><code id="example_one" class="language-css">opacity: 0;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_one">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+    <div class="example-choice" initial-choice="true">
```
wbamberg: `initial-choice` should only be set for one of the choices. Or you can omit it entirely, and it will select the first one - but in this case you probably don't want to do that :).

live-examples/css-examples/opacity.html
```diff
@@ -0,0 +1,32 @@
...
+        </button>
+    </div>
+
+
+</section>
+
+<div id="output" class="output large hidden">
+    <section id="default-example" class="default-example">
+      <div id="example-element-container">
+        <div id="example-element" class="transition-all">This is a text we might not see depending on the opacity.
```
wbamberg: I think the text here is a bit misleading, as it suggested to me that it was the text we were setting `opacity` on. Maybe just "This is an element we might not see depending on the opacity." ?

live-examples/css-examples/css/opacity.css
```diff
@@ -0,0 +1,8 @@
+#example-element-container {
+    border: 2px solid;
+    border-image: linear-gradient(90deg, white, turquoise) 1;
```
wbamberg: I found this border confusing. I do see the point of wanting to have some background element, but because the container and the example-element are the same size, it looks like the border is on example-element, then it's confusing that the rest of the element disappears but the border does not.

Having played a bit, I think it's probably best just to omit the border, and have the example-element appear and disappear.

live-examples/css-examples/css/opacity.css
```diff
@@ -0,0 +1,8 @@
...
+}
+
+#example-element {
+    background-color: turquoise;
```
wbamberg: Add a scrap of padding so the text doesn't look cramped.

live-examples/css-examples/opacity.html
schalkneethling: Ineed we do. It is read and used by the `isSupported` script but, as `opacity` is so widely supported we do not need it here. Thanks for the contribution @SphinxKnight o/\o

live-examples/css-examples/opacity.html
SphinxKnight: oh I didn't checked this one when duplicating the "unit" snippet of HTML though it's pretty obvious when you see it ^- ^ I'll amend the contribution guidelines to add this :)

SphinxKnight:
@wbamberg I think I've got the different points sorted, please tell if not :)

PR #376, opened 2017-12-23
Add margin example

wbamberg:
This is a pretty simple example, feedback welcome!

### DIFF
--- a/live-examples/css-examples/margin.html
+++ b/live-examples/css-examples/margin.html
@@ -33,7 +33,6 @@
     <span class="visually-hidden">Copy to Clipboard</span>
 </button>
 </div>
-
 </section>
 
 <div id="output" class="output large hidden">
--- a/live-examples/css-examples/margin.html
+++ b/live-examples/css-examples/margin.html
@@ -1,0 +1,47 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="margin">
+<div class="example-choice">
+<pre><code id="example_one" class="language-css">margin: 1em;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_one">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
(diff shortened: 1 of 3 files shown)


wbamberg:
This also uses very similar colors and borders to the box-sizing example. I wonder if we should try to standardise these?

schalkneethling:
> This also uses very similar colors and borders to the box-sizing example. I wonder if we should try to standardise these?

I am all for standards and less code to maintain :) Can probably create a CSS file called `examples-common.css`

live-examples/css-examples/margin.html
```diff
@@ -0,0 +1,47 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="margin">
```
schalkneethling: I am thinking that, if we know that the property is well supported given our browser support matrix, we do not have to add the data attribute.

live-examples/css-examples/margin.html
```diff
@@ -0,0 +1,47 @@
...
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_five" class="language-css">margin: 0;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_five">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
```
schalkneethling: Nit: Empty line not required

live-examples/css-examples/margin.html
wbamberg: At the moment it seems that if this is omitted, the editor assumes that the property is not supported, and just displays the static example. So I have left this in.

schalkneethling: Oh that is certainly a bug. I will look into it.

wbamberg: It should be fixed in https://github.com/mdn/interactive-examples/pull/381.

PR #377, opened 2017-12-24
Add padding example

wbamberg:
Again, happy to get feedback on this very simple example.

### DIFF
--- a/live-examples/css-examples/css/padding.css
+++ b/live-examples/css-examples/css/padding.css
@@ -1,5 +1,3 @@
-
-
 #example-element {
     border: solid 10px #FFC129;
     font: Georgia, sans-serif;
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/padding.html
```diff
@@ -0,0 +1,43 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="padding">
```
schalkneethling: Same comment here as with `margin`

live-examples/css-examples/css/padding.css
```diff
@@ -0,0 +1,6 @@
+#example-element {
+    border: solid 10px #FFC129;
+    font: Georgia, sans-serif;
```
schalkneethling: OCD: Not really using the `font` shorthand here so, should be specified as `font-family`

live-examples/css-examples/padding.html
wbamberg: As for `margin`, I can't remove this unless we change the editor logic.

PR #379, opened 2017-12-28
Update font-weight example

wbamberg:
Added some more options, changed the text to be in line with the other `font` examples. I'm happy to use a different text, if you don't like this one :).

I wonder if we should use common CSS for related examples. I saw there was some work to support this already.


live-examples/css-examples/font-weight.html
```diff
@@ -1,35 +1,49 @@
-<section id="example-choice-list" class="example-choice-list" data-property="fontWeight">
-<div class="example-choice">
+<section id="example-choice-list" class="example-choice-list" data-property="font-family">
```
schalkneethling: We can drop the data attribute here as `font-family` is widely supported so we do not need a check before showing the examples.

PR #380, opened 2017-12-28
Update font-style example

wbamberg:
Use the same text as the other `font` examples, and choose a font that has a more interesting italic style.


live-examples/css-examples/css/font-style.css
```diff
@@ -0,0 +1,4 @@
+#output section {
+    font-size: 1.2em;
+    font-family: Georgia serif;
```
schalkneethling: Missing comma between `Georgia` and `serif`?

PR #381, opened 2018-01-03
Handle omitted data-property attribute

wbamberg:
The current`isPropertySupported` code checks whether the given `dataset` is undefined, and skips the check if it is.

But actually if no `data-` attributes are given, then `dataset` will be an empty map, which will end up making the function return `false`.

This change should instead make the code check whether `dataset` contains a `property` property, and return `true` if it doesn't - so omitting `data-property` means all browsers support it.

I've also updated CONTRIBUTING.md to talk a bit about this attribute.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,7 +118,9 @@ Next, let's add the example CSS choices. Think of a few different ways that `bor
 </section>
 ```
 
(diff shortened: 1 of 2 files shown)


wbamberg:
I've manually tested this with (for example, using the `font` example):

* `data-property="font"`: displays the editor
* `data-property="fnt"`: displays the static example
* `data-property=""`: displays the static example
* no `data-property`: displays the editor
* `data-property="padding"`: displays the editor

schalkneethling:
r+ Thanks @wbamberg 🍰

PR #385, opened 2018-01-06
Fix parameter ordering Math.atan2 example

Anemy:
Small tweak:
Math.atan2 takes parameters in the order `(y, x)`.
I'm updating the example to show that:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2
I think that's how it should be at least 😅


wbamberg:
I think you are right, nice catch.

Do you think we should keep the `calcAngleDegrees` as `x,y`? Then the caller can pass in coordinates in the normal order.

Anemy:
@wbamberg Sounds good - updated

Issue #408, opened 2018-01-11
Add .taller class for JS editor


schalkneethling:
Some of the JS examples extend beyond the maximum number of lines we aim for, and the user therefore needs to scroll to see the entire example. In a lot of these cases, it is not immediately obvious that some of the code is hidden, which can lead to confusion.

To overcome this, we need to add a class that will make the editor taller, add the same class to Kuma to size the `iframe`, and update the Kumascript macro to allow passing a class parameter.

PR #409, opened 2018-01-11
Fix issue #408, adds .taller class for taller JS examples

schalkneethling:
This depends on the following, although it should be safe to merge before these are merged and deployed to production:

https://github.com/mozilla/kuma/pull/4623
https://github.com/mdn/kumascript/pull/555


wbamberg:
I think we will need more than 17 lines for "taller". Several of the examples in Proxy are at around 22 lines:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/setPrototypeOf
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/deleteProperty

One is [29](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/isExtensible)(!) but I think we can call that pathological. I feel like 23 lines should get us the vast majority of cases.

As I understand it, to use this you add "taller" to both the example HTML and the macro call? I'm sure I am missing something here, but I wondered why you couldn't just set the height of the iframe in the macro call, then have the editor adjust to use the height available? (the heading and Run/Reset buttons take up a fixed amount of space, so the editor height could be something like `calc(100%-fixedHeightOfOtherStuff)`. Like I say, I'm sure there's a good reason this doesn't work, but I had to mention it.

schalkneethling:
@wbamberg like calc(100%-fixedHeightOfOtherStuff)

Yeah, this would have been real useful but, because of the nesting this just does not work as one would hope it would.

So, the editor area, as well is the header and output, is inside a `section`, that is inside the `body` of a document that is inside an `iframe` :) So, if I say `100%` then, the browser will ask, `100%` of what? So, being pixel specific here and having to define the class in two places seems like the "best" way to go for now.

wbamberg:
@schalkneethling , do you think it would also be worthwhile to support "shorter" for pages like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length ?

schalkneethling:
> do you think it would also be worthwhile to support "shorter" for pages like 

The overhead is very low so, I would say sure. How many of these pages are there at moment?

wbamberg:
I'm not sure, but quite a few I think. I'd define "shorter" as about 7 lines or fewer.

schalkneethling:
> I'm not sure, but quite a few I think. I'd define "shorter" as about 7 lines or fewer.

I will add this to the enhancements list.

PR #411, opened 2018-01-11
Add demo for text-decoration-skip-ink

paulirish:
image

### DIFF
--- a/live-examples/css-examples/css/text-decoration-skip-ink.css
+++ b/live-examples/css-examples/css/text-decoration-skip-ink.css
@@ -1,0 +1,5 @@
+p {
+    font: 1.9em Georgia, serif;
+    color: #352995;
+    text-decoration: underline;
+}
--- a/live-examples/css-examples/text-decoration-skip-ink.html
+++ b/live-examples/css-examples/text-decoration-skip-ink.html
@@ -1,0 +1,22 @@
+<section id="example-choice-list" class="example-choice-list" data-property="textDecorationSkipInk">
+<div class="example-choice" initial-choice="true">
+<pre><code id="example_one" class="language-css">text-decoration-skip-ink: auto;</code></pre>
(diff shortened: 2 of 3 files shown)


paulirish:
awesome, thanks!

 when can i add `<div>{{EmbedInteractiveExample("pages/css/text-decoration-skip-ink.html")}}</div>` to https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink ?

I just got this error:
image

presumably gotta wait for some push.. like a few hours or a ~day? thx

wbamberg:
It should be deployed to https://interactive-examples.mdn.mozilla.net/pages/css/text-decoration-skip-ink.html (cf https://interactive-examples.mdn.mozilla.net/pages/css/text-decoration.html), pretty soon I think, although I'm not familiar with how that magic happens.

After that we will be able to include it in pages.

wbamberg:
Oh, it's up now: https://interactive-examples.mdn.mozilla.net/pages/css/text-decoration-skip-ink.html :)

paulirish:
cool! 

hmmmm. the `hidden` class is set on a bunch of it. 
compare https://interactive-examples.mdn.mozilla.net/pages/css/text-decoration.html to https://interactive-examples.mdn.mozilla.net/pages/css/text-decoration-skip-ink.html 

do you know what's up with that?

wbamberg:
I'm not sure what you mean. I see this in Chrome:

skip-ink-chrome

 ... and this in Firefox:

skip-ink-firefox

This is what I would expect - the editor does feature detection and hides the interactive bits when the browser doesn't support the property - which Firefox doesn't, in this case.

Is this not what you see, or not what you expected to see?

paulirish:
Ah cool! Ok I must have been looking at it in chrome 63. :) Cool!

PR #412, opened 2018-01-11
Fix case for titles of Expressions examples

wbamberg:
This adjusts the case of the titles used for ["Expressions and Operators"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators) examples, basically to match the title of the document.

This means:

* preserve case for keywords (`yield` not `Yield`)
* sentence case otherwise (`Operator precedence` not `Operator Precedence`)

I'm happy to discuss this if you disagree, but this seems like the best approach to me.

(I think we should do the same for CSS too - i.e. `box-shadow` not `Box Shadow`.)

### DIFF
--- a/site.json
+++ b/site.json
@@ -818,7 +818,7 @@
             "exampleCode":
                 "live-examples/js-examples/expressions-bitwiseoperators.html",
             "fileName": "expressions-bitwiseoperators.html",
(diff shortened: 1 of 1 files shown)


schalkneethling:
> I think we should do the same for CSS too - i.e. box-shadow not Box Shadow.

Completely agree!

PR #427, opened 2018-01-15
Add demo for Array.toString() and Array.unshift(), Fixes #421 and #422

### LINKED ISSUE #421: JS Examples: need example for Array.toString()
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toString

maddhruv:


live-examples/js-examples/array-unshift.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var array1 = [1, 2, 3];
+
+array1.unshift(4, 5);
+// return new length of the array i.e. 5
```
wbamberg: I think it might be worth logging this so we can show the return value ("show, don't tell"):

```
console.log(array1.unshift(4, 5));
// expected output: 5
```

What do you think?

maddhruv:
@wbamberg I have improved the example logging the _unshift_ return value

PR #429, opened 2018-01-16
Creating an example for the sepia filter function

SphinxKnight:
Shamelessly using the existing `filter` interactive example to create an example for https://developer.mozilla.org/fr/docs/Web/CSS/filter-function/sepia

### DIFF
--- a/site.json
+++ b/site.json
@@ -3272,6 +3272,13 @@
             "title": "CSS Demo: Position",
             "type": "css"
         },
+        "sepia": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "exampleCode": "live-examples/css-examples/sepia.html",
+            "fileName": "sepia.html",
+            "title": "CSS Demo: sepia()",
+            "type": "css"
+        },
         "textAlign": {
             "baseTmpl": "tmpl/live-css-tmpl.html",
             "cssExampleSrc":
--- a/live-examples/css-examples/sepia.html
+++ b/live-examples/css-examples/sepia.html
@@ -1,0 +1,27 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="filter">
+<div class="example-choice" initial-choice="true">
+<pre><code id="example_one" class="language-css">filter: sepia(60%);</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_one">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_two" class="language-css">filter: sepia(0.20);</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_two">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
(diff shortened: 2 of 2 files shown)


site.json
```diff
@@ -3272,6 +3272,13 @@
             "title": "CSS Demo: Position",
             "type": "css"
         },
+        "sepia": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "exampleCode": "live-examples/css-examples/sepia.html",
+            "fileName": "sepia.html",
+            "title": "CSS Demo: sepia()",
```
schalkneethling: The title is a bit misleading as there is no actual `sepia` function call in CSS. This begs the larger question. We do have a page for the `filter` property but, we do not cover `sepia` there. I kinda like this very specific focus on one possible use of `filter`, at the same time I wonder if this is the route we want to go consistently. I mean there is a page[https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/sepia] for `sepia` on MDN, however it is *very* sparse.

@wbamberg What are your thoughts?

wbamberg: > The title is a bit misleading as there is no actual sepia function call in CSS. 

I'm not sure I understand. There is a `sepia()` function in CSS, it's the one documented at https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/sepia, which is I assume the target for this example.

Just on the matter of the title, I think "CSS Demo: sepia()" is fine. As a rule of thumb, IMO, titles should be:

"*technology* Demo: *item*"

...where *technology* is "JS" or "CSS" or "HTML" and *item* is usually the page title (the only exceptions I can think of are the JS pages, where the titles are things like `String.prototype.foo()` but we think `String.foo()` is more useful).

Anyway, we should document that :).

>  We do have a page for the filter property but, we do not cover sepia there.

But it does cover `sepia()`: https://developer.mozilla.org/en-US/docs/Web/CSS/filter#sepia()_2. It's just one of the possible `<filter-function>` values.

>  I mean there is a page[https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/sepia] for sepia on MDN, however it is very sparse.

This is the page that the example is intended for. The way the CSS docs are structured is that CSS functions get their own pages, separate from the properties that can use them.

I think we should have interactive examples for CSS functions (and included a page in the tracker for them: https://docs.google.com/spreadsheets/d/1Qi1MHvx--6QGGFdboi4VaGVnKf0-kqxwOGMvfdAFhs0/edit#gid=191872351).

schalkneethling: > There is a sepia() function in CSS

Indeed there is but it cannot be used outside of the `filter` context right? Anyhow those are technical details :) I do agree with your reasoning here @wbamberg 

> Anyway, we should document that :)

Indeed

> But it does cover sepia(): developer.mozilla.org/en-US/docs/Web/CSS/filter#sepia()_2.

Here I was actually referring to the interactive example we currently have. It does not cover the use of `sepia` so, that I why I felt that this example made sense to have.

> This is the page that the example is intended for.

Yup. The page just seemed a little "empty" so, I was not sure if that page is really referenced much. I do however agree that having this example on there will go a long way to improving the page.

So, to be clear. I like this, support it, and think it is great ;) I just wanted to make sure we are all on the same page when it comes to things like this. With all of that, time to r+ and merge o/\o

PR #433, opened 2018-01-17
console.log support for multiple arguments

qwIvan:
such as `console.log('Hello', 'World', '!');`
image

### DIFF
--- a/js/editor-libs/console-utils.js
+++ b/js/editor-libs/console-utils.js
@@ -76,7 +76,7 @@ module.exports = {
     formatOutput: function(input) {
         'use strict';
         if (input === undefined) {
-            return input;
+            return String(input);
         } else if (typeof input === 'string') {
             return '"' + input + '"';
         } else if (Array.isArray(input)) {
--- a/js/editor-libs/console-utils.js
+++ b/js/editor-libs/console-utils.js
@@ -87,20 +87,6 @@ module.exports = {
         }
     },
     /**
-     * Concatenate the output of each argument with a blank space.
-     * @param args - Arguments of console.log
-     * @returns {*} Concatenated output as a string.
-     */
-    concatOutput: function(args) {
-        'use strict';
-        var output = '';
-        output = output + this.formatOutput(args[0]);
-        for (var i = 1, l = args.length; i < l; i++) {
-            output = output + ' ' + this.formatOutput(args[i]);
-        }
-        return output;
-    },
-    /**
      * Writes the provided content to the editor’s output area
      * @param {String} content - The content to write to output
      */
--- a/js/editor-libs/console-utils.js
+++ b/js/editor-libs/console-utils.js
@@ -87,6 +87,20 @@ module.exports = {
         }
     },
     /**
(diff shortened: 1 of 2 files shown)


js/editor-libs/console-utils.js
```diff
@@ -86,6 +86,20 @@ module.exports = {
...
+    /**
+     * Concatenate the output of each argument with a blank space.
+     * @param args - Arguments of console.log
+     * @returns {*} Concatenated output as a string.
+     */
+    concatOutput: function(args) {
+        'use strict';
+        var output = '';
+        output = output + this.formatOutput(args[0]);
+        for (var i=1; i<args.length; i++) {
```
schalkneethling: Code style nits: Change to

```
for (var i = 1, l = args.length; i < l; i++) {
```

js/editor-libs/console.js
```diff
@@ -13,8 +13,8 @@ module.exports = function() {
     };
 
     // eslint-disable-next-line no-console
-    console.log = function(loggedItem) {
-        consoleUtils.writeOutput(consoleUtils.formatOutput(loggedItem));
+    console.log = function() {
+        consoleUtils.writeOutput(consoleUtils.concatOutput(arguments));
```
schalkneethling: Perhaps the context is making this a little blurry but, it seems that this change will mean we loose the formatting done by `formatOutput`. That is definitely something we do not want to happen. Perhaps we should instead loop over the arguments, passing each one to the format function before concatenating together?

schalkneethling:
@qwIvan Thank you for your contribution. Couple of items to review. I also want to note that, unless you change your current avatar, I will not in good faith be able to merge this pull request and add you as a contributor.

The image depicts something that I feel is disrespectful to woman, and this goes against out [code of conduct](https://github.com/mdn/interactive-examples/blob/master/CODE_OF_CONDUCT.md)

qwIvan:
deleted concat function

qwIvan:
deleted concat function

qwIvan:
deleted concat function

qwIvan:
deleted concat function

schalkneethling:
@qwIvan Thank you for the updates. It is looking good. I have run into one problem where the value `undefined` is not output. Have not debugged it to see what is causing it. Feel free to look into that. I will also look into and let you know.

The example I saw that with is `weakmap-prototype-get.html`

qwIvan:
Convert undefined to String and resolved.

schalkneethling:
> Convert undefined to String and resolved.

That is indeed the simplest solution, I wonder though, do we want to output it as a string? It is not a `String` under the hood, and is generally not output as a string. With that said, the solution to this will add additional code(to loop over the final array and replace 'undefined' with `undefined` when writing the output) and complexity. I am wondering if it is worth the effort.

@wbamberg Your thoughts?

wbamberg:
I see this:

```
// input
var f;
console.log("Hello", "World", f)
```

```
// output
> "Hello" "World" undefined
```

This *looks* fine, so unless there is some subtlety I am missing, I'm happy with it :).

Issue #418, opened 2018-01-11
JS Examples: need example for Array.sort()


wbamberg:
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

We should be able to use the existing example-on-top content, ported to interactive-examples. This means converting the content to match the style guide: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md as well as embedding it in an interactive example.

diablero13:
I added example for Array.sort to the MDN [page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

wbamberg:
Thanks @maddhruv and @diablero13 !

PR #435, opened 2018-01-17
Add demo for Array.prototypes Fixes #420, #419

### LINKED ISSUE #419: JS Examples: need example for Array.splice()
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

### LINKED ISSUE #420: JS Examples: need example for Array.toLocaleString()
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toLocaleString

maddhruv:
- Array.toLocaleString() - Fixes #420 
- Array.splice() - Fixes #419

### DIFF
--- a/live-examples/js-examples/array-sort.html
+++ b/live-examples/js-examples/array-sort.html
@@ -1,0 +1,16 @@
+<pre>
+<code id="static-js">var array1 = ['March', 'Jan', 'Feb', 'Dev'];
+
+array1.sort();
+
+console.log(array1);
+// expected output: Array ["Dev", "Feb", "Jan", "March"]
+
+var array2 = [1, 30, 4, 21];
+
+array2.sort();
+
+console.log(array2);
+// expected output: Array [1, 21, 30, 4]
+</code>
+</pre>
--- a/live-examples/js-examples/array-sort.html
+++ b/live-examples/js-examples/array-sort.html
@@ -1,16 +1,12 @@
 <pre>
-<code id="static-js">var array1 = ['March', 'Jan', 'Feb', 'Dev'];
+<code id="static-js">var months = ['March', 'Jan', 'Feb', 'Dec'];
+months.sort();
+console.log(months);
+// expected output: Array ["Dec", "Feb", "Jan", "March"]
 
+var array1 = [1, 30, 4, 21];
 array1.sort();
-
 console.log(array1);
-// expected output: Array ["Dev", "Feb", "Jan", "March"]
-
-var array2 = [1, 30, 4, 21];
-
-array2.sort();
-
-console.log(array2);
 // expected output: Array [1, 21, 30, 4]
 </code>
 </pre>
--- a/live-examples/js-examples/array-splice.html
+++ b/live-examples/js-examples/array-splice.html
@@ -1,16 +1,13 @@
 <pre>
-<code id="static-js">var array1 = ['Jan', 'March', 'April', 'June'];
-
-array1.splice(1, 0, 'Feb');
+<code id="static-js">var months = ['Jan', 'March', 'April', 'June'];
+months.splice(1, 0, 'Feb');
 // inserts at 1st index position
-
-console.log(array1);
+console.log(months);
 // expected output: Array ['Jan', 'Feb', 'March', 'April', 'June']
 
-array1.splice(4, 1, 'May');
+months.splice(4, 1, 'May');
 // replaces 1 element at 4th index
-
-console.log(array1);
+console.log(months);
 // expected output: Array ['Jan', 'Feb', 'March', 'April', 'May']
 </code>
 </pre>
--- a/live-examples/js-examples/array-splice.html
+++ b/live-examples/js-examples/array-splice.html
@@ -1,0 +1,16 @@
+<pre>
+<code id="static-js">var array1 = ['Jan', 'March', 'April', 'June'];
+
+array1.splice(1, 0, 'Feb');
+// inserts at 1st index position
+
+console.log(array1);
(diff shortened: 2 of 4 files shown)


live-examples/js-examples/array-sort.html
```diff
@@ -0,0 +1,16 @@
+<pre>
+<code id="static-js">var array1 = ['March', 'Jan', 'Feb', 'Dev'];
```
wbamberg: I guess "Dev" should be "Dec". Remember to change the "expected output" line as well.

wbamberg: Also if these are months, I think we could use the descriptive variable name "months" instead of "array1".

live-examples/js-examples/array-sort.html
```diff
@@ -0,0 +1,16 @@
...
+
```
wbamberg: I think we should remove this blank line.

live-examples/js-examples/array-sort.html
```diff
@@ -0,0 +1,16 @@
...
+array1.sort();
+
```
wbamberg: I think we should remove this blank line too - then we will fit in the editor without needing a scrollbar.

live-examples/js-examples/array-splice.html
```diff
@@ -0,0 +1,16 @@
+<pre>
+<code id="static-js">var array1 = ['Jan', 'March', 'April', 'June'];
+
```
wbamberg: Could remove this blank line so we fit without a scrollbar. Could also use the variable name  "months".

live-examples/js-examples/array-tolocalestring.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">var array1 = [1, 'a', new Date()];
+
+var string = array1.toLocaleString();
```
wbamberg: I think `string` is too generic, `localeString` would be better.

live-examples/js-examples/array-tolocalestring.html
```diff
@@ -0,0 +1,9 @@
...
+
+console.log(string);
+// expected output: "1,a,17/01/2018, 15:54:20"
```
wbamberg: This output will be both timezone and locale dependent. Could we make timezone and locale fixed using something like:

```
var array1 = [1, 'a', new Date('04 Dec 1995 14:12:00 UTC')];

var localeString = array1.toLocaleString('de', {timeZone: "UTC"});

console.log(localeString);
// expected output: "1,a,4.12.1995, 14:12:00"
```

This had the added benefit of showing the parameters to `toLocaleString`.

live-examples/js-examples/array-sort.html
maddhruv: it was a typo, I fixed it.  :smiley:

live-examples/js-examples/array-tolocalestring.html
maddhruv: The timezones and locales are quite broad, so I used 'en' and 'UTC'. Would that be okay?

wbamberg: I'm not sure what you mean by "broad". Yes, it's fine to use 'en'. But you're still using `Date()` without any arguments, so the `//expected output:` will not match. I think it would be better to pass a date string to the `Date()` constructor, including the locale, so the date is fixed and will match `expected output`. Like:

```
var array1 = [1, 'a', new Date('04 Dec 1995 14:12:00 UTC')];
```

Actually, thinking about it some more, it would be great to have more than one call to `toLocaleString`, with different locales, so you can see the difference:

```
var array1 = [1, 'a', new Date('04 Dec 1995 14:12:00 UTC')];

var localeString = array1.toLocaleString('en', {timeZone: "UTC"});
console.log(localeString);
// expected output: "1,a,12/4/1995, 2:12:00 PM"

localeString = array1.toLocaleString('de', {timeZone: "UTC"});
console.log(localeString);
// expected output: "1,a,4.12.1995, 14:12:00"
```

maddhruv: from _broad_ I meant all the locales - [http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry](url).
We can't add the locales, so I used English ('en')

live-examples/js-examples/array-tolocalestring.html
```diff
@@ -1,8 +1,8 @@
 <pre>
-<code id="static-js">var array1 = [1, 'a', new Date()];
+<code id="static-js">var array1 = [1, 'a', new Date('21 Dec 1997 14:12:00 UTC')];
 var localeString = array1.toLocaleString('en', {timeZone: "UTC"});
 
 console.log(localeString);
-// expected output: "1,a,1/18/2018, 3:32:00 AM"
+// expected output: "1,a,21/12/1997, 19:42:00"
```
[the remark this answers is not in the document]
maddhruv: It seems the result of _toLocaleString_ is different on different platforms. I am using Google Chrome
screenshot from 2018-01-19 08-59-27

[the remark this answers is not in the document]
wbamberg: Oh. I guess that's because Chrome doesn't support the `options` or `locales` parameters :(.

Then perhaps the best thing would be to omit those parameters, and just add a comment like:

```
// expected output: "1,a,12/21/1997, 2:12:00 PM",
// This assumes "en" locale and UTC timezone - your results may vary
```

maddhruv:
Thanks for merging the PR!
Can you add me as a contributor to the repository, I would add all the examples.

maddhruv:
This PR also Fixes #418

Issue #416, opened 2018-01-11
JS examples: need example for Array.shift()


wbamberg:
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

We should be able to use the existing example-on-top content, ported to interactive-examples. This means converting the content to match the style guide: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md as well as embedding it in an interactive example.

wbamberg:
Sorry to reopen this, but I think there's one more thing that needs to be fixed: the variable returned from `shift()` is named `array2`, but it is not an array. It would be better to name it `element1`.

schalkneethling:
> It would be better to name it element1.

Very good point @wbamberg

wbamberg:
Fixed -> https://github.com/mdn/interactive-examples/pull/445

PR #439, opened 2018-01-21
Add demo for Array.prototypes Fixes #417 and #416

### LINKED ISSUE #417: JS examples: need example for Array.some()
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

maddhruv:


live-examples/js-examples/array-some.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id='static-js'>var array = [1, 2, 3, 4, 5];
+
+var even = function(element, index, array) {
+  // checks whether any element is even
+  return element % 2 == 0;
```
schalkneethling: Please use `===` here.

live-examples/js-examples/array-some.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id='static-js'>var array = [1, 2, 3, 4, 5];
+
+var even = function(element, index, array) {
```
schalkneethling: You are not using `index` nor `array` inside the function so, no need to declare them as parameters.

live-examples/js-examples/array-some.html
```diff
@@ -0,0 +1,12 @@
...
+  // checks whether any element is even
```
schalkneethling: // checks whether `an` element is even

schalkneethling:
Thanks @maddhruv for the update. Not sure how familiar you are with Git. Would you be ok with squashing(rebasing) your commits down to a single commit?

schalkneethling:
@maddhruv No worries, I see you worked directly on your master branch. This might be a little prickly and I do not want you to loose your work. In future, please create a feature branch instead of working directly on master.

https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow

maddhruv:
Thanks, @schalkneethling for your support. I actually didn't _rebase_

wbamberg:
> Would you be ok with squashing(rebasing) your commits down to a single commit?

@schalkneethling , I just use the "Squash and merge" big green button, so contributors don't have to squash commits. Is there a reason I shouldn't be doing this?

schalkneethling:
> Is there a reason I shouldn't be doing this?

Nope. If the person is used to doing it with Git its a quick thing. Also, trying to guide people in the direction of rebasing once they have made final changes and expect a r+

Definitely not something that should block anyone from using the Github squash and merge and holding up a PR

PR #440, opened 2018-01-22
Update break-word

CShepartd:
Update break-word according to update https://developer.mozilla.org/en-US/docs/Web/CSS/word-break and specs https://drafts.csswg.org/css-text-3/#word-break-property

### DIFF
--- a/live-examples/css-examples/word-break.html
+++ b/live-examples/css-examples/word-break.html
@@ -17,6 +17,12 @@
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
     </div>
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">word-break: break-word;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_four">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
 </section>
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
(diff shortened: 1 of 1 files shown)


schalkneethling:
@CShepartd Thank you for your contribution.

The current example content does not clearly show the difference when switching between say `normal` and `break-word`. To remedy this, I would suggest adding the following work at the end of this[1] sentence `Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu`

[1] https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/word-break.html#L24

CShepartd:
> The current example content does not clearly show the difference when switching between say normal and break-word.
Cuz FF/IE/Edge dont support it. Look in Chrome/Opera at exaple https://codepen.io/anon/pen/YYgEay
Or wait, look for "better demonstrate" at pic:
wbexample

schalkneethling:
> Cuz FF/IE/Edge dont support it.
 I did test in Chrome. You will see what I mean if you do not have your browser sized as narrowly as your screenshots show above.

CShepartd:
Ok, i get now and update example text to:
`Honorificabilitudinitatibus califragilisticexpialidocious Taumatawhakatangihangakoauauotamateaturipukakapikimaungahoronukupokaiwhenuakitanatahu 次の単語グレートブリテンおよび北アイルランド連合王国で本当に大きな言葉`

CShepartd:
Ok, i get now and updated example text

PR #445, opened 2018-01-22
Fix issue #416, rename array2 to firstElement

schalkneethling:


schalkneethling:
> and sorry to be so picky :)

Picky is good 😃

PR #448, opened 2018-01-22
Add `border-width` example

mfluehr:


live-examples/css-examples/border-width.html
```diff
@@ -0,0 +1,37 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="width">
```
wbamberg: I think this should be `borderWidth` (although it doesn't matter much in practice, since all browsers will support both properties).

mfluehr: Fixed.

wbamberg:
@mfluehr , I've updated the page: https://developer.mozilla.org/en-US/docs/Web/CSS/border-width

Looks awesome!

PR #452, opened 2018-01-23
Add interactive demo for Array.forEach(). Fixes #413

### LINKED ISSUE #413: JS Examples: need example for Array.forEach()
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

suknuk:
Adding interactive demo for Array.forEach(). Fixes #413 
I am not yet very used to contributing, so any advice or comment is appreciated :)

### DIFF
--- a/live-examples/js-examples/array-foreach.html
+++ b/live-examples/js-examples/array-foreach.html
@@ -1,0 +1,12 @@
+<pre>
+<code id="static-js">var array1 = ['a', 'b', 'c'];
+
+array1.forEach(function(element) {
+  console.log(element);
+});
+
+// expected output: "a"
+// expected output: "b"
+// expected output: "c"
+</code>
+</pre>
--- a/site.json
+++ b/site.json
@@ -88,6 +88,13 @@
             "title": "JavaScript Demo: Array.findIndex()",
             "type": "js"
         },
+        "arrayForEach": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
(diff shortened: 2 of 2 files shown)


live-examples/js-examples/array-foreach.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">const array1 = ['a', 'b', 'c'];
```
wbamberg: We should declare this with `var` not `const` (see https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6).

live-examples/js-examples/array-foreach.html
```diff
@@ -0,0 +1,13 @@
...
+array1.forEach(function(element) {
+  console.log(element);
+});
+  
+// expected output: "a"
+// expected output: "b"
+// expected output: "c"
+</code>
+</pre>
+
```
wbamberg: Please remove the spaces in line 13.

suknuk:
Thank for your review, I have changed the things you requested and amended the commit. Is this how you are supposed to change pull requests?

suknuk:
Ah I see what you mean, thanks for providing the link. I have amended the commit to include a return character at the end of the file

wbamberg:
I've updated the page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach :)

Issue #392, opened 2018-01-09
JS Examples: Need examples for Date.prototype.setUTCMinutes()


schalkneethling:
Currently we do not have an example for the `Date.prototype.setUTCMinutes()` function. This is not marked as experimental or deprecated so, we probably should have one.

Issue #391, opened 2018-01-09
JS Examples: Need examples for Date.prototype.setUTCMilliseconds()


schalkneethling:
Currently we do not have an example for the `Date.prototype.setUTCMilliseconds()` function. This is not marked as experimental or deprecated so, we probably should have one.

Issue #390, opened 2018-01-09
JS Examples: Need examples for Date.prototype.getUTCSeconds()


schalkneethling:
Currently we do not have an example for the `Date.prototype.getUTCSeconds()` function. This is not marked as experimental or deprecated so, we probably should have one.

Issue #388, opened 2018-01-09
JS Examples: Need examples for Date.prototype.getUTCMinutes()


schalkneethling:
Currently we do not have an example for the `Date.prototype.getUTCMinutes()` function. This is not marked as experimental or deprecated so, we probably should have one.

PR #459, opened 2018-01-24
Add UTC set examples for Minutes,Milliseconds and Seconds

### LINKED ISSUE #393: JS Examples: Need examples for Date.prototype.setUTCSeconds()
Currently we do not have an example for the `Date.prototype.setUTCSeconds()` function. This is not marked as experimental or deprecated so, we probably should have one.

suknuk:
Fixes #393, #392, #391, #390, #388, #387

### DIFF
--- a/live-examples/js-examples/date-setutcmilliseconds.html
+++ b/live-examples/js-examples/date-setutcmilliseconds.html
@@ -1,0 +1,12 @@
+<pre>
+<code id="static-js">var date1 = new Date('2018-01-24T12:38:29.069Z');
+
+console.log(date1.getUTCMilliseconds());
+// expected output: 69
+
+date1.setUTCMilliseconds(420);
+
+console.log(date1.getUTCMilliseconds());
+// expected output: 420
+</code>
+</pre>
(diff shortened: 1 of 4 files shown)


live-examples/js-examples/date-getutcmilliseconds.html
```diff
@@ -0,0 +1,12 @@
...
+
+console.log(date1.getUTCMilliseconds());
+// expected output: 69
+
+var date2 = new Date('2018-01-24T12:38:29.001Z');
+
+console.log(date2.getUTCMilliseconds());
+// expected output: 1
+</code>
+</pre>
```
wbamberg: Please end files with a return (this comment applies to all new files in this PR).

suknuk:
OK, I have removed the `getUTC*` examples and added the return character.

wbamberg:
Updated:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCMilliseconds
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCSeconds
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCMinutes

PR #460, opened 2018-01-24
Added examples for Array.prototype.reverse and  Array.prototype.keys

### LINKED ISSUE #415: JS Examples: need example for Array.reverse
-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse

diablero13:
#415 Add example for Array.prototype.reverse
#414 Add example for Array.prototype.keys

### DIFF
--- a/live-examples/js-examples/array-keys.html
+++ b/live-examples/js-examples/array-keys.html
@@ -1,10 +1,9 @@
 <pre>
 <code id="static-js">var array1 = ['a', 'b', 'c'];
 var iterator = array1.keys(); 
-
-console.log(iterator.next()); // expected output: Object { value: 0, done: false }
-console.log(iterator.next()); // expected output: Object { value: 1, done: false }
-console.log(iterator.next()); // expected output: Object { value: 2, done: false }
-console.log(iterator.next()); // expected output: Object { value: undefined, done: true }
+  
+for (let key of iterator){
+  console.log(key); // expected output: 0 1 2
+}
 </code>
 </pre>
--- a/live-examples/js-examples/array-keys.html
+++ b/live-examples/js-examples/array-keys.html
@@ -1,0 +1,10 @@
+<pre>
+<code id="static-js">var array1 = ['a', 'b', 'c'];
+var iterator = array1.keys(); 
+
+console.log(iterator.next()); // expected output: Object { value: 0, done: false }
+console.log(iterator.next()); // expected output: Object { value: 1, done: false }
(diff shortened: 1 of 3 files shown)


live-examples/js-examples/array-reverse.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var array1 = ['one', 'two', 'three'];
+var reversed = array1.reverse(); 
+
+console.log(array1);
+// expected output: Array ['three', 'two', 'one']
+console.log(reversed);
+// expected output: Array ['three', 'two', 'one']
+</code>
```
wbamberg: I really like that this example shows the return value as well as the in-place array :).

live-examples/js-examples/array-keys.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var array1 = ['a', 'b', 'c'];
+var iterator = array1.keys(); 
+
+console.log(iterator.next()); // expected output: Object { value: 0, done: false }
+console.log(iterator.next()); // expected output: Object { value: 1, done: false }
+console.log(iterator.next()); // expected output: Object { value: 2, done: false }
+console.log(iterator.next()); // expected output: Object { value: undefined, done: true }
+</code>
+</pre>
```
wbamberg: File should end with a return

live-examples/js-examples/array-reverse.html
```diff
@@ -0,0 +1,10 @@
...
+</pre>
```
wbamberg: File should end with a return.

live-examples/js-examples/array-keys.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var array1 = ['a', 'b', 'c'];
+var iterator = array1.keys(); 
+
+console.log(iterator.next()); // expected output: Object { value: 0, done: false }
+console.log(iterator.next()); // expected output: Object { value: 1, done: false }
+console.log(iterator.next()); // expected output: Object { value: 2, done: false }
+console.log(iterator.next()); // expected output: Object { value: undefined, done: true }
```
wbamberg: I think it would be more natural, given an iterator, to use it in a loop, like:

```
for (let key of iterator) {
  console.log(key);
}
```

...than to call `iterator.next()` directly.

(note that then the `// expected output` lines have to change, too.)

PR #462, opened 2018-01-25
Add .shorter css class for shorter JS examples

### LINKED ISSUE #436: JS Examples: Add .smaller option
As with the `.taller` class that was recently added for JS examples that needs a bit more vertical space, there are also some that does not need that much. For these we should add the `.smaller` class. This will restrict the height to show at most 7 lines of example code.

kenrick95:
Fixes #436

Preview:
capture-smaller

### DIFF
--- a/css/editable-js.css
+++ b/css/editable-js.css
@@ -72,6 +72,12 @@
     height: 490px;
 }
 
+.js .editor.smaller,
+.js .editor.smaller .CodeMirror {
+    height: 158px;
+    min-height: 158px;
+}
+
 @supports (display: flex) {
     .buttons-container {
         align-self: flex-start;
--- a/css/editable-js.css
+++ b/css/editable-js.css
@@ -72,10 +72,10 @@
     height: 490px;
 }
 
-.js .editor.smaller,
-.js .editor.smaller .CodeMirror {
-    height: 158px;
-    min-height: 158px;
+.js .editor.shorter,
+.js .editor.shorter .CodeMirror {
+    height: 200px;
+    min-height: 200px;
 }
 
 @supports (display: flex) {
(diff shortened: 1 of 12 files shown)


schalkneethling:
This looks great, thanks again @kenrick95 - I will pull this down and just do some sanity checks.

wbamberg:
I'll leave the proper review to @schalkneethling , but just one comment: in my opinion it would be better (clearer and more specific) to use "shorter" as the class name, rather than "smaller".

The other question is whether 7 lines height is the right place to put the divide between normal and "shorter". I've done an analysis of the length of JS examples here: https://github.com/mdn/interactive-examples/issues/464#issuecomment-360572522. There are 56 examples that are 8 lines long: would we rather have them in a "shorter" editor or in the current normal one?

For example:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Actually I think these pages look OK in the "normal" editor, so maybe 7 lines is just right.

schalkneethling:
> better (clearer and more specific) to use "shorter"

Yeah, I reckon that makes sense. taller and shorter

> would we rather have them in a "shorter" editor or in the current normal one?

I am thinking in shorter. It does mean that the height of the class needs to change to `200px` as apposed to the current `158px` in this PR. That still saves us `80px` over vertical real estate, and over a much larger number of examples. Thoughts?

kenrick95:
Alright, I've renamed the CSS class to "shorter" and changed the height to 200px.

schalkneethling:
Thanks @kenrick95

PR #463, opened 2018-01-25
Update math-round.html

krohne:
Added counter example showing potential problem with rounding floating point numbers


wbamberg:
Thanks for your contribution! But I think this addition goes beyond the goal of these interactive examples, which it to illustrate basic use of these APIs. If we do want to discuss this issue it would be better to do it in the main body of the page, where it's possible to explain it in a bit more detail.

PR #465, opened 2018-01-25
Add `border-color` example

mfluehr:


live-examples/css-examples/css/border-color.css
```diff
@@ -0,0 +1,6 @@
+#example-element {
+    background-color: #eee;
+    border: 4px solid;
```
wbamberg: I think it might be good to make the border wider than this, to make the color more obvious. Maybe 1em, or is that too much?

live-examples/css-examples/border-color.html
```diff
@@ -0,0 +1,37 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_two" class="language-css">border-color: red #4281de;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true" data-clipboard-target="#example_two">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_three" class="language-css">border-color: red rgb(240,30,50,.7) green;</code></pre>
```
wbamberg: `red` and `rgb(240,30,50,.7)` are too similar for the difference to be obvious, especially with this quite narrow border. I do like the fact that this demonstrates the use of `rgb()`, but it would be better to choose a more contrasting value for it.

live-examples/css-examples/css/border-color.css
wbamberg: It would be great to have a bit of padding for this - the text is squished at the top of the box. Maybe just `padding: 1em;`? This applies to the `border-width` example too, now I think of it.

wbamberg:
Page updated -> https://developer.mozilla.org/en-US/docs/Web/CSS/border-color :)

PR #472, opened 2018-01-26
Add `overflow-wrap` example

mfluehr:


wbamberg:
updated -> https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

Issue #466, opened 2018-01-25
CSS examples: need example for line-height


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

The interactive example could use for inspiration (and would ultimately replace) the "example outcome" currently in the page: https://screenshots.firefox.com/mWjoRmLKaDY4O1SS/developer.mozilla.org, although it could perhaps use some slightly more different line height values to emphasize the differences, and maybe the same text as we use in the [padding example](https://developer.mozilla.org/en-US/docs/Web/CSS/padding).

wbamberg:
Fixed -> https://github.com/mdn/interactive-examples/pull/485

PR #485, opened 2018-01-27
Add line-height CSS example

danielhickman:
Fixes: CSS examples: need example for line-height #466 

image

### DIFF
--- a/live-examples/css-examples/css/line-height.css
+++ b/live-examples/css-examples/css/line-height.css
@@ -1,0 +1,4 @@
+#example-element {
+    font-family: Georgia, sans-serif;
+    max-width: 200px;
+}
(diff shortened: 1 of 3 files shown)


danielhickman:
A few notes since I'm new:
- I used `padding` as the source.
- `.transition-all` transitions only for the lengths, should it be removed for consistency?
- The example text has the `max-width` of `200px` to make it wrap more often

wbamberg:
> .transition-all transitions only for the lengths, should it be removed for consistency?

Good question, I'm not sure. I think we should merge this anyway, but I'll raise an issue to see if anyone has strong feelings either way.

wbamberg:
I've updated the page: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height :)

Issue #477, opened 2018-01-27
 CSS examples: example needed for text-decoration-line


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-line

This could reuse most of the [text-decoration](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration) example.

PR #487, opened 2018-01-27
Add text-decoration-line CSS example

danielhickman:
Fixes: #477 reusing some code from `text-decoration`. Didn't include deprecated `blink` and set `underline` as the default.

image


danielhickman:
Nope, just thought to keep it simple but you're right, they do belong in there. I committed and squashed over my original commit. I didn't use more than 2 keywords because I got a horizontal scrollbar at 480px width (shown below), let me know if you'd like to demonstrate it anyways.

image

PR #488, opened 2018-01-27
Fixed some (all?) `data-clipboard-target` for CSS examples

danielhickman:
Easy to overlook, and I bet someone or myself will be making this oopsie down the road. Maybe should be automated or tested for?

image


wbamberg:
That's awesome work!

>  Maybe should be automated or tested for?

Yes, and I've filed https://github.com/mdn/interactive-examples/issues/489 to track this.

Issue #489, opened 2018-01-27
Test or automate `data-clipboard-target`


wbamberg:
@danielhickman fixes some errors in https://github.com/mdn/interactive-examples/pull/488 and points out that as this is an easy mistake to make, it would be very useful to test these values or to generate them automatically.

danielhickman:
It seems fairly easy to automate at runtime using some existing code. See my open PR.

PR #491, opened 2018-01-27
Dynamically get clipboard button targets

danielhickman:
addClippy will now construct a Clipboard where it finds the example code

Still can be overwritten by using a `data-clipboard-target` attribute, but that's no longer necessary.

Fixes: Test or automate `data-clipboard-target` #489

### DIFF
--- a/js/editor-libs/clippy.js
+++ b/js/editor-libs/clippy.js
@@ -1,3 +1,5 @@
+var mceUtils = require('./mce-utils');
+
 /**
  * Positions the copy to clipboard success message based on the
  * position of the button that triggered the copy event.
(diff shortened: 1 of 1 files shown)


danielhickman:
I was about to commit the removal of the attributes for testing since it works great but it wouldn't include pending PRs and touched lots of files making conflicts. But using Atom's project find and replace with this regex `data-clipboard-target="[#_\w]*"` removed only from CONTRIBUTING.md and the buttons.

This seems like a good idea to remove the existing attributes so contributors wouldn't need to worry about checking code they copy from their own or another file. Removing from the CONTRIBUTING.md file would also be smart so contributors will no longer be told to add this code in the future.

schalkneethling:
Thanks a lot for your contribution @danielhickman o/\o

wbamberg:
I think we should just remove the `data-clipboard-target` stuff from the contribution docs, don't you? @danielhickman , @schalkneethling?

It's really nice to see a reduction in the CSS 'boilerplate' :).

danielhickman:
@wbamberg I thought I made sense to remove it from both the docs and from the existing examples as the boilerplate seems to have been copied from file as it's often easier to do so.

schalkneethling:
Yup, agree 100%

danielhickman:
Okay, I added a PR for that. #508 Let me know if there are any issues as it's a bigger change. I didn't see any problems manually looking at the files changed and copying from 10 pages or so.

PR #499, opened 2018-01-29
Add `border-*-width` examples

### LINKED ISSUE #496: CSS examples: need examples for border-[top, right, bottom, left]-width
These could just clone the border-width example that @mfluehr wrote: https://developer.mozilla.org/en-US/docs/Web/CSS/border-width. The only difference would be of course that it only takes one argument, so the example could instead just focus on the different values it can take.

Hopefully it would be possible to use the [same CSS file](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/css/border-width.css), to make sure it stays consistent.

mfluehr:
Also updates the basic `border-width` example to be a little prettier and illustrate the three-value syntax. Fixes #496.

### DIFF
--- a/live-examples/css-examples/border-bottom-width.html
+++ b/live-examples/css-examples/border-bottom-width.html
@@ -1,35 +1,35 @@
-<section id="example-choice-list" class="example-choice-list large" data-property="border-width">
+<section id="example-choice-list" class="example-choice-list large" data-property="border-bottom-width">
 
     <div class="example-choice" initial-choice="true">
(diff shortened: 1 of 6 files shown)


wbamberg:
Added to pages:

https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-width
https://developer.mozilla.org/en-US/docs/Web/CSS/border-right-width
https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-width
https://developer.mozilla.org/en-US/docs/Web/CSS/border-left-width

:)

PR #500, opened 2018-01-29
Reorganize live-examples folder

### LINKED ISSUE #273: Split up site.json
The `site.json` file will end up getting huge. At some point, we might want to think about splitting it up.

kenrick95:
Fixes #273 

Right now, I only did the change for JS folder. Changes are:
- Moving .html files into its own "feature" folder
- For each feature folder, there is a "meta.json" (which is essentially the split-up site.json)
- I've also modified the root index.js to support scanning and reading the meta.json files scattered inside the folders.

I'd like to ask for feedback before continuing with CSS folder.

TODOs:

- ~~[ ] Use path relative to meta.json instead of  path relative to project root for "exampleCode" field, maybe?~~
- [x] Update contribution guide

Regards,
Kenrick


index.js
```diff
@@ -253,6 +254,14 @@ function init() {
             buildBundles(site.bundles);
             // generates the pages
             buildPages(site.pages);
+            // generated pages using glob.
+            const metaJsons = glob.sync('live-examples/**/meta.json', {});
```
schalkneethling: Nit: I would name this as `metaJSONArray`

index.js
```diff
@@ -253,6 +254,14 @@ function init() {
...
+            for (const metaJson of metaJsons) {
+                const file = fse.readJsonSync(metaJson);
```
schalkneethling: For this line and inside the loop, I would rather use `let`

index.js
```diff
@@ -253,6 +254,14 @@ function init() {
...
+                buildPages(file.pages);
+            }
+            
+
             // clean up
             removeJSBundles();
```
schalkneethling: With the above changes, it probably makes more sense to move the console statement from [1] to here:
[1] https://github.com/mdn/interactive-examples/blob/master/index.js#L136

schalkneethling:
In terms of moving forward with this, please go ahead @kenrick95

index.js
kenrick95: Hmm, why `let`? It is not like the variable will be reassigned to other thing.

schalkneethling: I generally like to reserve the use of `const` to things like config, file paths etc. Peppering them throughout a code base feels like asking for trouble ;)

It feels unnecessary to define those as constants.

kenrick95: Hmm, I feel quite the opposite because using `const` shows the intention of not reassigning the variable to something else. I will only use `let` if I'm sure I'm going to reassign this to something else. This is so that if we accidentally reassign something else to this variable, the runtime (Node in this case) will throw an error instead of silently mutating the variable.

What do you think?

schalkneethling: You convinced me :)

schalkneethling:
Oh ouch, bunch of conflicts @kenrick95 :-/

schalkneethling:
> Use path relative to meta.json instead of path relative to project root for "exampleCode" field, maybe?

@kenrick95 If the examples build and render successfully, and it is documented, I am fine with this.

kenrick95:
> Use path relative to meta.json instead of path relative to project root for "exampleCode" field, maybe?
Yeah, maybe we could do this as a separate enhancement.

----

Alright conflicts resolved :)

CONTRIBUTING.md
```diff
@@ -45,7 +45,7 @@ This will ensure that you have all the required development modules installed to
 
 ### Writing the example
 
-You start off by creating a new file inside `live-examples\css-examples\`. The name of this file should match the example you are adding. For example, if you are adding examples for [`border-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) you would call the file `border-radius.html`.
+You start off by creating a new file inside subfolder in `live-examples\css-examples\`. The name of this file should match the example you are adding. For example, if you are adding examples for [`border-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) you would call the file `border-radius.html`. On the other hand, the folder name should match the name of CSS Specification where the example belongs to. For example, `border-radius` is defined at "CSS Backgrounds and Borders Level 3", hence the example should be put at "backgrounds-and-borders" folder. This information is available at "[Specifications](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#Specifications)" section of the MDN article.
```
schalkneethling: > You start off by creating a new file inside subfolder in `live-examples\css-examples\`.

inside the subfolder `live-examples\css-examples\`.

> if you are adding examples for [`border-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) you would call the file `border-radius.html`.

you would call this file

> On the other hand, the folder name should match the name of CSS Specification where the example belongs to.

name of the CSS Specification to which the example belongs.

> For example, `border-radius` is defined at "CSS Backgrounds and Borders Level 3", hence the example should be put at "backgrounds-and-borders" folder.

is part of "CSS Backgrounds and Borders Level 3", hence the example should be created in the "backgrounds-and-borders" folder.

> This information is available at "[Specifications](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#Specifications)" section of the MDN article.

This information is available in the specifications section of the documentation on MDN. For example, for `border-radius` this can be found [here](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#Specifications).

CONTRIBUTING.md
```diff
@@ -174,15 +174,15 @@ Next, let's provide some extra styling for the example element. Create a new CSS
 
 ### Updating the metadata
 
-Next, you need to tell the page generator about your new page and its dependencies. To do this, open up the `site.json` file in the root of the project folder. Under `pages`, find an existing entry with a `type` of `css`.
+Next, you need to tell the page generator about your new page and its dependencies. To do this, open up the  To do this, open up `meta.json` at current folder (i.e. at "live-examples/css-examples/backgrounds-and-borders/meta.json").
```
schalkneethling: You are accidentally repeating `To do this, open up`

> open up `meta.json` at current folder (i.e. at "live-examples/css-examples/backgrounds-and-borders/meta.json").

open up the `meta.json` file in the current folder (i.e. "live-examples/css-examples/backgrounds-and-borders/meta.json").

CONTRIBUTING.md
```diff
@@ -215,7 +215,7 @@ Once satisfied with the example, [submit your pull request](https://help.github.
 
 ### Writing the exaple
 
-With a JavaScript example you start by creating a new `.html` file in `live-examples/js-examples`. The same naming convention applies here as it does for CSS. In this example we are going to contribute an example demonstrating the use of `Array.from` so, we'll create a new file called `array-from.html`.
+With a JavaScript example you start by creating a new `.html` file in subfolders of `live-examples/js-examples`. In this example we are going to contribute an example demonstrating the use of `Array.from` so, we'll create a new file called `array-from.html`. Since it is part of array, we're going to put it inside "array" folder.
```
schalkneethling: > With a JavaScript example you start by creating a new `.html` file in subfolders of `live-examples/js-examples`.

file in a subfolder of `live-examples/js-examples`.

> Since it is part of array, we're going to put it inside "array" folder.

Since it is part of the `Array` object, we're going to put it inside the "array" subfolder.

CONTRIBUTING.md
```diff
@@ -253,16 +253,16 @@ Please make sure the example conforms to the [JS Example Style Guide](JS-Example
 
 ### Updating the metadata
 
-All that remains is to tell the page generator about our new example. To do this, open up `site.json` at the root of the project folder. Under `pages`, find an existing entry with a `type` of `js`.
+All that remains is to tell the page generator about our new example. To do this, open up `meta.json` at current folder (i.e. at "live-examples/js-examples/array/meta.json").
```
schalkneethling: > To do this, open up `meta.json` at current folder

open up `meta.json` in the current folder

kenrick95:
Alright! I've changed the doc wordings and resolved the conflicts again. Hopefully this could be merged before any more conflicts 😅

kenrick95:
Yay! Thanks a lot too for the review 👍

wbamberg:
👏 this is an amazing bit of work @kenrick95 !

PR #501, opened 2018-01-29
Add `border-*-color` examples

### LINKED ISSUE #494: CSS examples: need examples for border-[top, right, bottom, left]-color
These could just clone the border-color example that @mfluehr wrote: https://developer.mozilla.org/en-US/docs/Web/CSS/border-color. The only difference would be of course that it only takes one argument, so the example could instead just focus on the different values it can take, such as:

* named color
* hex
* rgb
* hsl
* transparent

Hopefully it would be possible to use the [same CSS file](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/css/border-color.css), to make sure it stays consistent.

mfluehr:
Fixes #494.

### DIFF
--- a/live-examples/css-examples/border-bottom-color.html
+++ b/live-examples/css-examples/border-bottom-color.html
@@ -15,14 +15,14 @@
     </div>
 
     <div class="example-choice">
-        <pre><code id="example_three" class="language-css">border-bottom-color: rgb(170,50,220,.6);</code></pre>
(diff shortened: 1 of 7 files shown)


wbamberg:
I've updated the pages. https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-color is still showing the old example, I guess it's cached somewhere...

PR #508, opened 2018-01-30
Remove data-clipboard-target attribute

danielhickman:
Follow up to #491

It's now removed in all examples and the docs. The targets still work as intended.


danielhickman:
Awesome! #511 and any PRs after using old forks will still have them though.

Issue #450, opened 2018-01-23
SecurityError: The operation is insecure


kenrick95:
Steps to reproduce

- Go to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
- Open browser console
- Type anything at editor
- Observe the error: "SecurityError: The operation is insecure." is shown at browser console ([demo](https://giphy.com/gifs/3o752eqliSEcIob4pW/fullscreen))

I suspect that this is because the editor itself is trying to access local storage, but the editor's script origin (interactive-examples.mdn.mozilla.net) is not the same as the current page's origin (developer.mozilla.org).

schalkneethling:
@kenrick95 I was not able to replicate this. I see there is an array to the left of the error, can you expand that and post a screen shot of what is listed? Thanks a lot

FYI, we do not use `localStorage` but, we do use `postMessage`. There are origin checks on both end points so, you should never see any error, except when testing locally.

kenrick95:
Hi there, [I tried to debug using debugger](https://youtu.be/X0SsD3fiXq4) and found a call to `localStorage` [here](https://github.com/mdn/interactive-examples/blob/5e91b496edc055c9dfff8c62c01ba247a8c07574/js/editor-libs/events.js#L64). The error message expanded wasn't helpful though.

kenrick95:
Just an update, I have this setting on my Firefox that causes this console error. Apparently if I changed it to "Always", there will be no error at console.

![capture-cookies-visited](https://user-images.githubusercontent.com/3090380/35364225-dfd3ee06-01a8-11e8-932e-fd4c61300bae.PNG)

schalkneethling:
That's interesting that it is showing up that way. I have tracked down the spot where this is happening. I reckon wrapping it in a `try/catch` is the way to avoid the error being thrown. Thanks for reporting this @kenrick95

wbamberg:
Oh, I just realised that with this fix, the error is logged to the editor console, not just the browser console:

![screenshot-2018-1-31 array prototype filter](https://user-images.githubusercontent.com/432915/35663436-12182638-06d2-11e8-8b2f-f525e1e9ff3d.png)

This probably isn't what we want...

schalkneethling:
> This probably isn't what we want...

The joys of imitating the console. While I do not like swallowing errors, I think I am going to make an exception here. i.e. I am going to just silently fail.

kenrick95:
[...]

PR #509, opened 2018-01-30
Fix issue #450, wrap calls to localStorage in try/catch

schalkneethling:


wbamberg:
This looks fine, but is there any way I can test it?

I don't really understand how changing cookie settings can stop local storage from working.

schalkneethling:
> I don't really understand how changing cookie settings can stop local storage from working.

It is odd, but that is what @kenrick95 was seeing. Perhaps he can give this a go and see if it resolved the problem. @kenrick95 ?

wbamberg:
I can see the problem @kenrick95 reported on [prod](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter), if I change my cookie settings as per his comment: https://github.com/mdn/interactive-examples/issues/450#issuecomment-360319631.

In Chrome, I get a better error message:

> Failed to read the 'localStorage' property from 'Window': Access is denied for this document.

...which definitely points to localStorage.

But I can't reproduce the problem locally, so I can't test whether this fixes the problem :(.

schalkneethling:
> Maybe it would be simpler just to merge it and test afterwards :).

I am thinking that. I believe it will still fail but, at least now it will write a custom message to the console and not the error from the browser.

wbamberg:
I can confirm that this has the expected result :). With third party cookies off:

screen shot 2018-01-31 at 2 11 56 pm

PR #512, opened 2018-01-30
Adds the first iteration of HTML examples

schalkneethling:
This is the first alpha version to get HTML examples into the build pipeline. After @kenrick95 latest PR merges, I need to update this and will then open a PR for the tabbed UI for HTML examples.

### DIFF
--- a/tmpl/live-html-tmpl.html
+++ b/tmpl/live-html-tmpl.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
+    <title>%title%</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link href="../../css/codemirror-tabbed-5-31-0.css" rel="stylesheet" />
+    <link href="../../css/tabbed-editor.css" rel="stylesheet" />
+    %example-css-src%
+  </head>
+  <body>
+      <header class="output-header hidden">
+        <h4>%title%</h4>
+        <button id="reset" class="reset" type="button">Reset</button>
+        <button id="run" class="run" type="button">Run</button>
+      </header>
+      <section class="editor-container">
+<div id="html-editor" class="editor-wrapper"><pre><code>%example-code%</pre></code></div>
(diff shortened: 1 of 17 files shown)


tmpl/live-html-tmpl.html
```diff
@@ -0,0 +1,28 @@
...
+    %example-css-src%
+  </head>
+  <body>
+      <header class="output-header hidden">
+        <h4>%title%</h4>
+        <button id="reset" class="reset" type="button">Reset</button>
+        <button id="run" class="run" type="button">Run</button>
+      </header>
+      <section class="editor-container">
+<div id="html-editor" class="editor-wrapper"><pre><code>%example-code%</pre></code></div>
```
schalkneethling: Indent intentional

schalkneethling:
@wbamberg This ok to merge?

wbamberg:
@schalkneethling, sorry to be slow looking at this PR. I'll look at it today. I wasn't sure if you were going to resolve the merge conflicts before reviewing.

schalkneethling:
> I wasn't sure if you were going to resolve the merge conflicts before reviewing.

Also, this is not final in any way, shape or form. It just shows it working with the current build process. More work is waiting in the wings once this has been merged.

I wanted to break things up this way to ensure I do not hurt the contributor work flow.

wbamberg:
As I understand it, this does a few things:

* adds dependencies on the CodeMirror modes for displaying HTML
* refactors index.js into a collection of smaller modules
* adds the HTML editor interface

I've not done a detailed review of the code and am not sure I'm qualified to, but as far as I can tell this looks and works great. I tried adding a CSS file to the bundle and that worked fine too.

IMO we should show the output by default, and not require the user to press "Run". Unlike JS, but like CSS, these things tend to be highly visual, and it's helpful for people to be able to see what an element looks like without having to interact with the editor.

We should think about the wording of "Run". It doesn't seem right to me here. But we should test that.

schalkneethling:
> IMO we should show the output by default, and not require the user to press "Run".

Agree

> We should think about the wording of "Run". It doesn't seem right to me here. But we should test that.

Agree here as well. I chose run based on the convention for other interactive code editors like jsFiddle and Codepen. Also thinking ahead for when there are actual tabs for CSS and JS as well.

PR #517, opened 2018-01-31
Handle String object at formatObject

### LINKED ISSUE #293: Console.logging Object.getPrototypeOf can cause unexpected results
For example:

```
console.log(Object.getPrototypeOf('foo'));
```

Outputs `String` from browser Consoles, but doesn't output anything in the Interactive Editor.

kenrick95:
Because String object is different from string literal

Also restructure tests as nested suites

Fixes #293

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,8 @@
 module.exports = {
     env: {
         browser: true,
-        es6: true
+        es6: true,
+        jest: true,
     },
     extends: 'eslint:recommended',
     rules: {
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -1,7 +1,8 @@
 module.exports = {
     env: {
         browser: true,
-        es6: true
+        es6: true,
+        jest: true,
```
kenrick95: Note: This is to stop eslint from complaining at `console-utils.test.js` that "describe" is not defined

PR #518, opened 2018-01-31
Add `backface-visibility` example

mfluehr:


wbamberg:
I like the simplicity of this, but I think it's a bit distracting to have an animated element playing by default in the page.

I like the idea of using a 3D object to show this, and wondered if it would be possible to adapt the existing example, something like:

backface-hidden

backface-visible

This makes a few changes to the example:

* only applies the property to the back face
* makes all faces gray except the back, to highlight it
* adds a gradient to show that the faces are transparent

What do you think?

mfluehr:
Yes, your idea makes sense. I'll take another stab at it.

live-examples/css-examples/transforms/backface-visibility.css
```diff
@@ -0,0 +1,44 @@
...
+    height: 100%;
+    position: absolute;
+    backface-visibility: inherit;
+    background: rgba(0, 0, 0, 0.4);
+    font-size: 60px;
+    color: white;
+}
+
+.front {
+  transform: translateZ(50px);
```
wbamberg: nit: indentation, here and in the subsequent rules.

PR #524, opened 2018-02-01
Adding flex-wrap example

rachelandrew:
An initial example for the set of examples I'm building.


live-examples/css-examples/flexbox/flex-wrap.css
```diff
@@ -0,0 +1,14 @@
...
+    width: 80%;
+    display: flex;
+}
+
+#example-element > div {
+    background-color: rgba(0, 0, 255, 0.2);
+    border: 3px solid blue;
+    width: 100px;
+    margin: 10px;
+}
```
schalkneethling: Nit: missing empty line at the end of the file

live-examples/css-examples/flexbox/meta.json
```diff
@@ -0,0 +1,14 @@
...
+            "cssExampleSrc":
+                "../../live-examples/css-examples/flexbox/flex-wrap.css",
+            "exampleCode":
+                "live-examples/css-examples/flexbox/flex-wrap.html",
+            "fileName": "flex-wrap.html",
+            "title": "CSS Demo: flex-wrap",
+            "type": "css"
+        }
+    }
+}
```
schalkneethling: Nit: missing empty line at the end of the file

wbamberg:
This looks great, the only thing I'd say is to make sure it looks the way we want when the output window is at its minimum width.

On MDN pages the editor is laid out "side by side": that is with the example choices on the left and the output on the right, as it is in the local server that's started by `npm run start`. Then if the page width goes below some threshold it switches to "top and bottom", with the example choices above and the output below.

I'm not sure what the exact threshold is, but it seems to give us a minimum output window width of about 285 pixels. At that width the flex items wrap into a single column, which cuts off the top and bottom of the container:

screen shot 2018-02-01 at 3 29 11 pm

I wasn't sure if this was intentional, but if not perhaps it would be good to make the items narrower?

rachelandrew:
That's probably doable for this one but I can think of examples (especially in grid layout) that will be difficult to make so narrow and still useful.

Also, did I miss documentation about minimum widths? If that doesn't exist perhaps it should be added somewhere.

schalkneethling:
> Also, did I miss documentation about minimum widths? If that doesn't exist perhaps it should be added somewhere.

Very good point @rachelandrew I was thinking the exact same thing as I was reading the comments. I believe this has not been at the top of our minds as we were focus on the JS examples. I will add an issue to mention this clearly in the docs.

wbamberg:
> That's probably doable for this one but I can think of examples (especially in grid layout) that will be difficult to make so narrow and still useful.

I think that's fine, we should be pragmatic about it :).

> Also, did I miss documentation about minimum widths? If that doesn't exist perhaps it should be added somewhere.

No, I don't think it does, and yes, we should add it to the contributing guide.

rachelandrew:
ok - I've updated this for both of those issues.

PR #525, opened 2018-02-01
Fix dollar sign messing up template replacement

### LINKED ISSUE #313: incidences of $' corrupt the editor
Discovered while working on the regular expression examples, with this line:
```
var regex1 = new RegExp('^football$');
```

To reproduce create an example with the following code as a `live-examples/js-examples/test.html` - link it in the `site.json`:
```
var test = "$'abcdedghijk"
```
The code will split at `$'` and  will spill over into the area under the buttons, the `$'` will disappear - see image attached.

screen shot 2017-10-26 at 15 01 14

kenrick95:
Also TIL that the 2nd argument at String.prototype.replace has special handling of dollar signs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter

Fixes #313

### DIFF
--- a/index.js
+++ b/index.js
@@ -127,10 +127,10 @@ function buildPages(pages) {
             config.examplesDir + currentPage.type + '/' + currentPage.fileName;
         const exampleCode = fse.readFileSync(currentPage.exampleCode, 'utf-8');
 
-        // Note: Using String.prototype.replace's replacement function instead of
-        //       replacement string at 2nd argument
-        //       because replacement string has weird behavior to $ (dollar sign).
-        // More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter
+        /* Note: Using String.prototype.replace's replacement function instead of
+                 replacement string at 2nd argument
+                 because replacement string has weird behavior to $ (dollar sign).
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -125,12 +123,15 @@ function buildPages(pages) {
...
-        outputHTML = tmpl.replace(
-            '%example-code%',
-            fse.readFileSync(currentPage.exampleCode, 'utf-8')
-        );
+        const exampleCode = fse.readFileSync(currentPage.exampleCode, 'utf-8');
+
+        // Note: Using String.prototype.replace's replacement function instead of
+        //       replacement string at 2nd argument
+        //       because replacement string has weird behavior to $ (dollar sign).
+        // More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter
```
schalkneethling: Very nice catch @kenrick95 - One small nit here. Use JSDoc's multiline commenting i.e.

```
/* Note: Using String.prototype.replace's replacement function instead of
         replacement string at 2nd argument
         because replacement string has weird behavior to $ (dollar sign).
         More info: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter */
```

kenrick95: Alright, changed 😁

PR #526, opened 2018-02-01
Add `outline-*` examples

mfluehr:


live-examples/css-examples/basic-user-interface/outline-offset.html
```diff
@@ -0,0 +1,37 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_one" class="language-css">outline-offset: 4px;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_two" class="language-css">outline-offset: 0.5rem;</code></pre>
```
schalkneethling: Nit: Make this just `.5rem` for consistency across the code base.

live-examples/css-examples/basic-user-interface/outline-offset.html
```diff
@@ -0,0 +1,37 @@
...
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline-offset: 1ch;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
schalkneethling: I'm wondering, instead of having four examples that simply uses different units, perhaps you can have the first two, and then a third that is based on the example here? https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset#example

You probably only need to use the top two lines, i.e. 

```
outline: 1px dashed red;
outline-offset: 10px;
```

live-examples/css-examples/basic-user-interface/outline-offset.css
```diff
@@ -0,0 +1,7 @@
+#example-element {
+    background-color: #ffd955;
+    outline: .75em solid;
+    padding: .75em;
```
schalkneethling: perhaps also add a border around it to demonstrate the offset between the outline and the border?

live-examples/css-examples/basic-user-interface/outline-width.html
```diff
@@ -0,0 +1,37 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_three" class="language-css">outline-width: 1em;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline-width: 0.5ex;</code></pre>
```
schalkneethling: Instead of the three different units of measure, I would suggest one using px/em/rem and the other three use the  value keywords, thin, medium, and thick

live-examples/css-examples/basic-user-interface/outline.html
```diff
@@ -0,0 +1,45 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline: thick double #32a1ce;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_five" class="language-css">outline: 3mm ridge rgb(170, 50, 220, .6);
```
schalkneethling: mm is not a common unit on the web, unless you are specifying a print stylesheet so, I would recommend using px here instead.

live-examples/css-examples/basic-user-interface/outline-color.html
```diff
@@ -0,0 +1,44 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_two" class="language-css">outline-color: #32a1ce;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_three" class="language-css">outline-color: rgb(170, 50, 220, .6);</code></pre>
```
schalkneethling: The code here is invalid. You will need to change this to `rgba`

live-examples/css-examples/basic-user-interface/outline-color.html
```diff
@@ -0,0 +1,44 @@
...
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline-color: hsl(60, 90%, 50%, .8);</code></pre>
```
schalkneethling: The code here is invalid. You will need to change this to `hsla`

live-examples/css-examples/basic-user-interface/outline-style.html
```diff
@@ -0,0 +1,44 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline-style: groove;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_five" class="language-css">outline-style: inset;</code></pre>
```
schalkneethling: For both `inset` and `groove` it is very hard to see the styling as the contrast is very low. I would suggest adding `outline-color: #999;` to both this, and the `groove` examples.

live-examples/css-examples/basic-user-interface/outline.html
```diff
@@ -0,0 +1,45 @@
...
+
+    <div class="example-choice">
+        <pre><code id="example_four" class="language-css">outline: thick double #32a1ce;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code id="example_five" class="language-css">outline: 8px ridge rgb(170, 50, 220, .6);
```
schalkneethling: Note: Here you also need to switch to `rgba`

mfluehr:
@schalkneethling, I changed the background color of the box, so hopefully that addresses the contrast issue. As to the `rgb()` and `hsl()` color declarations, those are valid per the current spec. I would prefer to use that more modern notation, though I can change them if you are concerned about, e.g., compatibility.

schalkneethling:
> As to the rgb() and hsl() color declarations

@mfluehr Indeed, I would suggest using `rgba` and `hsla` for now. Thanks a lot!

wbamberg:
I've updated the outline* pages and they're looking awesome.

PR #533, opened 2018-02-02
Add transition example

wbamberg:
Here's an attempt at an example for `transition`.

It's pretty much an interactivisation of the existing example in the page: https://developer.mozilla.org/en-US/docs/Web/CSS/transition

### DIFF
--- a/live-examples/css-examples/transitions/transition-timing-function.html
+++ b/live-examples/css-examples/transitions/transition-timing-function.html
@@ -21,7 +21,7 @@
 </div>
 
 <div class="example-choice">
-<pre><code id="example_four" class="language-css">transition-timing-function: cubic-bezier(.29,1.01,1,-0.68);</code></pre>
+<pre><code id="example_four" class="language-css">transition-timing-function: cubic-bezier(.29, 1.01, 1, -0.68);</code></pre>
 <button type="button" class="copy hidden" aria-hidden="true">
     <span class="visually-hidden">Copy to Clipboard</span>
 </button>
--- a/live-examples/css-examples/transitions/transition-timing-function.html
+++ b/live-examples/css-examples/transitions/transition-timing-function.html
@@ -1,27 +1,27 @@
-<section id="example-choice-list" class="example-choice-list large" data-property="transition-property">
+<section id="example-choice-list" class="example-choice-list large" data-property="transition-timing-function">
 <div class="example-choice">
-<pre><code id="example_one" class="language-css">transition-property: margin-right;</code></pre>
+<pre><code id="example_one" class="language-css">transition-timing-function: linear;</code></pre>
 <button type="button" class="copy hidden" aria-hidden="true">
     <span class="visually-hidden">Copy to Clipboard</span>
 </button>
 </div>
 
 <div class="example-choice">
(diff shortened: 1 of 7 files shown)


wbamberg:
Actually, wait, this isn't ready. Sorry.

wbamberg:
OK, it's ready now.

I've thought before with these CSS examples actually - it would be useful sometimes for the example to be able to supply a JS function that's called when a new choice is selected. In this case it could change the transitioning properties, so the user doesn't need to hover over the output - they would just see the transition immediately on selection. Maybe there's some way to do this anyway...

live-examples/css-examples/transitions/transition.css
```diff
@@ -0,0 +1,15 @@
+#example-element {
+    background-color: #E4F0F5;
```
schalkneethling: Super Nit: lowercase e and f

live-examples/css-examples/transitions/transition.css
```diff
@@ -0,0 +1,15 @@
...
+    padding: 1rem;
+    border-radius: .5rem;
+    font: 1em monospace;
+    width: 100%;
+    transition: margin-right 2s;
+}
+
+#default-example:hover>#example-element {
+    background-color: #990099;
+    color: #FFF;
```
schalkneethling: Nits: colour can be `#909` and lowercase f

live-examples/css-examples/transitions/transition.html
```diff
@@ -0,0 +1,50 @@
...
+
+<div class="example-choice">
+<pre><code id="example_three" class="language-css">transition: margin-right 2s ease-in-out;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_four" class="language-css">transition: margin-right 2s ease-in-out 0.5s;</code></pre>
```
schalkneethling: I believe our code style is to write fractions as `.5s`, excluding the 0

live-examples/css-examples/transitions/transition-timing-function.html
```diff
@@ -0,0 +1,36 @@
...
+
+<div class="example-choice">
+<pre><code id="example_three" class="language-css">transition-timing-function: steps(6, end);</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_four" class="language-css">transition-timing-function: cubic-bezier(.29,1.01,1,-0.68);</code></pre>
```
schalkneethling: Nit, code style: spaces after commas.

schalkneethling:
> In this case it could change the transitioning properties, so the user doesn't need to hover over the output - they would just see the transition immediately on selection. Maybe there's some way to do this anyway...

I believe you can already to this by specifying a `jsExampleSrc` in `meta.json` that will then be stuck into the template here[1]

[1] https://github.com/mdn/interactive-examples/blob/master/tmpl/live-css-tmpl.html#L25

wbamberg:
Thanks @schalkneethling , I hope I got them all.

PR #537, opened 2018-02-02
Add table examples

mfluehr:
Includes `border-spacing`, `caption-side`, `empty-cells`, and `table-layout`.

### DIFF
--- a/live-examples/css-examples/table/table-layout.html
+++ b/live-examples/css-examples/table/table-layout.html
@@ -1,0 +1,31 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="table-layout">
+
+<div class="example-choice">
+<pre><code id="example_one" class="language-css">table-layout: auto;
+width: 150px;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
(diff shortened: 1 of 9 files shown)


live-examples/css-examples/table/table-layout.html
```diff
@@ -0,0 +1,31 @@
...
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_two" class="language-css">table-layout: fixed;
+width: 150px;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
```
wbamberg: I feel like we could use a couple more examples here.

I thought of:

```
table-layout: auto;
width: 100%;
```


```
table-layout: fixed;
width: 100%;
```

...to show that even when there is enough space, `auto` gives the columns relative sizes that match their content, and `fixed` doesn't.

But then also I thought just:

```
table-layout: auto;
```

...would be interesting. But there isn't enough space for all five of those, so perhaps that could be omitted.

PR #538, opened 2018-02-02
Add/update padding examples

mfluehr:
This is an attempt to add sub-property examples for `padding`. I don't know if I came up with the best solution, but I tried to make it more obvious when/where padding gets added, especially in the `padding-right` example.

### DIFF
--- a/live-examples/css-examples/basic-box-model/meta.json
+++ b/live-examples/css-examples/basic-box-model/meta.json
@@ -20,6 +20,46 @@
             "title": "CSS Demo: padding",
             "type": "css"
         },
+        "paddingBottom": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "cssExampleSrc":
+                "../../live-examples/css-examples/basic-box-model/padding.css",
+            "exampleCode":
+                "live-examples/css-examples/basic-box-model/padding-bottom.html",
+            "fileName": "padding-bottom.html",
+            "title": "CSS Demo: padding-bottom",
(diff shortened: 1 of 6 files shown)


wbamberg:
I appreciate the idea here of making the padding more obvious, but I'm concerned that it's a little abstract. It seems unusual (to me at least) to have the padding be a different background-color from the content, and this (I think) makes the examples harder to understand at a glance.

Another consequence of this is: usually I think of padding as a way to give content "breathing space", so there's that effect of crampedness with `padding:0;`, which goes away with `padding:1em;`, and I think it's helpful to show that to a beginner in the example.

So I'm not sure about the approach in the PR. But I'm hesitant to squash the work you've put into this, so I'm happy to talk about it :).

>  especially in the padding-right example.

I agree that padding-right is going to be a problem. One thing that would help a bit would be just using a text containing only short words (or `word-break: break-all;` but that's a bit extreme I think).

mfluehr:
I'm now trying a slightly different approach, with the element having a uniform background and a dashed line to represent the padding box. If this still doesn't work for you, I'll switch to your idea about using short words.

wbamberg:
This looks good, but can we swap the colors of the example element and its container, like:
screen shot 2018-02-05 at 9 46 00 am

(the container doesn't have to be that yellow, but I think I'd like the example element to be black text on a white background)

Sorry to be picky.

mfluehr:
I've added the examples to their respective MDN pages.

PR #539, opened 2018-02-02
Update class expression example

aefhm:
#### Updating Class Expression Example

Currently, we use a class declaration.

```
class Rectangle {
  constructor(height, width) {
    this.height = height;
    this.width = width;
  }
  area() {
    return this.height * this.width;
  }
}

console.log(new Rectangle(5,8).area());
// expected output: 40
```

after

```
var Rectangle = class {
  constructor(height, width) {
    this.height = height;
    this.width = width;
  }
  area() {
    return this.height * this.width;
  }
}

console.log(new Rectangle(5,8).area());
// expected output: 40
```

### DIFF
--- a/live-examples/js-examples/expressions/expressions-classexpression.html
+++ b/live-examples/js-examples/expressions/expressions-classexpression.html
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">class Rectangle {
+<code id="static-js">var Rectangle = class {
   constructor(height, width) {
     this.height = height;
     this.width = width;


aefhm:
Thanks @wbamberg.

danielhickman:
What is this for? I didn't see this in the style guide and I'm curious. Also, since `class` is ES6, `const` or `let` should be used, right?

danielhickman:
Ahh I see, that makes sense.

I think var is still okay for ES6 examples since anyone using ES6 will be exposed to it often enough that either works but was just checking.

aefhm:
Ah, I see the documentation on declaring with `let` or `const` fitting the ES6 patterns. Maybe the examples and page could be updated?

wbamberg:
I would accept a PR to change `var` to `let`[1] here. But personally I think it's OK as it is too. `var` is still a legitimate way to declare variables in some situations, and without more context, you can't say that the example should definitely use `let`.

[1] Probably not to `const` though, because one of the main differences between class expressions and class statements is that with class expressions you can redefine the variable, and if you use `const` you can't do this. At least that's my understanding, I'm not very familiar with these APIs.

aefhm:
Inertia it is. I think we get the same error `Identifier 'c' has already been declared`.

Issue #502, opened 2018-01-29
CSS examples: need examples for border-radius and associated properties


wbamberg:
We need an example for:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

Once that's done it should be possible to clone it for the corner-specific ones:

https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-left-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-right-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-left-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-right-radius

helmutgranda:
@wbamberg do you see breaking each one of the corner-specific requests being part of their own branch or can the same branch be updated?

wbamberg:
Hey @helmutgranda ! I think it would be fine to include the corner-specific properties in the same branch/PR :).

helmutgranda:
Hey @wbamberg, thanks for merging the PR, can I just keep adding the other border options to the same branch? On pull #546 you can see that it has been "closed".

wbamberg:
Sorry, I should have left it open. Please file a new PR for the other properties, in a new branch.

wbamberg:
Fixed! https://github.com/mdn/interactive-examples/pull/553

PR #541, opened 2018-02-05
Live Interactive Border Radius Example

helmutgranda:
Adding example for Border Radius:

https://github.com/mdn/interactive-examples/issues/502

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/border-radius.css
+++ b/live-examples/css-examples/backgrounds-and-borders/border-radius.css
@@ -1,0 +1,7 @@
+#example-element {
+	width: 200px;
+	height: 200px;
+	line-height: 200px;
+	background-color: #333333;
+	color: white;
+}
--- a/live-examples/css-examples/backgrounds-and-borders/border-radius.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-radius.html
@@ -1,0 +1,72 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="border-radius">
+
+	<div class="example-choice" initial-choice="true">
+		<pre><code id="example_one" class="language-css">border-radius: 10px;</code></pre>
+		<button type="button" class="copy hidden" aria-hidden="true">
+				<span class="visually-hidden">Copy to Clipboard</span>
+		</button>
+	</div>
+
+	<div class="example-choice">
+			<pre><code id="example_two" class="language-css">border-radius: 10px 5%;</code></pre>
+			<button type="button" class="copy hidden" aria-hidden="true">
+					<span class="visually-hidden">Copy to Clipboard</span>
+			</button>
+	</div>
+
+	<div class="example-choice">
+			<pre><code id="example_three" class="language-css">border-radius: 2px 4px 2px;</code></pre>
+			<button type="button" class="copy hidden" aria-hidden="true">
+					<span class="visually-hidden">Copy to Clipboard</span>
+			</button>
+	</div>
+
+	<div class="example-choice">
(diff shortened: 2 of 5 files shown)


live-examples/css-examples/backgrounds-and-borders/border-radius.css
```diff
@@ -0,0 +1,7 @@
+#example-element {
+	width: 200px;
+	height: 200px;
+	line-height: 200px;
+	background-color: #333333;
```
wbamberg: I might choose a slightly brighter background color. We use #5b6dcd for `margin` and that looks quite nice here.

live-examples/css-examples/backgrounds-and-borders/border-radius.html
```diff
@@ -0,0 +1,72 @@
...
+		</button>
+	</div>
+
+	<div class="example-choice">
+		<pre><code id="example_nine" class="language-css">border-radius: 100px;</code></pre>
+		<button type="button" class="copy hidden" aria-hidden="true">
+				<span class="visually-hidden">Copy to Clipboard</span>
+		</button>
+	</div>
+
```
wbamberg: I think the values chosen here are too subtle and too similar, so the difference between them isn't obvious enough.

Also, we should try to restrict ourselves to 6 choices here, so as to avoid vertical scrollbars. You have 9 here. Choosing 6 representative syntaxes is tricky for this one because the syntax is quite complex. I think we should aim to demonstrate:

* that you can pass a single value to get all 4 rounded corners the same
* that you can pass multiple single-radius values to get different corners with different roundednesses
* that you can pass pairs of radius-values to get elliptical corners
* that you can pass multiple pairs, to get different ellipses in different corners.

I'd probably go for something like:

```
border-radius: 30px;

border-radius: 25% 10%;

border-radius: 10% 30% 50% 70%;

border-radius: 10% / 50%;

border-radius: 20px 100px / 120px;

border-radius: 50% 20% / 10% 40%;
```

live-examples/css-examples/backgrounds-and-borders/border-radius.html
```diff
@@ -0,0 +1,72 @@
...
+</section>
+
+<div id="output" class="output large hidden">
+	<section id="default-example" class="default-example">
+		<div id="example-element" class="transition-all">Test Border Radius!</div>
```
wbamberg: To be consistent with the other border-* properties (e.g. [`border-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/border-color)), let's have the text:

> "This is a box with rounded corners."

live-examples/css-examples/backgrounds-and-borders/border-radius.html
```diff
@@ -0,0 +1,72 @@
...
+	</section>
+</div>
```
wbamberg: Files should end in a newline.

live-examples/css-examples/backgrounds-and-borders/border-radius.css
```diff
@@ -0,0 +1,7 @@
...
+	color: white;
+}
```
wbamberg: Files should end with a newline.

live-examples/css-examples/backgrounds-and-borders/border-radius.css
```diff
@@ -0,0 +1,7 @@
+#example-element {
+	width: 200px;
```
wbamberg: indentation should be 4 spaces, not tabs.

helmutgranda:
Hi @wbamberg,

Thanks for the review, I really appreciate you being very explicit other than just asking to "fix it" :). You caught me mid way moving the code to its own branch. Will submit a new PR with a branch and the fixes.

Issue #534, opened 2018-02-02
Use prefixes for CSS properties


danielhickman:
Re: [#528 > comment thread](https://github.com/mdn/interactive-examples/issues/528#issuecomment-362653145)

@danielhickman:
> [...]
Speaking of other issues, is creating another to discuss prefixing properties a good idea? I thought it would be smart for experimental technologies such as this one.

@wbamberg:
> [...]
> - how to handle prefixed properties
> - how to handle partially supported properties (i.e. when not all choices are supported)
> - how to do support detection for things that aren't properties
>
> Currently we just have data-property - it feels like we might need a more sophisticated way for an example to express the things the editor needs to test

danielhickman:
My first idea was [`mdn/browser-compat-data`](https://github.com/mdn/browser-compat-data). That's a new project, right? I honestly thought these two wonderful changes to MDN were meant to go hand in hand and I was surprised when I didn't see it used. This would take care of a lot of these issues together if implemented correctly, right?

I've always been taught to not screw around with user-agent parsing and this wouldn't work on all browsers in the same way so I'm unsure if "implemented correctly" is non-trivial or even possible.

wbamberg:
Structured browser compat data (aka BCD) is a wonderful project with lots of interesting applications both inside and outside the MDN website itself. But I'm not convinced it's appropriate here (although I could be convinced, perhaps).

I think it's worth thinking about what exactly our problem is and how we would like the editor to behave:

1. On some browsers, the examples won't work: meaning, the browser won't understand the particular CSS in the "example choice". The user won't see any effect, and will be confused or might think something is broken. In this kind of situation we'd like to explain to the user that this CSS isn't supported in the current browser, and that's why they don't see anything.

2. Sometimes this will only apply to particular choices.

3. Sometimes this will apply to the entire example (e.g. Firefox and `text-decoration-skip-ink`), and maybe we want to do something different in those cases (like not display the editor at all).

4. Sometimes the browser might need the property to be prefixed. In situations like this the example choices might contain the prefixed versions too (as in https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip) but I think we can't always rely on this.

5. Sometimes the user will edit our example choice. The result might (a) apply fine, (b) be unsupported by the browser, or (c) be invalid. It would be very helpful to distinguish these cases so we can give the user some feedback (at the moment we don't distinguish "unsupported" and "invalid" here).

[...]

PR #542, opened 2018-02-06
Implement vertical resize for JS code editor

### LINKED ISSUE #403: Idea: Make the interactive example code editor vertically resizable
To facilitate tinkering with examples in more and bigger and better ways, we should consider making the editor box for interactive examples resizable so the user can choose to have room for a larger number of lines of code than the initial sample code.

Just adding the ability to grab the bottom edge of the code box and drag it up and down to resize it would be a nice touch; have the rest of the interactive sample box and the surrounding page content move up and down as needed to make room.

kenrick95:
Fixes #403

([demo](https://giphy.com/gifs/xThtaw2UxnJBJUvMDC/fullscreen))

### DIFF
--- a/css/editable-js.css
+++ b/css/editable-js.css
@@ -3,9 +3,24 @@
     font-size: 14px;
 }
 
-.editor {
+.editor-container {
     position: relative;
     margin-bottom: 1em;
+}
+
+.editor {
+    position: relative;
+}
+
+.editor-resize-gap {
+    height: 7px;
+    position: absolute;
+    bottom: -3px;
+    cursor: s-resize;
+    width: 100%;
+    left: 0;
+    right: 0;
+    z-index: 10;
 }
 
 .output {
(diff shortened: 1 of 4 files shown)


schalkneethling:
This looks great @kenrick95 but, we have a bit of a problem. When this is embedded in a page on MDN, it lives inside an `iframe`. This means that the iframe needs to resize as the JS editor is resized. The problem is, the `iframe` is controlled by the MDN domain and not the interactive editor domain.

Currently we work around this using `pushMessage`, and we can surely do the same here to post height changes when the user stops resizing the editor. I wonder though, in its current iteration, if the amount fo code we are adding for this feature will provide a big enough payoff for end users to justify the performance impact. And then also the possible SEO impact of slower loading pages(although it might not be much, but Google seems to be cracking down ;)).

So I am on the fence about this one. @wbamberg your thoughts on this?

wbamberg:
@kenrick95 , this is really cool work, but I'm sorry to say I agree with @schalkneethling. We should have commented to this effect in the bug before you put time into your solution, and I'm sorry we didn't.

I'm not as familiar as you with the likely impact of this on code complexity or performance, but my feeling is that we should not implement this bug now.

There are IMO more important issues we should look at. One I can think of would be https://github.com/mdn/interactive-examples/issues/534, about having more a robust and sophisticated way to deal with unsupported features. But I think there too we (where "we" is everybody :) need to discuss a bit further what we want to happen, before implementing a solution.

kenrick95:
I see your points and I agree that it might not be wise to do this right now. Alright then, let's close this PR for now. Cheers.

schalkneethling:
Thanks so much for understanding and your continued amazing work @kenrick95 We really appreciate it.

PR #546, opened 2018-02-06
Adding border radius example with recommended changes.

helmutgranda:
Adding a new PR with changes recommended by @wbamberg for #502

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/border-radius.css
+++ b/live-examples/css-examples/backgrounds-and-borders/border-radius.css
@@ -1,5 +1,5 @@
 #example-element {
-    width: 200px;
+    width: 80%;
     height: 200px;
     line-height: 200px;
     background-color: #5b6dcd;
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/backgrounds-and-borders/border-radius.html
```diff
@@ -0,0 +1,52 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="border-radius">
+
+    <div class="example-choice" initial-choice="true">
+        <pre><code id="example_one" class="language-css">border-radius: 30px;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
```
wbamberg: Please indent HTML using 4 spaces, not 8 (here and elsewhere in this file).

live-examples/css-examples/backgrounds-and-borders/border-radius.css
```diff
@@ -0,0 +1,7 @@
+#example-element {
+    width: 200px;
```
wbamberg: I think we should use something like `width: 80%;` here - as it is this cuts off the end of the text.

PR #549, opened 2018-02-06
Add column examples

mfluehr:
This includes `column-count`, `column-fill`, `column-gap`, `column-rule-color`,`column-rule-style`,`column-rule-width`, `column-rule`, `column-width`, and `columns`.

### DIFF
--- a/live-examples/css-examples/multi-column-layout/column-width.html
+++ b/live-examples/css-examples/multi-column-layout/column-width.html
@@ -1,0 +1,35 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="column-width">
+<div class="example-choice">
+<pre><code id="example_one" class="language-css">column-width: auto;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_two" class="language-css">column-width: 11rem;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_three" class="language-css">column-width: 160px;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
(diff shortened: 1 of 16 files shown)


live-examples/css-examples/multi-column-layout/column-width.html
```diff
@@ -0,0 +1,35 @@
...
+
+<div class="example-choice">
+<pre><code id="example_three" class="language-css">column-width: 160px;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_four" class="language-css">column-width: 19ch;</code></pre>
```
schalkneethling: Not sure if it is a browser support issue but, clicking through these I do not see any changes to the columns on the right. Looking at browser support, it does not look like that is the issue. Perhaps this is not well suited for interactive examples, as it may rely on resizing the content output area? @mfluehr @wbamberg

live-examples/css-examples/multi-column-layout/columns.html
```diff
@@ -0,0 +1,35 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="columns">
+<div class="example-choice">
+<pre><code id="example_one" class="language-css">columns: 2;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+
+<div class="example-choice">
+<pre><code id="example_two" class="language-css">columns: 10rem 2;</code></pre>
```
schalkneethling: I wonder, could something like this be interesting here? `columns: 6rem auto;` - When you play around with the `rem` value you can see how the browser splits the content into more or less rows. Thoughts @mfluehr

live-examples/css-examples/multi-column-layout/column-width.html
mfluehr: I took another stab at this, so take a look. It's showing variety for me now in both Firefox and Chrome. (However, results do vary some depending on browser size.)

live-examples/css-examples/multi-column-layout/columns.html
mfluehr: Yes, good idea.

Issue #481, opened 2018-01-27
CSS examples: example needed for list-style-position


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position

wbamberg:
Fixed! https://github.com/mdn/interactive-examples/pull/559

PR #559, opened 2018-02-08
Add `list-style-position` css example

d7ark:
Hello. My very first pull request for mdn. I hope everything is alright. Pull's related to #481 
I've made two lists ul and ol to highlight that list-style-position works for both.

Looking forward to the feedback (especially on how to make it prettier).

### DIFF
--- a/live-examples/css-examples/lists/list-style-position.css
+++ b/live-examples/css-examples/lists/list-style-position.css
@@ -1,0 +1,6 @@
+#example-element {
+    text-align: left;
+}
+#example-element .bordered {
+    border: 1px solid #1b76c4;
+}
--- a/live-examples/css-examples/lists/list-style-position.css
+++ b/live-examples/css-examples/lists/list-style-position.css
@@ -1,6 +1,15 @@
+.default-example {
+    font-size: 1.3rem;
+}
+
 #example-element {
+    width: 100%;
+}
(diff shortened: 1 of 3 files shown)


wbamberg:
Thanks for contributing!

I think this is generally very good: however, @helmutgranda just landed an example for the shorthand of this property, `list-style`:
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style
https://interactive-examples.mdn.mozilla.net/pages/css/list-style.html

For consistency, I think it would be great if the `list-style-position` example used the same basic example as the `list-style` example: this means the same CSS file, unless that's not possible, and the same HTML contents, unless that's not possible.

@d7ark, what do you think? I'm sorry not to have made this clear in the bug.

d7ark:
It's fine. No problem. It's definitely doable.

I've even seen those changes as pull request. Should've thought of making mine similar. Ill change it and post new version.

Thanks.

schalkneethling:
> So I'm going to merge this and fix it up later, probably by moving list-style into "lists". But I'd also welcome @schalkneethling 's input :)

I agree @wbamberg

wbamberg:
I just updated the MDN page: https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position :)

PR #567, opened 2018-02-09
add css font-size examples

benji1304:
As per this issue 
https://github.com/mdn/interactive-examples/issues/563

Issue #544, opened 2018-02-06
Naming conflicts (e.g., `perspective`)


mfluehr:
I'm wanting to add an example for the `perspective` property, but it has a naming conflict with the already-existing `perspective()` function. What's the best way to handle this sort of thing?

wbamberg:
I was wondering when/if this would happen.

For the JS examples, we scoped them using the name of the "chapter", like [`typedarray-values`](https://interactive-examples.mdn.mozilla.net/pages/js/typedarray-values.html) or [`object-values`](https://interactive-examples.mdn.mozilla.net/pages/js/object-values.html).

For consistency with that, we could use the CSS module name to scope them, and this would mirror [the organization of the files in the repo](https://github.com/mdn/interactive-examples/tree/master/live-examples/css-examples). Of course that wouldn't help with the immediate problem, because `perspective` and `perspective()` are both in the same module. So perhaps we could also prefix (or suffix) functions, so we could have:

`transform-perspective`
and
`transform-perspective-function` (or `transform-function-perspective`)

What do you think? I'd also be interested to hear what @schalkneethling thinks.

schalkneethling:
Agreed, my vote would be for `transform-perspective-function`

wbamberg:
Fixed! https://github.com/mdn/interactive-examples/pull/568

PR #568, opened 2018-02-09
Use prefixes for functions and types

wbamberg:
Revewing https://github.com/mdn/interactive-examples/pull/567 I saw we already have a clash in production, between the `opacity` property and the `opacity()` function. So it's time to fix this semi-properly.

This PR:

* adds a `function-` prefix to the names of input HTML files containing examples for CSS functions
* adds a `type-` prefix to the names of input HTML files containing examples for CSS types
* updates the meta.jsons for this, and also to emit output files prefixed with `function-` or `type-`
* updates CONTRIBUTING.md to describe this rule

This is similar to the proposal in https://github.com/mdn/interactive-examples/issues/544 but a little different:

* I'm not including the name of the CSS module. This (1) doesn't really seem to add anything, practically, (2) makes all the names longer, and (3) would be a much more disruptive change. To that end I've reverted  `transform-perspective` to `perspective` and reverted `transform-perspective-function` to `function-perspective`

* I'm prefixing, not suffixing. The immediate reason for this is that some CSS property names end "-function", like `animation-timing-function`. The other reason is that it collects the functions together when they're listed alphabetically, which seems like a useful thing.


CONTRIBUTING.md
```diff
@@ -201,6 +201,41 @@ Under `pages`, copy and paste the example then update it to apply to your new ex
...
+* the name of the HTML file you write must be prefixed with `function-` for functions, or `type-` for types.
+* in the meta.json file, the name of the output HTML file must be prefixed in the same way.
+
+So the meta.json entry for a function would look like:
+
+```
+"translateX": {
+    "baseTmpl": "tmpl/live-css-tmpl.html",
+    "cssExampleSrc":
+        "../../live-examples/css-examples/transforms/translate.css",
```
schalkneethling: Nit: Should this also be `../../live-examples/css-examples/transforms/translateX.css`?

wbamberg: I don't think so - I think `translate`, `translateX`, and `translateY` all use the comment `translate.css`. Don't they?

schalkneethling: Doh! I was looking at the wrong entry when making the comment ;)

PR #570, opened 2018-02-10
CSS examples: example for z-index

### LINKED ISSUE #554: CSS examples: example needed for z-index
https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

veekas:
This updates the interactive demo for the z-index page. It is meant to somewhat mirror the [current example on the z-index page](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index).

## Current example on MDN
current example

## Proposed, interactive example
proposed example

Unfortunately, my linter added 4-space indents to the html that wasn't indented already. Sorry if that makes it more difficult to review the code.

Closes #554


live-examples/css-examples/transitions/z-index.html
```diff
@@ -1,38 +1,59 @@
...
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">z-index: 3;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
wbamberg: This choice is a duplicate of the one before (both `z-index: 3;`). I think you could just remove one of them.

veekas: Whoops! Fixed.

Issue #565, opened 2018-02-09
CSS examples: example needed for width


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/CSS/width

The page already contains an example that we should be able to interactivise:

<img width="1053" alt="screen shot 2018-02-08 at 10 21 47 pm" src="https://user-images.githubusercontent.com/432915/36014409-7caa72ce-0d1e-11e8-977e-1fef095ac23a.png">

We should omit any values that don't have good cross-browser support though.

dominicduffin1:
Am I correct in thinking that the interactivised example should only include the blue bits?

wbamberg:
Yes, kind of. I think the output pane should contain one blue block then each choice in the left-hand pane should be the contents of one of the blocks in that example, like:

```
width: 150px;
```

```
width: 20em;
```

```
 width: 75%;
```

```
...and so on.
```

At least, that seems like a reasonable approach for this one :).

dominicduffin1:
That sounds reasonable :) I'll have a go at that.

PR #571, opened 2018-02-10
CSS interactive examples: Add example for width.

dominicduffin1:
I've created the interactive example based on the four blue boxes as per #565 

This is my first ever pull request so I hope everything's alright. I'm pleased to take any feedback.

### DIFF
--- a/live-examples/css-examples/basic-box-model/width.css
+++ b/live-examples/css-examples/basic-box-model/width.css
@@ -1,4 +1,5 @@
 #example-element {
     background-color: #E4F0F5;
     height: 80%;
+    justify-content: center;
 }
--- a/live-examples/css-examples/basic-box-model/width.css
+++ b/live-examples/css-examples/basic-box-model/width.css
@@ -1,0 +1,4 @@
+#example-element {
+    background-color: #E4F0F5;
+    height: 80%;
+}
--- a/live-examples/css-examples/basic-box-model/width.css
+++ b/live-examples/css-examples/basic-box-model/width.css
@@ -1,6 +1,8 @@
 #example-element {
-    background-color: #E4F0F5;
+    display: flex;
+    flex-direction: column;
+    background-color: #5b6dcd;
     height: 80%;
     justify-content: center;
-    color: #000000;
+    color: #ffffff;
 }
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/basic-box-model/width.css
```diff
@@ -0,0 +1,6 @@
+#example-element {
+    background-color: #E4F0F5;
```
wbamberg: Style nit: use lowercase for the letters: `#e4f0f5`.

@schalkneethling , is there a style guide we can link from the contributing docs? It seems unfair to complain about something we don't document.

live-examples/css-examples/basic-box-model/width.css
```diff
@@ -0,0 +1,6 @@
...
+    height: 80%;
+    justify-content: center;
+    color: #000000;
+}
```
wbamberg: It would be nice to vertically-center the text. You could borrow what the [border-radius example](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/backgrounds-and-borders/border-radius.css) does, by adding:

```
 display: flex;
 justify-content: center;
 flex-direction: column;
```

You *might* like to use the same color scheme that border-radius uses, too. But that's entirely up to you.

live-examples/css-examples/basic-box-model/width.html
```diff
@@ -0,0 +1,37 @@
...
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+</section>
+
+<div id="output" class="output large hidden">
+    <section id="default-example" class="default-example">
+      <div id="example-element" class="transition-all">This is a light blue box</div>
```
wbamberg: Other similar examples have text like ["This is a box with a border around it."](https://developer.mozilla.org/en-US/docs/Web/CSS/border-width). For consistency, you could change this text to "This is a box whose width you can change.", or something like that.

dominicduffin1:
Those changes seem fine :) I'll add them fairly soon.

live-examples/css-examples/basic-box-model/width.css
dominicduffin1: Yes, the colour scheme on border-radius _is_ better.

PR #578, opened 2018-02-12
Add/update background examples

### LINKED ISSUE #562: CSS examples: example needed for background
https://developer.mozilla.org/en-US/docs/Web/CSS/background

This is quite a complex one I think. It might be worth looking at some of the related examples we've already done, like: https://developer.mozilla.org/en-US/docs/Web/CSS/background-image

It would also be worth writing examples for the other background-* properties at the same time, as they could share a lot of the design.

mfluehr:
Fixes #562. This adds all the `background-*` properties that are not yet created. It also makes various updates to some existing ones.

The moon image is taken from https://commons.wikimedia.org/wiki/File:Moon.jpg, which is in the public domain.


live-examples/css-examples/backgrounds-and-borders/background-position-x.html
```diff
@@ -37,7 +37,7 @@
 </section>
 
 <div id="output" class="output large hidden">
-    <section id="default-example" class="display-block">
+    <section id="default-example">
```
wbamberg: This example, and the other two background-position examples, don't seem to work for me any more: I don't see the star or the background colour. It seems to be the removal of `class="display-block"` that's done this. Do you see that as well?

live-examples/css-examples/backgrounds-and-borders/background-attachment.html
```diff
@@ -0,0 +1,44 @@
...
+</div>
+</section>
+
+<div id="output" class="output hidden">
+    <section id="default-example">
+          <div id="example-element">
+London. Michaelmas term lately over, and the Lord Chancellor sitting in Lincoln's Inn Hall. Implacable November weather. As much mud in the streets as if the waters had but newly retired from the face of the earth, and it would not be wonderful to meet a Megalosaurus, forty feet long or so, waddling like an elephantine lizard up Holborn Hill.
+          </div>
+    </section>
+</div>
```
wbamberg: This is a really nice example :).

live-examples/css-examples/backgrounds-and-borders/background.html
```diff
@@ -0,0 +1,44 @@
...
+</button>
+</div>
+
+<div class="example-choice" initial-choice="true">
+<pre><code class="language-css">background: center / contain no-repeat url("../../media/examples/firefox-logo.svg"),
+            #cf4 35% url("../../media/examples/lizard.png");</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
```
wbamberg: This one is a bit...lively. I wonder if we should tone down the colours a bit. But that's just a subjective opinion.

live-examples/css-examples/backgrounds-and-borders/background-position-x.html
mfluehr: Yes, sorry about that.

PR #587, opened 2018-02-15
Increase padding for code area

mfluehr:
Sometimes CSS code examples overflow horizontally, When this happens, part of the code is covered by the copy button. For example, see https://developer.mozilla.org/en-US/docs/Web/CSS/background-image.

To fix this, this PR adds some more padding to the right side of the `<pre>` elements.

### DIFF
--- a/css/editable-css.css
+++ b/css/editable-css.css
@@ -70,7 +70,7 @@
 }
 
 .live .example-choice pre {
-    padding: 0 10px;
+    padding: 0 40px 0 10px;
 }
 
 .live .example-choice code {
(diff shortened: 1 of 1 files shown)


wbamberg:
This is definitely a good idea. What do you think of restricting it to the selected choice, with something like:

```
.live .example-choice.selected pre {
    padding: 0 40px 0 10px;
}
```

...since the "Copy" button only shows for the selected choice, and then you get your 40 pixels back for the choices that aren't selected (but that you still might want to read).

PR #589, opened 2018-02-15
Make container scroll for background-attachment example

wbamberg:
It's a shame that the [background-attachment example](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment) doesn't distinguish between `scroll` and `fixed` (compare, for example, https://codepen.io/tjacobdesign/pen/wBeXZz). I had hoped it would work, but it doesn't, presumably because it's in an iframe.

This PR makes the container scroll, so if you scroll the container, you can see the difference between the two values. What do you think, @mfluehr ? Maybe there's a better approach?

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/background-attachment.css
+++ b/live-examples/css-examples/backgrounds-and-borders/background-attachment.css
@@ -1,5 +1,13 @@
+#output {
+  overflow: scroll;
+}
+
+#default-example {
+  height: 600px;
+}
+
 #example-element {
-    max-width: 30rem;
+    max-width: 20rem;
     height: 100%;
     background: url(/media/examples/lizard.png) right 3rem top 1rem / 15rem no-repeat,
                 url(/media/examples/moon.jpg) center / 10rem;
--- a/live-examples/css-examples/backgrounds-and-borders/background-attachment.css
+++ b/live-examples/css-examples/backgrounds-and-borders/background-attachment.css
@@ -1,9 +1,9 @@
 #output {
-  overflow: scroll;
+    overflow: scroll;
 }
 
 #default-example {
-  height: 600px;
+    height: 600px;
 }
 
 #example-element {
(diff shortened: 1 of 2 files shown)


mfluehr:
I can't think of a better solution. Maybe the `overflow` property should be `scroll` to eliminate the horizontal scroll bar?

The scroll bars perhaps add unnecessary complication to most of the examples. Can you use `height: 600px;` only in the `fixed` example?

wbamberg:
>  Can you use height: 600px; only in the fixed example?

Except that we want to be able to contrast `fixed` with `scroll`, so we would want it for `scroll` as well. But I do agree that it's not ideal. I wonder if @schalkneethling has any bright ideas here :).

schalkneethling:
@wbamberg I can poke at it a bit but, I don't feel that it should be a blocker for now. We can always improve later if we think of/stumble upon a better solution.

schalkneethling:
@wbamberg I reckon this is good and how we need to handle this within the `iframe`. One thing though. I personally feel the background is to noisy with the moons and the lizard. Could we go with just the lizard?

schalkneethling:
@wbamberg Thoughts on the above ^^

wbamberg:
> One thing though. I personally feel the background is to noisy with the moons and the lizard. Could we go with just the lizard?

It needs two backgrounds, to show the effect of `local, scroll` and `scroll, local`. I didn't really think it was too busy when I reviewed it, and it's consistent with the other background-* examples. I agree that it might not be what you choose for a real world page, but as a demo it seems to work well.

I'm on the fence about this PR really. I think it's a shame we can't show the different between `scroll` and `fixed` in the current example (especially as the docs themselves are a bit impenetrable, and an example really helps you to understand it), but the approach in this PR feels quite suboptimal too, and I agree with @mfluehr that the double scrollbar looks pretty bad :(.

schalkneethling:
I am going to go ahead and merge this. Even though this may or may not be the optimal solution, it at least allows these items to be demonstrative. Thx @wbamberg

PR #592, opened 2018-02-16
Add border-image examples

### LINKED ISSUE #575: CSS examples: examples needed for border-image-*
We already have an example for the shorthand property: https://developer.mozilla.org/en-US/docs/Web/CSS/border-image.

But it would be worth:
* removing the comment lines
* adding more choices (without comment lines we would have space for 6 choices

I haven't looked in detail at whether we could design better example choices for this one.

Then we need an example for each of the sub-properties:
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-source
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-slice
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-width
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-outset
https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-repeat

Ideally they'd share the same basic setup, for consistency.

mfluehr:
Fixes #575.


wbamberg:
I haven't reviewed the code, just played with these examples. I think they are nice examples, work well, and I'd be happy to merge them, subject to the code being good (which it probably is).

But... I think I now understand [your comment](https://github.com/mdn/interactive-examples/issues/575#issuecomment-364943943):

> border-image examples are very hard to understand without seeing the source image and knowing its dimensions

Part of the problem is that the dimensions given in border-image-slice aren't visible in the #example-element, because they're applied to the source image and scaled.

I wonder if it would be feasible to include the source image in the example output, alongside (or underneath) the example-element, and draw thin red dashed guidelines showing how the source image is sliced. I don't think it would need to show the actual dimensions, because at least the user would be able to see how they change when they change the CSS.

But I'm not sure whether this is a good idea, for a couple of reasons:
* it would be quite complicated to implement
* it might look too complicated and "fussy"

What do you think? If you don't think this is a good plan I'll be happy to review these examples as-is.

media/examples/border-florid.svg
```diff
@@ -0,0 +1 @@
...,1.38-1.4-1.41-2-3.34-3.45-4.52-3-2.38-7.47-1.36-7.64,2.85-.07,1.57.3,3,2,3.63s2.25-.68,3.71-.68c3.58,0,.71,2.73-1.41,3.5-6.38,2.28-8.37-7.34-6.5-11.41a14.6,14.6,0,0,1,19.07-7.3c3.06,1.37,6.65,3,9.07,5.41,1.24,1.22,5.11,3.55,6.48,4.72C243.83,66.15,244.17,70.75,244.78,73.53Z" transform="translate(0 0)"/><path class="b" d="M220.26,46.67c-2.7-.41-4.55-2-6.49-4-1-1-2.16-3-3.48-3.55s-3.25.78-3.87-.48,1.83-2.44,2.36-3.3c.79-1.28.84-2.89,1.25-4.28,2.27-7.63,10.37-12.28,6.86-21C215.9,7.66,213,7,212,5c-1.57-3.18,3.47-.93,5,.29,3.12,2.45,4.21,6.26,3.67,10.26-.08.62-1.79,4.2-1.49,4.72s3.78-.88,5.13-1c4.79-.49,11.12,1.45,15.53-1,2-1.09,1.59-.13.19-1.27-1.78-1.43-4.11-1.33-5.45-3.62C229.91,5.41,241,.91,246.75.23,258.12-1.12,271.38,3.45,278.42,13c3.17,4.29,5.22,8.21,5.56,13.52.24,3.62-2,15.27-7.87,10-2.34-2.1-.87-2.67,1.53-3,2.57-.38,2.77.22,3.94-2.3,1.79-3.84.1-8.82-3.76-10.44-7.22-3-17.7,1-21.54,6.9s-8.46,9.47-14.1,6.51c-3-1.6-5.05-4.53-4.33-8.16.19-1,3.14-6,2.94-6-1.79,0-4.11.84-6,1.13-2.88.45-5.67.23-8.56.44a12.55,12.55,0,0,0-9,4.22,13.14,13.14,0,0,0-2.8,12.4c.7,2.52,2.84,5.79,5.56,6.46,3.55.87,4.93-2.14,7.62-3.34l.19.8C227.86,45.78,223,47.09,220.26,46.67Z" transform="translate(0 0)"/></svg>
```
schalkneethling: Nit: missing empty line at end of file

mfluehr:
@wbamberg, I like the idea of showing the slices with dotted lines. You're right that it would be tricky to implement. Also, it would require having the example scroll, which isn't ideal. Nevertheless, I think it would make the examples (especially -slice) a lot more useful.

That said, it may take me a bit to find time to do this. In the meantime, what do you think about merging these examples as they are? And maybe create an issue for version 2?

PR #594, opened 2018-02-19
Add example for list-style-type CSS property

### LINKED ISSUE #483: CSS examples: example needed for list-style-type
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type

It should be possible to reuse most of the existing example (https://screenshots.firefox.com/lShH53qBOC6sS6eU/developer.mozilla.org).

We'd need to give some thought to the `@counter-style` option. If we want to include it, we could perhaps move the *"custom-emoji" Assume the following rule: ...* bit into the output pane of the example, and maybe reword it a little:

> With `custom-emoji` selected, the following `@counter-style` rule is applied: ...

ddbeck:
This PR adds an example for the [`list-style-type`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) CSS property. This is my first attempt at an example (🎉) so I'm open to any and all input on this. One area in particular I'm concerned about is the appearance of the `@counter-style` definition.

This ought to fix #483 (which looked unclaimed to me, at the moment).

### DIFF
--- a/live-examples/css-examples/lists/list-style-type.html
+++ b/live-examples/css-examples/lists/list-style-type.html
@@ -1,5 +1,12 @@
 <section id="example-choice-list" class="example-choice-list large" data-property="list-style">
     <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">list-style-type: space-counter;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
         <pre><code class="language-css">list-style-type: disc;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
                 <span class="visually-hidden">Copy to Clipboard</span>
@@ -15,13 +22,6 @@
 
     <div class="example-choice">
         <pre><code class="language-css">list-style-type: "\1F44D"; // thumbs up sign</code></pre>
-        <button type="button" class="copy hidden" aria-hidden="true">
-                <span class="visually-hidden">Copy to Clipboard</span>
-        </button>
-    </div>
-
-    <div class="example-choice">
-        <pre><code class="language-css">list-style-type: space-counter;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
                 <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/lists/list-style-type.html
+++ b/live-examples/css-examples/lists/list-style-type.html
@@ -36,22 +36,18 @@
                 <li>Apollo</li>
                 <li>Hubble</li>
                 <li>Chandra</li>
-                <li>Cassini-Huygens</li>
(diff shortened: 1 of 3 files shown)


wbamberg:
This looks good, and I love space-counter, but I agree that the presentation of `@counter-style` is not ideal. One concrete issue is that it will overflow horizontally even at quite generous widths.

I would suggest instead of side-by-side, having the output and code top-and-bottom. You'll need to add `flex-direction: column;` to .output.section:

You'll also need to compress the content vertically. I would do things like:
* trim some vertical margins
* remove two of the list items (3 should be enough, just)
* reformat the text a bit: I think you could probably omit the `<p><code>space-counter</code> is defined as:</p>` bit, possibly remove `system`

I'd also have an `<hr>` to divide the output from the code. Something like:

screen shot 2018-02-19 at 5 24 51 pm

What do you think? It's sad to only have 3 list items, I tried putting all the symbols on one line, but they'll be cut off at narrower widths.

wbamberg:
Also, I think space-counter is the best-looking example, so I would make that the initial choice.

ddbeck:
I like these suggestions, @wbamberg! I've resized the example like you suggested and space-counter has blasted off to the initial choice. Let me know if you want to see any other changes. Thanks!

wbamberg:
This looks great. I noticed that the change introduced a mismatch between the symbols listed in the HTML and the actual symbols in the CSS.

I also realised that by reducing the font to `1.2rem` we could fit 4 list items in, which I liked, and that the example would look better without the `background-color` on the list items (this is only there to help demonstrate `list-style-position`, which we are not doing here). So I made those changes.

I hope you don't mind me pushing them to your branch - it's just to stop you having to go round another review cycle, for things I really should have picked up the first time.

wbamberg:
I've updated the page: https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type :)

wbamberg:
@ddbeck , looking at this example, I'd be tempted, now, to move the whole `@counter-style` bit out of the example and just have it in the MDN page underneath. The main disadvantage to this is that it means the `@counter-style` CSS itself and the documentation is it are further apart, so it's easier for them to inadvertantly diverge. But I think the example would have more space (no pun intended).

What do you think? I'm sorry to be indecisive about it, and I'd be happy to make the change, unless you wanted to.

ddbeck:
@wbamberg Yeah, I understand where you're coming from. I don't think moving the `@counter-style` out of the example is a bad idea, though I'm a little reluctant to have part of the example outside of the repository.

Before settling on that, I wonder what you'd think of this idea. Instead of showing how `@counter-style` works at all, we could just assume that it does. Like this, maybe?

screen shot 2018-02-21 at 7 43 15 pm

That would buy us a lot of… room and puts off the complexity to the page for `@counter-style`.

wbamberg:
> Instead of showing how @counter-style works at all, we could just assume that it does. Like this, maybe?

I like this!

ddbeck:
OK, I'll open a new PR with the new look!

Issue #555, opened 2018-02-07
CSS examples: example needed for cursor


brianlmacdonald:
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

I'll take this one!

PR #597, opened 2018-02-20
feat: adds cursor interactive examples

brianlmacdonald:
Addresses #555 

Adds cursor examples and a fun light blue div to mouse over, and in doing so, change your cursor's style.


live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,15 @@
+.example-choice-list {
+  height: 500px;
+  width: 250px;
+}
```
wbamberg: I don't think we should be adjusting the editor styles in this CSS (but actually, you don't need this anyway).

live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,15 @@
...
+
+.cursor {
+  display: flex;
+  background-color:#1766aa;
```
wbamberg: Should have a space here between the `:` and the value.

live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,15 @@
+.example-choice-list {
+  height: 500px;
+  width: 250px;
+}
+
+.cursor {
```
wbamberg: I'm not sure why we don't simply apply this CSS to `#example-element`.

live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,15 @@
...
+  display: flex;
+  background-color:#1766aa;
+  color: white;
+  height: 100px;
+  width: 200px;
```
wbamberg: I'd suggest making this bigger, like 80% / 80%.

live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,15 @@
...
+  justify-content: center;
+  align-items: center;
+  font-size: 14pt;
+}
```
wbamberg: We use 4 spaces to indent (sorry not to have a style guide linked from the contribution docs).

live-examples/css-examples/cursor/cursor.html
```diff
@@ -0,0 +1,257 @@
+<section id="example-choice-list" class="example-choice-list" data-property="cursor">
+  <div>
+  <div class="example-choice" initial-choice="true">
+    <pre><code class="language-css">cursor: auto;</code></pre>
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
```
wbamberg: We should restrict the number of choices to 6, to avoid scrolling. Which 6 values you pick is up to you, but try to choose a representative/commonly used set. I know it's hard in a case like this when there are so many values.

live-examples/css-examples/cursor/cursor.html
```diff
@@ -0,0 +1,257 @@
...
+
+</section>
+
+<div id="output" class="output hidden large">
+
+  <section id="default-example" class="default-example container">
+    <div class='cursor 'id="example-element">Mouse over to see cursor style</div>
+  </section>
+
+</div>
```
wbamberg: File should end with a newline.

live-examples/css-examples/cursor/cursor.html
```diff
@@ -0,0 +1,257 @@
...
+    </button>
+  </div>
+</div>
+
+</section>
+
+<div id="output" class="output hidden large">
+
+  <section id="default-example" class="default-example container">
+    <div class='cursor 'id="example-element">Mouse over to see cursor style</div>
```
wbamberg: If you make the change I suggested in the CSS file you don't need this `'cursor'` class.

wbamberg: In [similar examples](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius) we end the sentence with a period and refer to the element explicitly, like: "Move over this element to see the cursor style."

live-examples/css-examples/cursor/cursor.html
brianlmacdonald: @wbamberg done and done. Thanks!

wbamberg:
>  would a eslint-config for mdn's style be something useful/helpful? 

Yes, very! @schalkneethling , is there a style guide that @brianlmacdonald could use for this?

live-examples/css-examples/basic-user-interface/cursor.html
```diff
@@ -0,0 +1,56 @@
+<section id="example-choice-list" class="example-choice-list" data-property="cursor">
+  <div>
```
wbamberg: There's an extra `<div>` here that should be removed.

live-examples/css-examples/basic-user-interface/cursor.html
```diff
@@ -0,0 +1,56 @@
...
+  </div>
+
+  <div class="example-choice">
+    <pre><code class="language-css">cursor: grab;</code></pre>
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
+
+</div>
```
wbamberg: Here's the other end of that `<div>`.

live-examples/css-examples/cursor/cursor.css
```diff
@@ -0,0 +1,11 @@
+#example-element {
```
wbamberg: You still need to remove these files under `cursor`.

brianlmacdonald:
Wahoo! @wbamberg  Let me know if you'd like me to open the Eslint-config as an issue.

schalkneethling:
> Yes, very! @schalkneethling , is there a style guide that @brianlmacdonald could use for this?

@wbamberg @brianlmacdonald So, there are a couple of things.

1. The `eslintrc` for this project - https://github.com/mdn/interactive-examples/blob/master/.eslintrc.js
2. The `eslintrc` for Kuma(MDN Web Docs) - https://github.com/mozilla/kuma/blob/master/.eslintrc.js
3. And then there is a code styleguide in the works here that will also cover all these things.

With all of that said, there is a couple of things I am doing at the moment over and above the above ;)

1. Merging the `eslintrc` from Kuma and the interactive examples
2. Ensure that both projects have a pre-commit hook that lints the relevant JS
3. As soon as this PR[1] is merged, I am going to decouple the build tool from this project. One of the first items on the list to improve the builder to lint the example JS and CSS when pages are built so things are flagged at dev/build time.

wbamberg:
I just updated the page -> https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

PR #598, opened 2018-02-21
Logical properties

rachelandrew:
A set of examples for logical properties. I have tried to keep these consistent with their physical counterpart examples, which I think makes sense, although these need to show the changes in Writing Mode or direction.


live-examples/css-examples/logical-properties/border-inline-end-color.html
```diff
@@ -0,0 +1,42 @@
...
+        </code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+    
+    <div class="example-choice">
+        <pre><code class="language-css">border-inline-end-color: rgb(170, 50, 220, .6);
+writing-mode: horizontal-tb;
+direction: rtl;</code></pre>
```
wbamberg: Because this choice is 3 lines, this example (and all the `inline` ones I think) is going to overflow, which I would like to avoid.

I think I'd suggest only having 3 examples, and dropping the 4th one:

```
border-inline-end-color: hsl(60, 90%, 50%, .8);
writing-mode: vertical-lr;
```

...which is close enough to the second:

```
border-inline-end-color: #32a1ce; 
writing-mode: vertical-rl;
```

Does that sound OK to you?

live-examples/css-examples/logical-properties/inline-size.html
```diff
@@ -0,0 +1,43 @@
...
+        </code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+</section>
+
+<div id="output" class="output large hidden">
+    <section id="default-example" class="default-example">
+      <div id="example-element" class="transition-all">This is a box where you can change the block-size.</div>
```
wbamberg: Should be `inline-size`?

live-examples/css-examples/logical-properties/margin-block.css
```diff
@@ -0,0 +1,21 @@
...
+    align-content: flex-start;
+    flex-direction: column;
+    justify-content: flex-start;
+}
+
+.row {
+    height: 33.33%;
+    display: inline-block;
+    border: solid #5B6DCD 10px;
+    background-color: rgba(229, 232, 252, 0.6);
```
wbamberg: Please omit leading zeroes: `0.6` -> `.6`.

live-examples/css-examples/logical-properties/margin-block.css
```diff
@@ -0,0 +1,21 @@
...
+    flex-shrink: 0;
+}
+
+#example-element {
+    border: solid 10px #FFC129;
+    background-color: rgba(255, 244, 219, 0.6);
```
wbamberg: Also here, please omit leading zeroes: `0.6` -> `.6`.

live-examples/css-examples/logical-properties/margin-inline.css
```diff
@@ -0,0 +1,20 @@
...
+    height: 200px;
+    display: flex;
+    align-content: flex-start;
+    justify-content: flex-start;
+}
+
+.col {
+    width: 33.33%;
+    border: solid #5B6DCD 10px;
+    background-color: rgba(229, 232, 252, 0.6);
```
wbamberg: Also here, please omit leading zeroes: `0.6` -> `.6`.

live-examples/css-examples/logical-properties/margin-inline.css
```diff
@@ -0,0 +1,20 @@
...
+    flex-shrink: 0;
+}
+
+#example-element {
+    border: solid 10px #FFC129;
+    background-color: rgba(255, 244, 219, 0.6);
```
wbamberg: Also here, please omit leading zeroes: `0.6` -> `.6`.

live-examples/css-examples/logical-properties/padding-block-end.html
```diff
@@ -0,0 +1,44 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="padding-block-start">
```
wbamberg: This should be `padding-block-end`.

live-examples/css-examples/logical-properties/border-inline-end-color.html
rachelandrew: yup, I've updated the inline ones accordingly.

live-examples/css-examples/logical-properties/border-block-end-color.html
```diff
@@ -0,0 +1,42 @@
...
+        </button>
+    </div>    
+</section>
+    
+<div id="output" class="output large hidden">
+    <section id="default-example" class="default-example">
+        <div id="example-element" class="transition-all">This is a box with a border around it.</div>
+    </section>
+</div>
+
```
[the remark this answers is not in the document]
rachelandrew: I think they all do end in a blank line

rachelandrew:
Made a bunch of updates to address the requests. Chrome doesn't support these yet, would be good to apply some encouragement there, as having to use the physical properties with writing modes is horrible.

Also, some of the nits (leading zeros and so on) were values copied over from the physical examples which are a bit untidy compared to the new ones. I might do a round of tidying up at some point to make things consistent.

wbamberg:
Thanks @rachelandrew , for these updates! They look really good, and I appreciate the consistency with the physical properties.

> Also, some of the nits (leading zeros and so on) were values copied over from the physical examples which are a bit untidy compared to the new ones. I might do a round of tidying up at some point to make things consistent.

Sorry about that. I've not been very consistent in picking these things up. I know Schalk has [plans to have these things linted automatically](https://github.com/mdn/interactive-examples/pull/597#issuecomment-367409830), which will be great.

>  I might do a round of tidying up at some point to make things consistent.

That would be awesome. Maybe it would be worth waiting until we have linting, then it can find the places to update?

live-examples/css-examples/logical-properties/border-block-end-color.html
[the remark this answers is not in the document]
wbamberg: It looks to me like they end with a line containing only 4-spaces, but I don't think it's worth not merging for :).

PR #601, opened 2018-02-22
Clean up appearance of list-style-type example

ddbeck:
Following up from the conversation on #594, this removes the complicated description of `@counter-style` in the `list-style-type` example.

### DIFF
--- a/live-examples/css-examples/lists/list-style-type.html
+++ b/live-examples/css-examples/lists/list-style-type.html
@@ -42,14 +42,6 @@
 
         <hr>
 
-        <div class="note">
-            <pre><code>@counter-style space-counter {
-    symbols: "\1F680"
-             "\1F6F8"
-             "\1F6F0"
-             "\1F52D";
-    suffix: " "
-}</code></pre>
-        </div>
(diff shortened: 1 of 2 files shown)


live-examples/css-examples/lists/list-style-type.html
```diff
@@ -42,14 +42,6 @@
...
-        <div class="note">
-            <pre><code>@counter-style space-counter {
-    symbols: "\1F680"
-             "\1F6F8"
-             "\1F6F0"
-             "\1F52D";
-    suffix: " "
-}</code></pre>
-        </div>
+        <div class="note"><p><code>space-counter</code> is defined with <a href="https://developer.mozilla.org/docs/Web/CSS/%40counter-style"><code>@counter-style</code></a></p>
```
schalkneethling: Oh wait, seeing that this will be in an iframe, the anchor will need `target="_parent"`, else this will load inside the iframe. Also, the URL need to be preceded with `//` else it will "inherit" the domain of the iframe.

Lastly, it seems like MDN is not playing nice with the encoded `@` symbol so, perhaps make the URL `//developer.mozilla.org/docs/Web/CSS/@counter-style`

ddbeck:
Fixed the link. Thank you @schalkneethling!

PR #602, opened 2018-02-22
Add example for animation-fill-mode CSS property

ddbeck:
This PR adds an example for [`animation-fill-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode).

I think this might still be a work in progress. It's quite difficult to show the relationship between the mode and the animation without a lot of context for the animation itself (particularly the animation delay). I'm open to ideas on how to improve this. Let me know if you have any suggestions. Thanks!

### DIFF
--- a/live-examples/css-examples/animation/animation-fill-mode.css
+++ b/live-examples/css-examples/animation/animation-fill-mode.css
@@ -1,6 +1,7 @@
 #example-element {
     background-color: #1766aa;
-    margin: 20px;
+    margin: auto;
+    margin-left: 0;
     border: 5px solid #333;
     width: 150px;
     height: 150px;
--- a/live-examples/css-examples/animation/animation-fill-mode.css
+++ b/live-examples/css-examples/animation/animation-fill-mode.css
@@ -1,5 +1,6 @@
 #example-element {
     background-color: #1766aa;
+    color: white;
     margin: auto;
     margin-left: 0;
     border: 5px solid #333;
@@ -24,10 +25,12 @@
 @-webkit-keyframes slide {
     from {
         background-color: orange;
+        color: black;
         margin-left: 0;
     }
     to {
         background-color: orange;
+        color: black;
         margin-left: 80%;
     }
 }
@@ -35,10 +38,12 @@
 @keyframes slide {
     from {
         background-color: orange;
+        color: black;
         margin-left: 0;
     }
     to {
         background-color: orange;
+        color: black;
         margin-left: 80%;
     }
 }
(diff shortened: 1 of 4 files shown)


wbamberg:
For little bits of context, you can include them as part of the example choice. The example choice doesn't have to contain only the property that is the subject of the example: it can contain other declarations too, where they help the user understand what's going on.

An existing example where we do this is [`position`](https://developer.mozilla.org/en-US/docs/Web/CSS/position).

The big restriction here of course is whether it will fit vertically. 
So you could have something like:

```
animation-fill-mode: none;
animation-delay: 1s;
```
```
animation-fill-mode: forwards;
animation-delay: 1s;
```
```
animation-fill-mode: backwards;
animation-delay: 1s;
```
```
animation-fill-mode: both;
animation-delay: 1s;
```
...that seems to fit. But you couldn't include the keyframes, so I guess we'd just have to try to choose an animation that makes the keyframe values reasonably obvious :).

Was there a reason for not including `backwards` and `both`? From what I understand we would need a `from` keyframe to see the effect of these values.

ddbeck:
> For little bits of context, you can include them as part of the example choice. 

OK, great. I think that'll help a lot!

> Was there a reason for not including `backwards` and `both`? From what I understand we would need a `from` keyframe to see the effect of these values.

I ended up dropping `backwards` and `both` for simplicity, after I couldn't make a non-cryptic example (I actually showed my first draft to someone and they got a distinctly wrong idea about what was happening). But I'm going back to the drawing board on this—now I've got at least two ideas about how to make this work. Thanks!

ddbeck:
OK, let's try this again. A few points on this change:

* I've updated the example to show `backwards` and `both`.
* After a little more informal user testing, I changed the animation itself to be a little more obvious about where it begins and ends (the snapping back on the spinning animation was misinterpreted as a full rotation).
* I added the play state indicator, to avoid explicitly showing the keyframes while still reinforcing the relationship with `animation-delay`.

Open to more feedback on this. Thanks!

ddbeck:
> the non-optional: I think the black text on dark blue, for the initial state, is not contrasty enough. Maybe start with white text, and switch to black along with the background-color change?

Good call! Fixed.

> the optional one: 2s delay is quite long, maybe 1s is enough to get the point across? But this is very subjective. Maybe I just drank too much coffee this morning.

Yeah, I agree with this. I had it as short as half a second when I was testing this because I knew what to look for. I may have overcompensated a bit with `2s`. Fixed.

While fixing these other issues, I noticed a little jerk in the animation caused by mismatches in the margins. That's fixed too. Hopefully this is ready to land, but let me know if you want to see any other changes. Thank you!

PR #603, opened 2018-02-22
Layout examples

rachelandrew:
A bunch of box model examples, plus positioned layout physical offsets.


live-examples/css-examples/basic-box-model/clear.css
```diff
@@ -0,0 +1,21 @@
+.example-container {
+    background-color: #eee;
+    border: .75em solid;
+    padding: .75em;
+    text-align: left;
+}
+
+.floated-left {
+    border: solid 10px #FFC129;
```
wbamberg: We should have lowercase for hex values, here and elsewhere in this PR.

live-examples/css-examples/basic-box-model/clear.css
```diff
@@ -0,0 +1,21 @@
...
+    background-color: rgba(255, 244, 219, 0.6);
```
wbamberg: We should omit leading zeroes, here and elsewhere in this PR.

live-examples/css-examples/basic-box-model/clear.css
rachelandrew: fixed.

live-examples/css-examples/basic-box-model/clear.css
rachelandrew: fixed.

PR #604, opened 2018-02-22
Revert "Adds next iteration of the HTML interactive editor"

wbamberg:
Reverts mdn/interactive-examples#576

I believe this change breaks the CSS examples. I think it's due to this line: https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/tabby.js#L3, which assumes the previous line does not return `null` - but it does, I guess because the HTML editor is not defined in CSS examples.

So the example looks like this: https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow?v=b


wbamberg:
Oh, this also revert https://github.com/mdn/interactive-examples/commit/785633bfd735cb7e443f551e0037e164ea757cde, which I wasn't expecting.

metadave:
Should this commit be included? [Auto run on load, and change Run to Update](https://github.com/mdn/interactive-examples/pull/576/commits/c6f9e2732843ad92a64a64b6765e2070e5bafa56)

PR #605, opened 2018-02-22
Revert "Adds next iteration of the HTML interactive editor (#576)"

metadave:
This reverts commit 432e90b7c71b3a57bb5222128ee7f4f8bba1a180.


metadave:
See https://github.com/mdn/interactive-examples/pull/604

wbamberg:
This reverts the change to add the HTML editor, because that change breaks the CSS examples. I think it's due to this line: https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/tabby.js#L3, which assumes the previous line does not return `null` - but it does, I guess because the HTML editor is not defined in CSS examples.

So the example looks like this:
screen shot 2018-02-22 at 12 41 23 pm

PR #606, opened 2018-02-22
Issue 561 Added example for vertical-align in text context

LLyaudet:
Hello @wbamberg,
Here is the pull request for vertical-align. I put my commit in a branch issue-561-css-vertical-align if there is any correction to add to the branch.
Best regards,
  Laurent Lyaudet


LLyaudet:
Hello @wbamberg,
I pushed the commit for moving the source in a new directory inline-layout under css-examples.
I modified meta.json in text directory and created one such file in inline-layout directory.
Is there anything else to do to have the inline-layout examples up and running ?
There is also merge commits since I messed up by fetching upstream and merging from upstream master into my branch instead of pulling my branch with the merge you did from my fork.
I need practice on this. That's my first pull request.
Best regards,
   Laurent Lyaudet

wbamberg:
This looks fine! Just you forgot to update the paths inside meta.json when you moved the files from "text" to "inline-layout". I just pushed a commit taking care of that. I also made the CSS selector more specific - as you pointed out before, it is possible for styles in here to interfere with the whole example - they don't in this particular case, but still it seems like a good idea to avoid that possibility.

wbamberg:
I've just updated the page: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align :)

PR #607, opened 2018-02-23
Remaining logical properties, min and max, offsets

rachelandrew:
The offset properties are a little tricky to show as to change the value of the property and see the change in direction requires changing two things (the abspos item AND the writing mode of the container). I have therefore chosen to change the writing mode, and detail in the example what setting the abspos item has.

Though if anyone has a better solution, I'm happy to change it!

### DIFF
--- a/live-examples/css-examples/logical-properties/max-block-size.html
+++ b/live-examples/css-examples/logical-properties/max-block-size.html
@@ -1,0 +1,41 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="max-block-size">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">max-block-size: 150px;
+writing-mode: horizontal-tb;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">max-block-size: 150px;
+writing-mode: vertical-rl;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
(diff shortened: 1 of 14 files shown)


live-examples/css-examples/logical-properties/max-block-size.html
```diff
@@ -0,0 +1,41 @@
...
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">max-block-size: 20px;
+writing-mode: horizontal-tb;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
wbamberg: I don't think we need this option, here and in the other related examples. It's essentially just option 1.

wbamberg:
>  I have therefore chosen to change the writing mode, and detail in the example what setting the abspos item has.

This seems like a good approach, they work well. The only thing I could imagine would be: wouldit be worth having a kind of ghost version of the positioned element, to highlight what the offset value is doing?

Like:

```
#originalpos {
    background-color: rgb(255, 255, 0, 0.4);
    border: 1px dashed red;
    position: absolute;
    inline-size: 140px;
    min-block-size: 120px;
}
```

giving us something like:

screen shot 2018-02-23 at 2 09 07 pm

Do you think this is helpful, or too fussy?

rachelandrew:
I wonder if it's potentially confusing, as the person playing with this can't see the CSS that adds this element, so might it just be a bit mysterious?

wbamberg:
Yes, fair enough, let's keep it simple :).

live-examples/css-examples/logical-properties/max-block-size.html
rachelandrew: The intention was to keep it consistent with max-height/width and show that it would cause content to be hidden. I think that is useful.

wbamberg: OK, that makes sense.

PR #608, opened 2018-02-23
Next iteration of the html editor

schalkneethling:
Alrighty. Discovered the problem that was breaking all the things. Not sure how neither of us picked this up :-/ I triple tested the CSS, JS and tabbed examples and all works now.

The problem was a `require` for `tabby.js` inside `mce-utils`. For the tabbed UI this resolves but, the CSS and JS editors does not include this file in their bundles so, this failed.

Instead of bloating the CSS and JS examples, and simple refactored out the `getOutput` function into `editor.js`


js/editor-libs/mce-utils.js
```diff
@@ -16,15 +14,6 @@ module.exports = {
...
     },
-    /**
-     * Called by the tabbed editor to combine code from all tabs.
-     * @returns Concatenated code from all tabs
-     */
-    getOutput: function() {
-        var html = tabby.editors.html.editor.getValue();
-        var style = '<style>' + tabby.editors.css.editor.getValue() + '</style>';
-        return style + html;
-    },
```
schalkneethling: This here and at line 1 above was the root cause of the problem.

PR #609, opened 2018-02-23
Add examples for counter-reset and counter-increment

ddbeck:
This PR adds examples for [`counter-reset`](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-reset) and [`counter-increment`](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-increment). Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/lists/counter-increment.html
+++ b/live-examples/css-examples/lists/counter-increment.html
@@ -1,0 +1,41 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="list-style">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">counter-increment: chapter-count;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">counter-increment: chapter-count 0;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">counter-increment: chapter-count 5;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">counter-increment: chapter-count -5;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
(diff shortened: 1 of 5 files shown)


wbamberg:
I think it seems strange to have counters used for chapters when they don't increment for each chapter, so it looks broken:

```
chapter 0
chapter 1
chapter 1
chapter 1
```

That said I'm not sure what the best approach is. We could change it so the chapters increment normally, then the example increments the increment, so you get:

```
chapter 0
chapter 2
chapter 3
chapter 4
```

...and so on. But maybe that's too much magic behind the scenes? Alternatively we could just go really simple and have something like:

```
// CSS
#example-element::after {
    content: counter(example-counter);
}
```

```
// HTML
<div id="example-element" class="transition-all">Counter value: </div>
```

...so we just display the value. I think this would be OK to demonstrate the property, although it's not very exciting.

ddbeck:
I think we started with similar thoughts on this. I started with incrementing the increment, then switched to the broken-looking increment as a compromise. It's not very exciting, but the counter value idea is good—I wish I had thought of it. 😄 

Is the `counter-reset` example OK, though?

live-examples/css-examples/lists/counter-increment.html
```diff
@@ -0,0 +1,35 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="list-style">
```
wbamberg: `list-style` -> `counter-increment`

live-examples/css-examples/lists/counter-reset.html
```diff
@@ -0,0 +1,41 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="list-style">
```
wbamberg: `list-style` -> `counter-reset`

live-examples/css-examples/lists/counter-reset.html
```diff
@@ -0,0 +1,41 @@
...
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">counter-reset: chapter-count 0;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+                <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
wbamberg: I think it would be good to have (possibly as the first option and selected initially), an extra option with `counter-reset: none;` so we can see the initial effect without any `counter-reset` happening.

What do you think?

live-examples/css-examples/lists/counter-increment.html
ddbeck: Oops, the hazards of copying and pasting. I'll fix this in _all_ the places I made this particular mistake.

live-examples/css-examples/lists/counter-reset.html
ddbeck: Yeah, this sounds like a good idea. Adding it now.

PR #612, opened 2018-02-24
Shapes and clip-path

rachelandrew:
I went for a simple clip-path example, there is a more complex SVG demo on the page which really needs the whole thing visible rather than just toggling a value. So I think it worth keeping that one as well as a simple demo in the live example.


wbamberg:
> I went for a simple clip-path example, there is a more  complex SVG demo on the page which really needs the whole thing visible  rather than just toggling a value. So I think it worth keeping that one  as well as a simple demo in the live example.

Yes, this seems like a good approach. I really like these examples, but a couple of comments:

* the top is cut off at smaller (but not that much smaller) widths. For instance, the `clip-path` one at 900 px, with "polygon" selected:

screen shot 2018-02-26 at 1 02 41 pm

What do you think of changing:

```
output section {
    align-items: center;
}
```

to

```
output section {
    align-items: start;
}
```
? This gives us:

screen shot 2018-02-26 at 1 03 24 pm

* Second, just for `shape-image-threshold`, I wondered if it's worth using an example closer to the [existing live sample](https://developer.mozilla.org/en-US/docs/Web/CSS/shape-image-threshold#Example), with a gradient. I found it easier to understand the property with that example. Just a suggestion though.

rachelandrew:
both good suggestions, have updated.

wbamberg:
> So I think it worth keeping that one as well as a simple demo in the live example.

Makes sense to me. I kept the big example, but moved it down the page under the "Example" section.

PR #615, opened 2018-02-25
Added datalist example; slight reorg of table example

wbamberg:
Adds an example of [`datalist`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist), for user testing.

I chose this because it has slightly interesting syntax, a clear visual result, and is a common element.

I also slightly reorged the table example, as it seemed over-engineered. Like this we just have example HTMLs all flat, and their CSS files in a separate "css" directory.

I might also try to make the `table` example a bit less dry, but that could wait for another PR. 

Happy for any feedback :).


live-examples/html-examples/css/datalist.css
```diff
@@ -0,0 +1,3 @@
+label, input {
```
schalkneethling: Nit, code style: Each selector on its own line

live-examples/html-examples/css/datalist.css
```diff
@@ -0,0 +1,3 @@
...
+  font-size: 1.2em;
```
schalkneethling: Nit, code style: indent with 4 spaces. I believe we should stick to `rem` units for font sizes?. This would be the recommendation in the code style guide I am working on.

Issue #532, opened 2018-02-02
CSS editor is confusing when the feature isn't supported


wbamberg:
When a CSS feature isn't supported, the editor just displays a static version of the example choices, like this:

<img width="809" alt="screen shot 2018-02-02 at 8 43 07 am" src="https://user-images.githubusercontent.com/432915/35744403-229b4a02-07f5-11e8-93da-4c0c9e7e9885.png">

I think this is not ideal. Part of the original reason for displaying the static list of choices was that we thought the editor would replace the existing PRE list, but I now think we should keep the PRE list, as it is kind of complementary to the editor. But that makes the static list look quite repetitive, and when there are only a couple of choices, they look lost in the iframe and it's easy to think something is broken.

I'm not sure what would be better though. We could just hide the iframe completely, or could resize it and display in it a message like "this feature is not supported by your browser, so we can't display the interactive example". I'm not really sure how technically feasible these are though - I know we already do some postMessaging between the MDN page and the editor page, so perhaps it could use that.

majidzeno:
Can't we take a screenshot for this unsupported property and show a static image in the example area with a title "This Feature is not supported by your browser, you can only display a static image" something like mirroring.
![mozilla](https://user-images.githubusercontent.com/18016774/37226742-99fe3848-23e3-11e8-89af-16f4f61980a9.png)

wbamberg:
I've promised for ages to write something up about this. Here are some suggestions for discussion.

# Dealing with unsupported CSS

One big weakness of the CSS interactive examples editor is that it doesn’t handle unsupported CSS well. At the moment this is stopping us from including examples for properties that don’t have good cross-browser support (including, for example, all the logical properties).

## The current implementation

The current implementation goes like this: the example includes an attribute `data-property` whose value is the name of the property. The editor uses that to test whether the browser supports the property (https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/mce-utils.js#L36-L54). If it does, all’s good. Otherwise, it displays a static rendering of just the example choices, like this: https://screenshots.firefox.com/q2euYmzsNczUT0ws/interactive-examples.mdn.mozilla.net.

There are a couple of problems with this. The first is that it confuses people. They just think the example is broken somehow (https://github.com/mdn/interactive-examples/issues/528). The second is that it’s not very fine-grained: it doesn’t help if the property is prefixed, or if some syntax forms are supported but not others.

## What would we like to happen?

[...]

PR #617, opened 2018-02-26
Add example for CSS pointer-events property

ddbeck:
This PR adds an example for [`pointer-events`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events). Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/svg/pointer-events.html
+++ b/live-examples/css-examples/svg/pointer-events.html
@@ -1,0 +1,48 @@
+<section id="example-choice-list" class="example-choice-list" data-property="animation">
+    <div class="example-choice">
+        <pre><code class="language-css">pointer-events: auto;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">pointer-events: none;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">pointer-events: stroke; /* SVG-only */</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">pointer-events: fill; /* SVG-only */</code></pre>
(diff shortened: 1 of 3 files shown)


wbamberg:
This is a nice example! I appreciate you including the SVG-specific values. What do you think about making the stroke wider, to make the difference more obvious:

screen shot 2018-02-26 at 1 51 19 pm

ddbeck:
I tried that initially because you're right, it is more obvious for the `stroke` value. But it makes things a bit odd for the `fill` value. Apparently, the stroke is drawn from the edge of the circle both inward and outward, so the cursor changes halfway through the stroke, like this:

svg_slow

I'm not sure which value we ought to optimize for, to be honest.

wbamberg:
Fair enough, I hadn't noticed that :). Let's keep it then.

live-examples/css-examples/svg/pointer-events.html
```diff
@@ -0,0 +1,48 @@
+<section id="example-choice-list" class="example-choice-list" data-property="animation">
```
wbamberg: Should be "pointerEvents" ?

ddbeck:
A quick question: should the `data-property` value be camel case or the same as the property itself? I'm a little mixed up across PRs about which is the right way to do it.

wbamberg:
> should the data-property value be camel case or the same as the property itself? 

As I understand things, it doesn't matter. It's used here: https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/mce-utils.js#L41, and as far as I can tell, either form works:

```
var a = document.createElement('div');

a.style["pointerEvents"]
< ""

a.style["pointer-events"]
< ""

a.style["bananas"] 
< undefined
```

But still, it's a shame we are not consistent here :(.

ddbeck:
@wbamberg OK, I did a quick-and-dirty census of the `data-property` values with ripgrep:

```fish
$ rg 'data-property="[\w-]+"' | wc -l  # total
238

$ rg --case-sensitive 'data-property="[a-z-]+"' | wc -l # hyphenated
192

$ rg --case-sensitive 'data-property="\w+[A-Z]+\w+"' | wc -l # camelCase
46
```

So I guess I'll stick with hyphenated going forward. If you want, I can open an issue around making them consistent.

wbamberg:
Thanks for this analysis, @ddbeck !

>  If you want, I can open an issue around making them consistent.

Yes, please! If you can get a list of the camelCase examples, that would be a really helpful thing to add. We should also make the docs on this clearer.

I expect this mechanism might change in the future, as I want us to have a better way of dealing with limited support (see also https://github.com/mdn/interactive-examples/issues/532 and https://github.com/mdn/interactive-examples/issues/534). But that might be a while away.

PR #621, opened 2018-02-26
Add example for text-align-last CSS property

ddbeck:
This PR adds an example for [`text-align-last`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align-last). Let me know if you want to see any changes. Thanks!


live-examples/css-examples/text/text-align-last.html
```diff
@@ -0,0 +1,30 @@
+<section id="example-choice-list" class="example-choice-list" data-property="textAlign">
```
wbamberg: Should be textAlignLast?

ddbeck:
Fixed the attribute. Thanks!

PR #622, opened 2018-02-26
Break inside

rachelandrew:
I put a note in the spreadsheet, but these breaking properties are something of a struggle, as support is so poor. I've done one - let me know if you think we should have these or they are better left without until there is better support.


rachelandrew:
cool I'll close this PR then and mark them as such on the spreadsheet.

PR #631, opened 2018-02-28
Adds note regarding CSS code style and link to docs

schalkneethling:


schalkneethling:
Also added a comment regarding SVG and SVGOMG

CSS-Example-Style-Guide.md
```diff
@@ -70,9 +70,14 @@ Sometimes you'll want to include images with the example. If you do:
...
 * run them through https://tinypng.com or https://imageoptim.com, to reduce the page weight of the examples.
+* For `SVG`, run the code through (SVGOMG)[https://jakearchibald.github.io/svgomg/], and ensure that the `SVG` file has an empty line at the end of the file
 
 ## Output width considerations
 
 On MDN pages the editor is laid out "side by side": that is with the example choices on the left and the output on the right, as it is in the local server that's started by `npm run start`. Then if the page width goes below some threshold it switches to "top and bottom", with the example choices above and the output below.
 
 This means that ideally, the example should still work with an editor width of about 730 pixels: https://screenshots.firefox.com/YYrEvqLEmLjJCddS/developer.mozilla.org. This can be a difficult constraint to satisfy, but you should test at this width, and try to make it work, if it's possible.
+
+## CSS Syntax
```
wbamberg: Nit: "CSS syntax"

CSS-Example-Style-Guide.md
```diff
@@ -70,9 +70,14 @@ Sometimes you'll want to include images with the example. If you do:
...
+
+The interactive-examples project follows the same guidelines with regards to CSS code style as other projects that form part of the MDN Web Docs project. To ease review time for everyone, and ensure good coding practice, please review the guidelines on [terminology](https://schalkneethling.github.io/mdn-fiori/patterns/css/terminology/) and [formatting](https://schalkneethling.github.io/mdn-fiori/patterns/css/formatting/) when writing CSS.
```
wbamberg: Regarding this: https://schalkneethling.github.io/mdn-fiori/patterns/css/formatting/#why-no-id-selectors - personally I don't agree with this blanket ban, and we use IDs all the time in our examples, in particular (but not only) `#example-element`.

schalkneethling: Yes, and it is something that I have been meaning to change. There is no reason why that cannot be a class as apposed to an id. You can even have the `id` as a JS hook, and then a class with the same name for styling.

Using id selectors in CSS quickly causes specificity problems. Within the context of the interactive examples, one might be able to get away with that but, in a codebase such as Kuma it is a bad idea.

As the code style is for all things MDN, I feel we need to enforce the same rules across all code bases.

PR #634, opened 2018-02-28
Rewrite the Object.create example.

tatellos:
The previous version worked just the same even when deleting the line which had Object.create in it!
Also, this version seems simpler to me, as it does not assume knowledge of functions as constructors and the new keyword, etc.
This version however assumes (teaches?) knowledge about the context of inherited methods. It could of course be even simpler, if we just leave out the sayName method and focus on primitive inherited fields.

### DIFF
--- a/live-examples/js-examples/object/object-create.html
+++ b/live-examples/js-examples/object/object-create.html
@@ -1,15 +1,17 @@
 <pre>
 <code id="static-js">const person = {
-  isHuman: true,
-  sayName: function () {
-    console.log(`My name is ${this.name}`);
+  isHuman: false,
+  printIntroduction: function () {
+    console.log(`My name is ${this.name}. Am I human? ${this.isHuman}`);
   }
 };
 
 const me = Object.create(person);
-me.name = "Matthew"; // name is now a property set on the me object, but not on the person object
 
-me.isHuman == true;
-me.sayName() // -> "My name is Matthew"
+me.name = "Matthew"; // "name" is a property set on "me", but not on "person"
+me.isHuman = true; // inherited properties can be overwritten
+
+me.printIntroduction();
+// expected output: "My name is Matthew"
 </code>
 </pre>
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/object/object-create.html
```diff
@@ -1,16 +1,15 @@
...
-}
+const me = Object.create(person);
+me.name = "Matthew"; // name is now a property set on the me object, but not on the person object
 
-Rectangle.prototype = Object.create(Shape.prototype);
-const rect = new Rectangle();
-
-console.log(rect.name);
-// expected output: "Shape 1"
+me.isHuman == true;
```
schalkneethling: I don't believe that this line is needed as it is here.

It is not an assignment but a comparison, and the result is not output anywhere. Perhaps you intended for it to be an assignment? Then, the default for `isHuman` is `true` by default so, perhaps it would make more sense to set it to `false`? 

You could then amend the functionality, and name of the `sayName()` function to output something like

```
`My name is ${this.name}. Am I human? ${this.isHuman}`
```

tatellos: I meant to show that the `isHuman` field is accessible on the `me` instance.
I like your idea to show this by using it inside the sayName (I will rename it to `introduceMyself`) method.

live-examples/js-examples/object/object-create.html
```diff
@@ -1,15 +1,17 @@
...
 const me = Object.create(person);
-me.name = "Matthew"; // name is now a property set on the me object, but not on the person object
 
-me.isHuman == true;
-me.sayName() // -> "My name is Matthew"
+me.name = "Matthew"; // "name" is a property set on "me", but not on "person"
+me.isHuman = true; // inherited properties can be overwritten
+
+me.printIntroduction();
+// expected output: "My name is Matthew"
```
schalkneethling: The expected output has changed to `My name is Matthew. Am I human? true`

PR #637, opened 2018-03-01
Add example for text-underline-position CSS property

ddbeck:
This PR adds an example for [`text-underline-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-position). I got far along with this before realizing that the property is not particularly well-supported (works in Chrome, not so much elsewhere), but since it was finished I figured I should offer it up anyway. Let me know if you want to see any changes (but I'm also OK with not merging it too). Thanks!

### DIFF
--- a/live-examples/css-examples/text-decoration/text-underline-position.css
+++ b/live-examples/css-examples/text-decoration/text-underline-position.css
@@ -3,5 +3,5 @@ p {
 }
 
 #example-element {
-	text-decoration-line: underline;
+    text-decoration-line: underline;
 }
--- a/live-examples/css-examples/text-decoration/text-underline-position.css
+++ b/live-examples/css-examples/text-decoration/text-underline-position.css
@@ -1,0 +1,7 @@
+p {
+    font: 1.5em sans-serif;
+}
+
+#example-element {
+	text-decoration-line: underline;
+}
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/text-decoration/text-underline-position.css
```diff
@@ -0,0 +1,7 @@
+p {
+    font: 1.5em sans-serif;
+}
+
+#example-element {
+	text-decoration-line: underline;
```
wbamberg: This uses a tab to indent, should use 4 spaces.

wbamberg:
> I got far along with this before realizing that the property is not particularly well-supported (works in Chrome, not so much elsewhere), but since it was finished I figured I should offer it up anyway. Let me know if you want to see any changes (but I'm also OK with not merging it too). Thanks!

The way I look at this is, there are 3 types of CSS thing:

1. well-supported, standardized, and make good examples. We need example for these.
2. proprietary, deprecated, or don't make good examples. We will never need examples for these.
3. some are standardized and make good examples, but don't (yet?) have good cross-browser support.

For type 3, in a perfect world the editor would nicely handle this (probably by just hiding the editor when the browser doesn't support the item). The editor doesn't at the moment handle this nicely, so I think we should not (yet) add these examples to the MDN pages. But I'm very happy for us to have good examples of them in the repo, then we can add them to the pages when the editor is improved or when other browsers support it.

So I think they are kind of optional. But as it is, we actually have examples for almost all type 3 items, which I think is great :).

ddbeck:
Thanks, @wbamberg! That's really helpful guidance. In that case, I'll keep on submitting PRs for these type 3 cases. Thanks again!

PR #639, opened 2018-03-01
Text examples

rachelandrew:
This includes widows and orphans which are a bit tricky to show as you need to resize to see it apply.

Anyway, if anyone has any better ideas to show that - let me know.


live-examples/css-examples/rhythm/line-height-step.css
```diff
@@ -0,0 +1,20 @@
...
+    grid-gap: 20px;
+    grid-template-columns: 1fr 2fr;
+}
+
+.example-container h2, p {
+    margin: 0;
+}
+  
+.example-container h2 {
+    font-size: 33px;
```
wbamberg: This overflows vertically quite easily - at about 1000 pixels wide it looks like:

screen shot 2018-03-05 at 7 53 59 pm

Could we reduce the font size and/or have a shorter text?

wbamberg:
> This includes widows and orphans which are a bit tricky to show as you need to resize to see it apply.

Yes, I agree. I think as it is we should omit these examples, because it's quite confusing when they seem to do nothing.

I could think of one option, but I'm not sure if it's very appealing.

It's possible to include JavaScript in the CSS examples. Not many examples do this, but a couple do: [`transform-origin`](https://github.com/mdn/interactive-examples/tree/master/live-examples/css-examples/transforms) and the [animation](https://github.com/mdn/interactive-examples/tree/master/live-examples/css-examples/animation) examples.

So if there's a way to make the browser recalculate the effect of this property from JavaScript, we could do it that way. In Chrome at least, setting the container size - even to the same thing it was before - seems to do the trick, something like:

```
'use strict';

var oldOrphans = null;
var el = document.getElementById('example-element');

window.addEventListener('load', function() {

    function update(mutation) {
        var orphans = window.getComputedStyle(el).orphans;
        if (orphans !== oldOrphans) {
          oldOrphans = orphans;
          el.style.width = window.getComputedStyle(el).width;
        }
    }

    var observer = new MutationObserver(function(mutations) {
        update(mutations);
    });

    observer.observe(el, {
        attributes: true,
        attributeFilter: ['style']
    });

    update();
});
```

But I don't know if this works in other browsers, and maybe there's a better way :). What do you think?

rachelandrew:
yeah, I'd wondered about doing something similar. However I wondered if it might be confusing to someone expecting their test to behave in the same way and not realising they need to reload the page.

wbamberg:
Sorry to be slow getting back about this.

> However I wondered if it might be confusing to someone expecting their test to behave in the same way and not realising they need to reload the page.

Good point. I hadn't thought of this, and agree it might be a concern. But I think it is less confusing than the current example, which shows no change when the user selects an option.

I think we should either do something like https://github.com/mdn/interactive-examples/pull/639#issuecomment-370656619 (although it is pretty hacky) or omit these examples at least for now, which I think would also be a fine option. Maybe better.

Could you please resolve the merge conflicts in this PR, too?

rachelandrew:
Perhaps omit the examples for now (or at least don't add them to a page) and perhaps we can figure out a better way to do it.

I wonder if we could somehow alert the viewer to the fact that we have to cause the browser to redraw it.

PR #640, opened 2018-03-01
Changes HTML editor to autoupdate, removes update button

schalkneethling:


css/editable-js.css
```diff
@@ -89,10 +89,6 @@
         align-items: center;
     }
 
-    .run {
-        margin-right: 10px;
-    }
-
     .output {
```
wbamberg: I think because of this change, we are losing the gap between the buttons and the output in the JS examples:

screen shot 2018-03-01 at 4 59 33 pm

cf

screen shot 2018-03-01 at 4 59 59 pm

PR #641, opened 2018-03-01
Add examples for font-variant and font-variant-caps CSS property

ddbeck:
This PR adds an example for [`font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant) and [`font-variant-caps`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps). Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/fonts/font-variant-caps.css
+++ b/live-examples/css-examples/fonts/font-variant-caps.css
@@ -1,0 +1,13 @@
+@font-face{
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
+         url('../../../media/fonts/FiraSans-Regular.woff2') format('woff2');
+    font-weight: normal;
+    font-style: normal;
+}
+
+#output section {
+    font-family: 'Fira Sans', sans-serif;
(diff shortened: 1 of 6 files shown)


live-examples/css-examples/fonts/font-variant.css
```diff
@@ -0,0 +1,22 @@
+@font-face{
```
wbamberg: Super-nit: space before `{`

ddbeck: Fixed, thank you!

live-examples/css-examples/fonts/font-variant-caps.css
```diff
@@ -0,0 +1,13 @@
+@font-face{
```
schalkneethling: Nit, code style: I reckon you just overlooked this one 👁

ddbeck: Yeah I did. I just pushed a change to find-and-replace the problem everywhere. :-)

PR #643, opened 2018-03-02
Scroll snapping

rachelandrew:
Here is an example for scroll-snap-type. For the new spec so only works in Safari. Building this highlights various issues in the MDN docs for scroll-snapping.

We have all the properties now removed from the spec (but supported by Firefox), we don't have scroll-snap-align. The current example on the scroll-snap-type page is incorrect now as to the updated spec, although uses the properties as supported by Firefox. It looks like the browser information is incorrect too.

Quite difficult to know what to do with these specs which have a bit of implementation but then change in a major way.

### DIFF
--- a/live-examples/css-examples/scroll-snap/meta.json
+++ b/live-examples/css-examples/scroll-snap/meta.json
@@ -1,0 +1,14 @@
+{
+    "pages": {
+        "scrollSnap": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "cssExampleSrc":
+                "../../live-examples/css-examples/scroll-snap/scroll-snap-type.css",
+            "exampleCode":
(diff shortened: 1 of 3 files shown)


schalkneethling:
@wbamberg I am going to leave this one to you as @rachelandrew also added some comments with regards to the actual page content of the `scroll-snapping` and related items.

wbamberg:
I think there are a few things in here.

First, the example looks good, so let's merge it.

Second, for the docs, I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1443884 to update them.

Third, as you say, it's hard to know how to handle this situation, where the same property has such incompatible behavior across browsers. I think for now, we should not embed this example in MDN (but it's good to have in the repo all the same). Perhaps eventually we'll have some nice way of indicating that it's not usable on all browsers.

PR #647, opened 2018-03-05
Add example for image-rendering CSS property

ddbeck:
This PR adds an example for [`image-rendering`](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering), though I think it might still be a work in progress. The main issue here is that it's hard to show this property in action without using vendor prefixes on the values, but I think that would really complicated the example. Any advice on how to improve this one would be appreciated. Thanks!

### DIFF
--- a/live-examples/css-examples/images/image-rendering.css
+++ b/live-examples/css-examples/images/image-rendering.css
@@ -1,0 +1,5 @@
+#example-element {
+    height: 500px;
+    object-fit: cover;
+    border: 1px solid red;
+}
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/images/image-rendering.css
```diff
@@ -0,0 +1,5 @@
+#example-element {
+    height: 500px;
+    object-fit: cover;
+    border: 1px solid red;
```
schalkneethling: This is a tricky one indeed, but a really need new CSS thing that is on the horizon. My only suggestion here would be to remove the border, and set the `height` to `480px`

ddbeck: Fixed, thank you!

PR #648, opened 2018-03-05
Add examples for more animation- properties

ddbeck:
This PR adds examples for:

* [`animation-direction`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction)
* [`animation-duration`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-duration)
* [`animation-play-state`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-play-state)

Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/animation/animation-direction.css
+++ b/live-examples/css-examples/animation/animation-direction.css
@@ -7,10 +7,6 @@
     width: 150px;
     height: 150px;
     border-radius: 50%;
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    flex-direction: column;
 }
 
 .animating {
--- a/live-examples/css-examples/animation/animation-direction.css
+++ b/live-examples/css-examples/animation/animation-direction.css
@@ -1,0 +1,47 @@
+#example-element {
+    background-color: #1766aa;
+    color: white;
+    margin: auto;
+    margin-left: 0;
+    border: 5px solid #333;
+    width: 150px;
+    height: 150px;
+    border-radius: 50%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    flex-direction: column;
+}
+
+.animating {
+    animation-name: slide;
+    animation-duration: 3s;
+    animation-timing-function: ease-in;
+    animation-iteration-count: infinite;
+}
+
+@-webkit-keyframes slide {
+    from {
+        background-color: orange;
+        color: black;
+        margin-left: 0;
+    }
+    to {
+        background-color: orange;
+        color: black;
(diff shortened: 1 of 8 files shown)


live-examples/css-examples/animation/animation-direction.css
```diff
@@ -0,0 +1,47 @@
...
+    margin: auto;
+    margin-left: 0;
+    border: 5px solid #333;
+    width: 150px;
+    height: 150px;
+    border-radius: 50%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    flex-direction: column;
```
wbamberg: I'm not sure you need lines 10-13 for this or the other new examples. I might be wrong though (I know you do for animation-delay, because that has extra bits. And I think maybe you'll need to add extra bits to these examples as well.

ddbeck:
> But also, I'm worried that (especially infinite looping) animations will be annoying for people who just want to read the docs. Not everyone is going to want to engage with the interactive examples - some people will just ignore them.
>
> That said I'm not sure what's best. The obvious suggestion would be a Start/Stop button in the output. What do you think?

Yeah, I think that's an important consideration. I'm a little worried that a start/stop button poses some risks, though. If it's click-to-start, will readers miss that the examples are interactive? Or if it's click-to-pause, will we spare the readers who don't want the animation much annoyance?

I'm wondering if we can't be a bit more clever, maybe disabling the animation after a certain amount of time, if there's no interaction with the interactive example?

wbamberg:
Yes, I agree that buttons are tricky. I'd like to prioritize not annoying people who don't like animations though. The `transition` example has a similar problem, and its solution is to ask the user to hover over the output box: https://developer.mozilla.org/en-US/docs/Web/CSS/transition. Do you think that could work here?

ddbeck:
I saw this last night and I really liked the hover idea, but now that I've slept on it, I think I prefer the idea of play/pause buttons. I'm not quite sure what hover means if you're not using a mouse pointer (e.g., on my phone in Chrome for Android, hover is something like "tap and hold something that's not an anchor" but it feels a little unnatural).

I'll redo these with a button and we'll see how it goes.

ddbeck:
OK, I've added play/pause buttons to `animation-direction` and `animation-duration`. For `animation-play-state`, I thought it was simpler to set `paused` to be the default. Let me know what you think!

ddbeck:
And of course now that I've done it, setting `animation-duration`'s default value to `0` would also be an OK option (we still need the button for `animation-direction`, though). Let me know if you'd rather see that instead.

PR #651, opened 2018-03-06
Add examples for text-emphasis CSS properties

ddbeck:
This PR adds examples for:

* [`text-emphasis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis)
* [`text-emphasis-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-color)
* [`text-emphasis-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style)

I'm mostly satisfied with how these have turned out except for one thing: these emphasis marks aren't really intended for Latin script, but I don't really know any Japanese or Chinese text to use as an example. Open to suggestions there or on anything else. Thanks!

### DIFF
--- a/live-examples/css-examples/text-decoration/meta.json
+++ b/live-examples/css-examples/text-decoration/meta.json
@@ -50,6 +50,16 @@
             "title": "CSS Demo: text-decoration-style",
             "type": "css"
         },
+        "textEmphasis": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "cssExampleSrc":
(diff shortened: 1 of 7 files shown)


wbamberg:
These look good, but I wonder if we should ask for help to get a proper CJK example here.

ddbeck:
@wbamberg If you know someone to ask, I think it's worth doing, particularly because there are a few other properties that are also focused on CJK characters (like text-combine-upright or ruby-position) which might also benefit from some good sample text.

PR #652, opened 2018-03-06
Apply code in code block instead of choice element

kenrick95:
Reading the code, the previous behavior includes the text "Copy to clipboard" at `code` parameter of `applyCode` of `css-editor-utils.js`. 

So, for example, at [text-decoration-color example](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color), when switching to the option `text-decoration-color: #21ff21;`, the value of `code` parameter will be as follows

```
text-decoration-color: #21ff21;

Copy to clipboard
```

It doesn't cause any bug though, as somehow, `element.style.cssText` only applies the first line of `code` argument.

I submits this PR, because I think it is not the intended behavior. 😁 

Note: to see this "bug", you could put `console.log(code)` before `element.style.cssText = code;` line

### DIFF
--- a/js/editor-libs/css-editor-utils.js
+++ b/js/editor-libs/css-editor-utils.js
@@ -39,7 +39,7 @@ module.exports = {
         codeBlock.setAttribute('spellcheck', false);
         codeBlock.focus();
 
-        module.exports.applyCode(choice.textContent, choice);
+        module.exports.applyCode(codeBlock.textContent, choice);
     },
     /**
      * Resets the default example to visible but, only if it is currently hidden


schalkneethling:
Thanks @kenrick95 - I reckon the browser just throws away the code that is not valid, but this is much better. Nice catch!

schalkneethling:
Hey @kenrick95 - apologies for not having merged this one. We have so many live examples on MDN Web Docs now that we have to be super careful what we merge. This looks good and will most likely be 100% fine to merge. I just want to do some additional testing to be as close to 100% sure :)

Thanks for understanding and your continued contributions.

kenrick95:
No worries 😁 

That means this repo needs more automated tests right? 😅

kenrick95:
I'm wondering if this is still pending anything? Thanks 😅

schalkneethling:
@kenrick95 I just wanted to make sure I do a bunch of testing to make absolutely sure all is good as this is now live all over MDN ;)

PR #654, opened 2018-03-06
Add example for transform-style CSS property

ddbeck:
This PR adds an example for [`transform-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-style). This is kinda a tough one to illustrate. Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/transforms/meta.json
+++ b/live-examples/css-examples/transforms/meta.json
@@ -169,6 +169,16 @@
             "title": "CSS Demo: transform-origin",
             "type": "css"
         },
+        "transformStyle": {
+            "baseTmpl": "tmpl/live-css-tmpl.html",
+            "cssExampleSrc":
+                "../../live-examples/css-examples/transforms/transform-style.css",
+            "exampleCode":
+                "live-examples/css-examples/transforms/transform-style.html",
+            "fileName": "transform-style.html",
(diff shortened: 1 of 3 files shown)


ddbeck:
The question in #648 about infinite animations might need to be answered for this one too, or the animation dropped.

wbamberg:
I think this is tricky to illustrate partly because, if I understand correctly, the property is applied to the parent, but the visible effect is determined by properties set on the child. So it might be worth showing the CSS for the child in the output.

I also think the 3D transform in your example is a little complex: to me at least this is not a simple concept, so using a very obvious 3D transform helps me to understand.

What do you think of something like:

screen shot 2018-03-07 at 12 53 09 pm

screen shot 2018-03-07 at 12 53 18 pm

... which is given by something like 

```
.layer {
    transform: perspective(200px) rotateY(30deg);
    background: #f69d3c;
    border-radius: 1rem;
}

.numeral {
    color: black;
    background-color: #e66465;
    margin: 1rem;
    font-size: 1.5rem;
    border-radius: .2rem;
    padding: .2rem;
    transform: rotate3d(1, 1, 1, 45deg);
}
```

ddbeck:
I like this! I think I got a bit hung up on emphasizing the third dimension and not confusing it with two dimensional transforms.

PR #656, opened 2018-03-09
Add example for font-feature-settings CSS property

ddbeck:
This PR adds an example for [`font-feature-settings`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings). If this looks like a clone of `font-variant` that's because it is—it does almost the exact same thing, with a different syntax. Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/fonts/font-feature-settings.css
+++ b/live-examples/css-examples/fonts/font-feature-settings.css
@@ -1,7 +1,7 @@
 @font-face {
     font-family: 'Fira Sans';
     src: local('FiraSans-Regular'),
-         url('../../../media/fonts/FiraSans-Regular.woff2') format('woff2');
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/fonts/font-feature-settings.css
```diff
@@ -0,0 +1,22 @@
+@font-face {
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
+         url('../../../media/fonts/FiraSans-Regular.woff2') format('woff2');
```
wbamberg: Is this going to have the same problem that you fixed in https://github.com/mdn/interactive-examples/pull/661?

ddbeck: It certainly would. Fixed!

PR #659, opened 2018-03-10
Add example for (-moz-) appearance CSS property

ddbeck:
This PR adds [`-moz-appearance`](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance). This one's experimental and not-well supported yet, but it was on the to-do list so I did a pretty minimal example here. Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/basic-user-interface/appearance.html
+++ b/live-examples/css-examples/basic-user-interface/appearance.html
@@ -1,0 +1,39 @@
+<section id="example-choice-list" class="example-choice-list">
+  <div class="example-choice">
+    <pre><code class="language-css">-moz-appearance: none;
+appearance: none;</code></pre>
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
+
+  <div class="example-choice">
+    <pre><code class="language-css">-moz-appearance: button;
+apperance: button;</code></pre>
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
+
+  <div class="example-choice">
+    <pre><code class="language-css">-moz-appearance: checkbox;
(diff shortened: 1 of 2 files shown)


live-examples/css-examples/basic-user-interface/appearance.html
```diff
@@ -0,0 +1,39 @@
...
+    <pre><code class="language-css">-moz-appearance: none;
+appearance: none;</code></pre>
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
+
+  <div class="example-choice">
+    <pre><code class="language-css">-moz-appearance: button;
+apperance: button;</code></pre>
```
wbamberg: apperance -> appearance, here and in the next two occurrences.

But wouldn't it be better to use `-webkit-appearance` throughout? Then it would also work in Chrome.

live-examples/css-examples/basic-user-interface/appearance.html
```diff
@@ -0,0 +1,39 @@
+<section id="example-choice-list" class="example-choice-list">
```
wbamberg: This is missing `data-property` but I'm not sure what we should put for it. Maybe we should omit it for now, and when the editor can handle prefixing properties we could use `appearance`.

live-examples/css-examples/basic-user-interface/appearance.html
ddbeck: OK, yeah. In place of the non-functioning `appearance` (and typo variant), I've put in the `-webkit-` prefixed versions.

live-examples/css-examples/basic-user-interface/appearance.html
ddbeck: That was my thinking with omitting it in the first place. I'll leave it for now.

ddbeck:
Thanks for the feedback, @wbamberg! Let me know if you want to see any other changes.

PR #660, opened 2018-03-12
Add example for text-justify CSS property

ddbeck:
This PR adds an example for [`text-justify`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-justify). I think this one works pretty well for now, though like #651, I think it might benefit from some added CJK text to show the inter-character spacing option a bit better (though that could come in a later PR). Let me know if you want to see any changes. Thanks!

### DIFF
--- a/live-examples/css-examples/text/meta.json
+++ b/live-examples/css-examples/text/meta.json
@@ -72,6 +72,15 @@
             "title": "CSS Demo: text-indent",
             "type": "css"
         },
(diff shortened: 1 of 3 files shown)


wbamberg:
> like #651, I think it might benefit from some added CJK text to show the inter-character spacing option a bit better (though that could come in a later PR).

Yes, I agree, but also agree with merging this as it is and optionally improving it later :).

PR #662, opened 2018-03-12
Add example for font-size-adjust CSS property

ddbeck:
This PR adds an example for [`font-size-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust). Let me know if you want to see any changes. Thanks!


live-examples/css-examples/fonts/font-size-adjust.css
```diff
@@ -0,0 +1,24 @@
...
+    font-family: serif;
+    font-size: 20px;
+}
+
+.unadjusted code {
+    font-size-adjust: .45;
+}
+
+#example-element {
+    font-family: 'Fira Sans', sans-serif;
```
wbamberg: I'm not sure why this example uses different `font-family` and `font-size` values for adjusted and unadjusted. I think it would be better to keep everything the same except the `font-size-adjust` value.

Or am I missing something?

live-examples/css-examples/fonts/font-size-adjust.css
```diff
@@ -0,0 +1,24 @@
...
+    flex-direction: column;
+}
+
+.unadjusted {
+    font-family: serif;
+    font-size: 20px;
+}
+
+.unadjusted code {
+    font-size-adjust: .45;
```
wbamberg: It seems like this sets the value for the "font-size: 20px;" text, but I'm not sure why we're doing this.

I guess I'm also not sure what the meaning of the "font-size: 20px;" text is: what is it referring to? I thought initially it was referring to both "This text is unadjusted." and "This text is adjusted.", but "This text is adjusted." is only 16px.

live-examples/css-examples/fonts/font-size-adjust.css
ddbeck: OK, there's a few issues here:

- The `font-size` was a mistake on my part (the example text was all supposed to have the same `font-size` and I goofed up something fierce with git stash).
- The different `font-family` properties were meant to illustrate the intended use (as described in the spec) for dealing with `font-size` discrepancies caused by different `font-family` values.
- I'll try it again with one `font-family` only, but I wonder if it would confuse why you want to use this property—that it's some alternate to `font-size` or something like that.

live-examples/css-examples/fonts/font-size-adjust.css
ddbeck: Yeah, this is compounding my earlier mistake. I wanted the `font-size` of everything to be the same, but the `<code>` text was enormous. This is going to go away based on your other feedback.

ddbeck:
> One alternative, rather than having it there, would be to have font-size in the input choices

Yeah, I'll do this with that in mind. I was trying to avoid overloading the input side of things, but apparently my solution was unclear.

ddbeck:
OK, I've pushed a change with `font-size` alongside `font-size`. Let me know what you think of it!

wbamberg:
> The different font-family properties were meant to illustrate the intended use (as described in the spec) for dealing with font-size discrepancies caused by different font-family values.
>
> I'll try it again with one font-family only, but I wonder if it would confuse why you want to use this property—that it's some alternate to font-size or something like that.

I'm sorry @ddbeck , I did not understand this point, and now I do, I think you were right to show two examples with different `font-family` values. I think it would be good to be explicit about the `font-family` and the `font-size` used, and possibly to show the unadjusted value as well. This is quite like the existing live sample:

```
This text uses the Verdana font (10px).
This text uses the Times font (10px, unadjusted).
This text uses the Verdana font (10px, adjusted).
```

Then we should revert the input choices to omit the `font-size`.

Anyway. I'm sorry to change my mind about this, I should have looked into this property more deeply in the first place. If you want to try making these changes, or similar ones to make it closer to how it was before, please do. But if you're happy with how it is now I'll merge it :).

ddbeck:
No problem, @wbamberg! I'll make the changes like you suggest. I think it's a good approach, to be more explicit about the character of the text than I was in my original attempt.

ddbeck:
OK, let me know what you think of attempt 3. Thanks!

live-examples/css-examples/fonts/font-size-adjust.html
```diff
@@ -0,0 +1,45 @@
...
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+</section>
+
+
+<div id="output" class="output hidden">
+    <section id="default-example">
+        <p clsas="times">This text uses the Times font (10px).</p>
```
wbamberg: typo: `clsas` -> `class`

ddbeck:
Fixed. Thank you, @wbamberg!

PR #678, opened 2018-03-21
Update object-assign.html

Abhicoding:
Just thought this tiny change would make the example easier to understand.

### DIFF
--- a/live-examples/js-examples/object/object-assign.html
+++ b/live-examples/js-examples/object/object-assign.html
@@ -5,7 +5,7 @@
   c: 3
 };
 
-const object2 = Object.assign({}, object1);
+const object2 = Object.assign({c: 4}, object1);
 
 console.log(object2.c);
 // expected output: 3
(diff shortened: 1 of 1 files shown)


wbamberg:
This is nice! What do you think about also adding a property to the first object that *won't* be overridden by the second? e.g.

```
const object2 = Object.assign({c: 4, d: 1}, object1);

console.log(object2.c, object2.d);
// expected output: 3 1
```

?

Abhicoding:
Added the improvements you mentioned.

Abhicoding:
Thanks a lot!

wbamberg:
The first argument to `Object.assign` is the target, the second is one or more sources. Values are copied from the source(s) into the target.

In this case the target is `{c: 4, d: 5}` and the source is `{a: 1, b: 2, c: 3}`, so in the result, `c` is 3.

>  Could you elaborate a little bit about why this is happening in the explanation above it?

I'm reluctant to include a lot of explanatory content in these examples. They are examples, not documentation - if people want to know the "why" of what the examples are doing, they should consult the docs page itself.

Issue #690, opened 2018-04-02
Add output label to HTML output panel


schalkneethling:
As part of the user interview feedback for the HTML interactive editor, we learned that it is not immediately clear that the output pane of the editor is connected to the editor on the left.

One suggestion to resolve this was to add an output label to the output panel.

PR #691, opened 2018-04-03
Fix issue #690, add output label to tabbed editor output pane

schalkneethling:
@wbamberg This is what it looks like in the context of a MDN page:

screen shot 2018-04-02 at 16 56 22

### DIFF
--- a/css/tabbed-editor.css
+++ b/css/tabbed-editor.css
@@ -36,6 +36,19 @@
     width: 40%;
 }
 
+.output-label {
+    position: absolute;
+    background-color: #333;
+    color: #83d0f2;
+    right: 0;
+    margin: 0;
+    padding: 3px 10px;
+    width: 70px;
+    height: 25px;
+    font-size: .9rem;
+    letter-spacing: .035rem;
+}
+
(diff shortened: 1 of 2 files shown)


wbamberg:
What do you think about "Output" instead of "output"? It looks inconsistent, especially right under "Reset".

Other nits, although they are strictly out of the scope for this bug:

* the bottom edge of the output doesn't align with the bottom edge of the editor
* the output looks very flat to me, what do you think about giving it something like `box-shadow: inset .2px .2px 2px gray;`? nb I am not a designer.

Apart from that I think this is fine - at any rate the user testing results seemed to me clearly strong enough that we can go ahead with this editor and tweak it later if we want to.

schalkneethling:
> What do you think about "Output" instead of "output"?

Yeah, I had it like that, and then it looked a little odd... Agreed though, it is more consistent.

> Other nits, although they are strictly out of the scope for this bug:

I totally agree with these, and I was intent of making additional updates. I also agree that it is best added in a follow on pull request.

Thanks for the review @wbamberg

PR #692, opened 2018-04-06
Show that `Object.assign()` argument order matters.

asiplas:
Illustrate that when `Object.assign()` is called with arguments of multiple
objects with duplicate key(s), the last object's value will prevail.


asiplas:
https://www.ecma-international.org/ecma-262/6.0/#sec-object.assign

PR #697, opened 2018-04-11
Add <abbr> example

wbamberg:
Add a simple example for [`<abbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr). Feedback welcome!

I think we should wrap lines in the editor, what do you think?

### DIFF
--- a/live-examples/html-examples/abbr.html
+++ b/live-examples/html-examples/abbr.html
@@ -1,0 +1,1 @@
+<p>You can use <abbr title="Cascading Style Sheets">CSS</abbr> to style your <abbr title="HyperText Markup Language">HTML</abbr>.</p>
--- a/live-examples/html-examples/css/abbr.css
+++ b/live-examples/html-examples/css/abbr.css
@@ -1,0 +1,3 @@
+abbr {
+    font-variant: all-small-caps;
+}
--- a/live-examples/html-examples/meta.json
+++ b/live-examples/html-examples/meta.json
@@ -1,5 +1,13 @@
 {
     "pages": {
+        "abbr": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode": "live-examples/html-examples/abbr.html",
+            "cssExampleSrc": "live-examples/html-examples/css/abbr.css",
+            "fileName": "abbr.html",
+            "title": "HTML Demo: <abbr>",
+            "type": "tabbed"
+        },
         "audio": {
             "baseTmpl": "tmpl/live-tabbed-tmpl.html",
             "exampleCode": "live-examples/html-examples/audio.html",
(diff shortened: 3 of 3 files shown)


live-examples/html-examples/meta.json
```diff
@@ -1,13 +1,21 @@
 {
     "pages": {
+        "abbr": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode": "live-examples/html-examples/abbr.html",
+            "cssExampleSrc": "live-examples/html-examples/css/abbr.css",
+            "fileName": "abbr.html",
+            "title": "HTML Demo: &lt;abbr&gt;",
```
schalkneethling: Ah, so I see it is a problem in the title. This will resolve that problem - https://github.com/mdn/bob/issues/18

live-examples/html-examples/meta.json
```diff
@@ -1,13 +1,21 @@
...
+        },
         "audio": {
             "baseTmpl": "tmpl/live-tabbed-tmpl.html",
             "exampleCode": "live-examples/html-examples/audio.html",
             "cssExampleSrc": "live-examples/html-examples/css/audio.css",
             "fileName": "audio.html",
             "title": "HTML Demo: audio",
             "type": "tabbed"
-      },
+        },
```
schalkneethling: Nit: unintended indent?

live-examples/html-examples/abbr.html
```diff
@@ -0,0 +1 @@
+<p>You can use <abbr title="Cascading Style Sheets">CSS</abbr> to style your <abbr title="HyperText Markup Language">HTML</abbr>.</p>
```
schalkneethling: I wonder, should we break this onto two lines to avoid horizontal scrolling?

live-examples/html-examples/css/abbr.css
```diff
@@ -0,0 +1,3 @@
+abbr {
+    font-variant: all-small-caps;
```
schalkneethling: I wonder, should we bump the `font-size` here? Overall the text is pretty small, and specifically, it is hard to notice the dotted underline added for `abbr`

screen shot 2018-04-12 at 10 01 06

Perhaps `font-size: 1.5rem;`?

screen shot 2018-04-12 at 10 02 29

live-examples/html-examples/abbr.html
wbamberg: I thought it might be better for the editor to wrap lines, as I suggested in https://github.com/mdn/interactive-examples/issues/699. Since the editor is resizable, it's hard to know at what width we should wrap, if we have to do it manually.

live-examples/html-examples/meta.json
wbamberg: No, I think this is fixing a previous indent error, isn't it? (see the current master: https://github.com/mdn/interactive-examples/blob/master/live-examples/html-examples/meta.json)

live-examples/html-examples/css/abbr.css
wbamberg: You're right that it is too small. But isn't it weird to use small caps and then bump the font size? Perhaps it would be better to use some alternative style than small caps, like a different text color?

PR #701, opened 2018-04-11
Add <strong> example

wbamberg:
Also a good case for wrapping, I guess.

### DIFF
--- a/live-examples/html-examples/strong.html
+++ b/live-examples/html-examples/strong.html
@@ -1,0 +1,1 @@
+<p>... the most important rule, the rule you can never forget, no matter how much he cries, no matter how much he begs: <strong>never feed him after midnight</strong>.</p>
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/strong.html
```diff
@@ -0,0 +1 @@
+<p>... the most important rule, the rule you can never forget, no matter how much he cries, no matter how much he begs: <strong>never feed him after midnight</strong>.</p>
```
schalkneethling: Here as well, wondering is it makes sense to break onto two lines to avoid horizontal scrolling.

wbamberg: Same answer as here: https://github.com/mdn/interactive-examples/pull/697#discussion_r181115779

Of course if making the editor wrap is not practical/sensible, we can break manually.

PR #703, opened 2018-04-12
Encode '<' and '>' in CONTRIBUTING

wbamberg:
Another attempt at https://github.com/mdn/interactive-examples/pull/702.


schalkneethling:
Oh wait, I see this is actually in the docs but, it is probably not a bad idea to encode the titles anyhow.

PR #709, opened 2018-04-13
Add <optgroup> example

wbamberg:
Please let me know if you think this is too much! It does have a slightly serious point, though, that in user testing some people thought examples like this were part of the page.


wbamberg:
Added an image from Unsplash with acceditation, as in https://github.com/mdn/interactive-examples/pull/712.

live-examples/html-examples/css/optgroup.css
```diff
@@ -0,0 +1,10 @@
+.output {
+    background: top/cover no-repeat url("/media/examples/dino2.jpg");
```
schalkneethling: Nit: Code Style: single quotes

live-examples/html-examples/css/optgroup.css
```diff
@@ -0,0 +1,10 @@
...
+    position: relative;
+}
+
+footer {
+    font-size: .8rem;
+    position: absolute;
+    bottom: 10px; left: 10px;
```
schalkneethling: Nit: Code Style: Each selector on its own line

live-examples/html-examples/css/optgroup.css
schalkneethling: File name should be `dino.jpg`

PR #712, opened 2018-04-13
Add background image to datalist example

wbamberg:
This adds a background image to the example. Some of the user testing participants seemed to have a little trouble recognizing the output pane in this example. The background image is supposed to help with that.

I'm interested in your feedback about whether this is a good idea!

Also, I wanted to credit the photographer, but wasn't sure whether this is the best approach, or whether `footer` is the best element to choose. Probably not.


live-examples/html-examples/css/datalist.css
```diff
@@ -1,8 +1,10 @@
...
+    background: top/cover no-repeat url("/media/examples/ice-cream.jpg");
+    position: relative;
 }
 
-input {
-    width: 10rem;
+footer {
+    font-size: .8rem;
+    position: absolute;
+    bottom: 10px; right: 10px;
```
schalkneethling: Nit: Code Style: Should be separate lines.

PR #718, opened 2018-04-16
fix: scroll output when content overflows

schalkneethling:
Scrolls output container when content overflows.


wbamberg:
@schalkneethling , it looks like this one was already in https://github.com/mdn/interactive-examples/pull/719/, yes?

schalkneethling:
@wbamberg Indeed, looks like it somehow snuck in there. Weird. Anyhow, closing.

PR #724, opened 2018-04-16
Remove hard-coded 'height'

MatsPalmgren:
Hard-coded 'height' is bad web design because they cause bugs like these:
https://bugzilla.mozilla.org/show_bug.cgi?id=1453903

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/box-shadow.css
+++ b/live-examples/css-examples/backgrounds-and-borders/box-shadow.css
@@ -4,6 +4,5 @@
     padding: 0;
     border: 2px solid #333;
     width: 80%;
-    height: 100px;
     text-align: center;
 }


schalkneethling:
Thank you again for the PR @MatsPalmgren - I do agree that setting the height is not ideal. In this case we sometimes need to because out viewport is "fixed", and so we need to make the best use of the space we have.

Removing the height here definitely does not make the example any less effective but, there are surely others where removing the height will not be as simple.

Before I go ahead and merge this one, I wonder if @wbamberg has some additional thoughts?

schalkneethling:
r+ Thanks @MatsPalmgren - Does this mean we can also close the bug on Bugzilla?

MatsPalmgren:
Thanks for merging; I've closed the Bugzilla bug report now.

PR #727, opened 2018-04-17
fix: update to tabbed editor ui

schalkneethling:
More UI updates to the tabbed editor.


schalkneethling:
@wbamberg Thank you for suggestion. I think it might be a bit light for the border. Also, this way there is some continuity with the JS editor.

PR #790, opened 2018-04-23
Add <u> example

wbamberg:


schalkneethling:
@wbamberg Great example. Looking at it at a smaller viewport, the line height seems to be a bit tight:

screen shot 2018-04-24 at 14 35 44

Perhaps add a line height like so:

```
p {
    font: 1rem/1.5 'Fira Sans', sans-serif;
}
```

schalkneethling:
r+w/c

schalkneethling:
On second though, it seems that it will be beneficial to have a line height of 1.5 as standard for the output area so, will add it globally.

PR #793, opened 2018-04-24
adding form element example

chrisdavidmills:
as per https://github.com/mdn/interactive-examples/issues/746.

However, there are two problems with this example, one minor and one major:

* Minor: I had to add some extraneous classes to namespace the block elements in my example, otherwise my CSS was clashing with the styling of the editor UI, and messing things up. How I've got it now is not ideal, but it's working.
* Major: I think it must be because there's a form in the example, but whenever I click on the CSS, it seems to submit the form and therefore refresh the page, meaning you can't see or edit the CSS. I tried adding some preventDefault() code inline on the form element, but this didn't seem to work either. It makes me think that this is probably going to be a problem with any example that features a form element. Could you add a function of some kind that preventDefault()s all forms included in examples?

### DIFF
--- a/live-examples/html-examples/forms/meta.json
+++ b/live-examples/html-examples/forms/meta.json
@@ -21,7 +21,7 @@
             "exampleCode": "live-examples/html-examples/forms/form.html",
             "cssExampleSrc": "live-examples/html-examples/forms/css/form.css",
             "fileName": "form.html",
-            "title": "HTML Demo: <form>",
+            "title": "HTML Demo: &lt;form&gt;",
             "type": "tabbed"
         },
         "optgroup": {
--- a/live-examples/html-examples/forms/meta.json
+++ b/live-examples/html-examples/forms/meta.json
@@ -16,6 +16,14 @@
             "title": "HTML Demo: &lt;fieldset&gt;",
             "type": "tabbed"
         },
+        "form": {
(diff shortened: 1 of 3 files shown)


schalkneethling:
@chrisdavidmills The first one is known and being actively worked on. The second one is truly weird. Technically this form should not submit unless both fields are filled in :-/

Looking into it.

live-examples/html-examples/forms/meta.json
```diff
@@ -16,6 +16,14 @@
             "title": "HTML Demo: &lt;fieldset&gt;",
             "type": "tabbed"
         },
+        "form": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode": "live-examples/html-examples/forms/form.html",
+            "cssExampleSrc": "live-examples/html-examples/forms/css/form.css",
+            "fileName": "form.html",
+            "title": "HTML Demo: <form>",
```
schalkneethling: @chrisdavidmills Ok phew, so that took a while :) This is the problem. This branch does not yet have the auto escaping in BoB so, the `form` is rendered and whenever you click on the tabs, it submits.

So, simple escape the opening `<` and all will be good with the world again :) Thanks Chris!

chrisdavidmills:
OK, done!

PR #794, opened 2018-04-24
adding label element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/747


live-examples/html-examples/forms/css/label.css
```diff
@@ -0,0 +1,15 @@
+.label-example {
+    font-family: sans-serif;
+}
+
+p {
+  font-size: 0.8rem;
```
schalkneethling: Nit: Code style: omit the leading zero for 0 based values `0.8rem` => `.8rem`

PR #795, opened 2018-04-24
adding legend element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/748

### DIFF
--- a/live-examples/html-examples/forms/css/legend.css
+++ b/live-examples/html-examples/forms/css/legend.css
@@ -1,0 +1,19 @@
+.output {
+    font: 1rem 'Fira Sans', sans-serif;
+}
+
+@font-face {
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/forms/css/legend.css
```diff
@@ -0,0 +1,19 @@
...
+    url('/media/fonts/FiraSans-Regular.woff2') format('woff2');
+}
+
+input {
+    margin: .4rem;
+}
+
+legend {
+  background-color: #000000;
+  color: #ffffff;
```
wbamberg: For hex codes, use the short form: https://mdn.github.io/mdn-fiori/patterns/css/formatting/#hex-color-codes

wbamberg: Also, do you think it would be good to have this rule at the top of the file? Since it applies to the element we are demonstrating?

chrisdavidmills:
Hrm, I managed to pick up a conflict somehow. I've fixed it, but let me know if it everything looks OK.

PR #798, opened 2018-04-24
adding meter element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/749


live-examples/html-examples/forms/css/meter.css
```diff
@@ -0,0 +1,16 @@
+.meter-example {
+    display: table;
+}
+
+.meter-example div {
+    display: table-row;
+}
+
+label, meter {
```
wbamberg: Formatting nit "When using multiple selectors in a rule declarations, give each selector its own line": https://mdn.github.io/mdn-fiori/patterns/css/formatting/#general

live-examples/html-examples/forms/meter.html
```diff
@@ -0,0 +1,16 @@
+<div class="meter-example">
+  <div>
```
wbamberg: Formatting nit: we seem to be using 4-space indent for the HTML.

PR #800, opened 2018-04-24
adding option element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/750 and https://github.com/mdn/interactive-examples/issues/753

Note — a select example already existed, so I just copied it for the option example, as it worked well for both. Too cheap-assed? Let me know if you want more ;-)


wbamberg:
I think it's fine to use a single example for more than one element. But this `select` example was really just thrown together for the user testing, and it might be worth considering how we could pretty it up. 

For similar examples like [`optgroup`](https://interactive-examples.mdn.mozilla.net/pages/tabbed/optgroup.html) I added a background image, partly just to make it look more interesting, but also (hopefully) to help make it clear that this is the output of the example and not part of the page. Do you think it's worth doing something similar here? -> https://unsplash.com/search/photos/parrot

chrisdavidmills:
I think this makes sense. I've updated the example (for both select and option) to follow the same kind of pattern as yours, but with a cute hamster.

zbjornson:
At the risk of sounding boring, can I propose removing the background image? The background + footer are a bit distracting, and at first I expected the `<select>` to change the background image to the selected pet. Happy to open a PR.

wbamberg:
We originally added these because when the styling of the output pane was too plain, some people thought that it was part of the page UI itself. So we wanted to make it stand out a little. But perhaps we have gone too far! and something more subtle would work as well without being so distracting.

I'd be happy to look at a PR for this :).

PR #801, opened 2018-04-24
adding progress element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/752


live-examples/html-examples/forms/css/progress.css
```diff
@@ -0,0 +1,16 @@
+.progress-example {
+    display: table;
+}
+
+.progress-example div {
+    display: table-row;
+}
+
+label, progress {
```
wbamberg: Formatting nit "When using multiple selectors in a rule declarations, give each selector its own line": https://mdn.github.io/mdn-fiori/patterns/css/formatting/#general

live-examples/html-examples/forms/progress.html
```diff
@@ -0,0 +1,14 @@
+<div class="progress-example">
+  <div>
```
wbamberg: Formatting nit: we seem to be using 4-space indent for the HTML.

PR #804, opened 2018-04-24
Add <footer> example

2alin:
Hi, this is my first contribution to this repo. Hope I'm doing it right.

### DIFF
--- a/live-examples/html-examples/content-sectioning/css/footer.css
+++ b/live-examples/html-examples/content-sectioning/css/footer.css
@@ -1,0 +1,42 @@
+.output {
+    background-color: #f9f9fa;
+}
+
+h1 {
+    font-size: 1.8em;
+    margin-bottom: 0.5em;
+}
+
+footer {
+    display: flex;
+    padding: 1em .5em;
+    margin: 1.5em 0 0;
+    font-size: 0.9em;
+    font-family: sans-serif;
+    background-color: #45a1ff;
+    color: #fff;
+}
+
+footer p {
+    flex: 1;
+    margin: 0;
+}
+
+footer ul {
+    flex: 1;
+    display: flex;
+    justify-content: space-around;
+    margin: 0;
+    list-style: none;
+}
+
+footer a {
+    text-decoration: none;
+    padding: 2px;
+}
+
+footer a:link,
+footer a:visited{
+    color: #fff;
+}
+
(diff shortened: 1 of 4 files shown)


live-examples/html-examples/content-sectioning/css/footer.css
```diff
@@ -0,0 +1,42 @@
+.output {
+    background-color: #f9f9fa;
+}
+
+h1 {
+    font-size: 1.8em;
+    margin-bottom: 0.5em;
```
schalkneethling: Nit: Code style: Zero based values should omit the leading 0 ~ `0.5em` => `.5em`

live-examples/html-examples/content-sectioning/css/footer.css
```diff
@@ -0,0 +1,42 @@
...
+}
+
+footer {
+    display: flex;
+    padding: 1em .5em;
+    margin: 1.5em 0 0;
+    font-size: 0.9em;
```
schalkneethling: Nit: Code style: Zero based values should omit the leading 0 ~ `0.5em` => `.5em`

Also, prefer `rem` based units for font sizes.

2alin:
Done, changed to `rem` units in font size, `px` in padding/margin and no more leading 0. Also changed back the line in package.json

2alin:
@schalkneethling  and if you think html/css is too much. I could remove the anchor tags in the footer, so the elements there end up as list items only (no clickable). That will reduce horizontal space in the html, and css will go from 40 lines to 30 lines of code.
What do you think?

live-examples/html-examples/content-sectioning/footer.html
```diff
@@ -0,0 +1,15 @@
...
+    <li>Wear a tall pointed hat. </li>
+    <li>Befriend the eagles.</li>
+    <li>Have I mentioned the beard?</li>
+</ol>
+<footer>
+    <p>&copy; 2018 Gandalf</p>
+    <ul>
+        <li><a href="http://example.com/terms">Terms</a></li>
+        <li><a href="http://example.com/blog">Blog</a></li>
+        <li><a href="http://example.com/about">About</a></li>
```
schalkneethling: @2alin Yes, let's make these just list items. Then you can also remove lines 32-40 of the CSS file. Thanks so much.

2alin:
done, now html/css are both cleaner :wink:

wbamberg:
It's a great example :).

>  I really like this example but, it does use a lot of HTML and CSS. Is it to much?

As it is I think this is fine.

2alin:
Thank you (both) for the review :grinning:

Issue #754, opened 2018-04-19
[HTML Examples] Forms ~ textarea


schalkneethling:
Need an HTML example for the `textarea` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

## Contributing

https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md#contributing-an-html-example

PR #809, opened 2018-04-25
adding textarea element example

chrisdavidmills:
As per https://github.com/mdn/interactive-examples/issues/754

### DIFF
--- a/live-examples/html-examples/forms/textarea.html
+++ b/live-examples/html-examples/forms/textarea.html
@@ -7,7 +7,7 @@
     <div>
         <label for="advanced">Advanced example:</label>
         <textarea id="advanced" name="advanced"
-                  rows="3" cols="50" maxlength="200"
+                  rows="3" cols="33" maxlength="200"
                   wrap="hard">
 Textarea with sizing and wrap attribute (try values of hard, soft, and off to see how it affects wrapping). The maximum number of characters is constrained to 200 by the maxlength attribute.
         </textarea>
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/forms/textarea.html
```diff
@@ -0,0 +1,40 @@
...
+        <textarea id="sizing-wrapping" name="size"
+                  rows="5" cols="50" wrap="hard">
+Textarea with explicit sizing and wrap attribute to control wrapping. try values of hard, soft, and off in the wrap attribute to see how it affects the text wrapping
+        </textarea>
+    </div>
+    <div>
+        <label for="min-max">minlength/maxlength example:</label>
+        <textarea id="min-max" name="min-max"
+                  rows="4" cols="50"
+                  minlength="3" maxlength="45">
```
wbamberg: minlength did not work for me here (Chrome 66, Firefox 61). At least, it let me delete everything in the box. Is that the intended behavior?

chrisdavidmills:
thanks for the feedback @wbamberg! I have simplified it quite a bit, getting rid of all the examples except the basic and sized ones (which I renamed to advanced).

Other points to note:

* I added a maxlength constraint to the advanced one, to show how that works
* I reduced the height of it so it all fits into the output - no scrollbars. This has the added side effect that it makes the textarea itself scroll, and I think it is useful to see how that looks.
* I've removed minlength altogether — the trouble with it is that it behaves differently to maxlength — it doesn't stop people reducing the number of characters below the specified minlength, but the value is invalid if that is the case. So to demonstrate this would require an extra set of complication, setting up valid/invalid styles, etc. And I'm not sure if it worth it really. It would make things additionally complicated.

chrisdavidmills:
OK, makes sense - column size reduced!

PR #820, opened 2018-04-25
adding caption element example

chrisdavidmills:
I was thinking of adding some fun images to this one, but decided to leave it because of potential copyright issues. Toy companies tend to be very protective of brand.


live-examples/html-examples/table-content/caption.html
```diff
@@ -0,0 +1,28 @@
+<table>
+    <caption>He Man and Skeletor facts</caption>
```
wbamberg: It's He-man :)

live-examples/html-examples/table-content/caption.html
```diff
@@ -0,0 +1,28 @@
...
+    <tr>
+        <th scope="row">Weapon</th>
+        <td>Power Sword</td>
+        <td>Havoc Staff</td>
+    </tr>
+    <tr>
+        <th scope="row">Dark secret</th>
+        <td>Expert florist</td>
+        <td>Cries at romcoms</td>
+    </tr>
```
wbamberg: I think we should omit one of these rows (but not "Dark secret", we should keep that one).

As it is, at narrower widths the caption is partly outside its container.

live-examples/html-examples/table-content/css/caption.css
```diff
@@ -0,0 +1,33 @@
...
+    border-collapse: collapse;
+    border: 2px solid rgb(200, 200, 200);
+    letter-spacing: 1px;
+    font-family: sans-serif;
+    font-size: 0.8rem;
+}
+
+td, th {
+    border: 1px solid rgb(190, 190, 190);
+    padding: 10px 20px;
```
wbamberg: I think we should reduce horizontal padding to 10px, otherwise the table overflows horizontally at narrower widths.

wbamberg:
> Do you think we should use some gross font for the `<th>` content?

Maybe not, I don't trust my own judgement any more :)

chrisdavidmills:
I have answered you nits, and added some fun font effects for the character names. WYT?

wbamberg:
This looks fantastic - except that even at quite generous widths the caption is pushed below the fold:

screen shot 2018-04-26 at 11 03 51 am

I think the main cause here is that He-man's font, so I'd suggest reducing his font size (and reducing Skeletor's a bit, too, but not so much because that one is hard to read at small sizes). This:

```
.heman {
    font: 1.2rem molot;
    text-shadow: 1px 1px 1px #fff,
                 2px 2px 1px #000;
}

.skeletor {
  font: 1.5rem rapscallion;
  letter-spacing: 3px;
  text-shadow: 1px 1px 0px #fff,
               0 0 9px #000;
}
```
...gives us:

screen shot 2018-04-26 at 11 05 47 am

What do you think?

PR #821, opened 2018-04-25
adding col and colgroup examples

chrisdavidmills:
Boom! Just added onto my he man/skeletor example for this one.


wbamberg:
I think this example doesn't make enough sense to me - I don't know why you'd want to put a green border round the first two cols and a red border round the third, and the way the green overlaps the red looks messy.

How about something like:

```
    <colgroup>
        <col/>
        <col class="hero">
        <col class="villain">
    </colgroup>
```

```
.hero {
   background-color: #ebf8e1;
}

.villain {
   background-color: #ffe7e8;
}
```
(you have to remove the existing `tr:nth-child(even) td` and `tr:nth-child(odd) td` rules)

The only drawback is that it doesn't demonstrate `span`, but I prefer it, as it looks better and makes more sense.

Also the same comments from https://github.com/mdn/interactive-examples/pull/820 apply here.

wbamberg:
If you did this I'd also consider reordering the CSS so the `.hero` and `.villain` rules come first.

chrisdavidmills:
Yeah, I don't think these examples were very good on reflection. I've taken your advice and changed the theme, so I can now use col/colgroup to visually associate superheros and their sidekicks.

wbamberg:
This is nice, and I'm sorry to sound like a broken record, but - it won't fit at the target widths:

screen shot 2018-04-26 at 2 05 35 pm

It would be better I think to have just three columns - either one header column and two data columns, as for the caption example, or three data columns and no header column, if you want to demo `span`.

Also, and this is more of a nit, black on magenta/black on red is quite lurid and hard to read. I think it would be worth picking one of the "Lighter base shades" from https://mdn.github.io/mdn-fiori/patterns/scss/variables/.

chrisdavidmills:
More fiddling done, and new colors chosen! I was determined to keep both the headers and the four content columns, as I felt both were important to the example

chrisdavidmills:
Also fixed a merge conflict that had somehow appeared.

live-examples/html-examples/table-content/css/col.css
```diff
@@ -0,0 +1,29 @@
...
+
+table {
+    border-collapse: collapse;
+    border: 2px solid rgb(100, 100, 100);
+    letter-spacing: 1px;
+    font-family: sans-serif;
+    font-size: .7rem;
+}
+
+td, th {
```
wbamberg: Formatting nit: each selector should be on its own line.

live-examples/html-examples/table-content/css/colgroup.css
```diff
@@ -0,0 +1,29 @@
...
+
+table {
+    border-collapse: collapse;
+    border: 2px solid rgb(100, 100, 100);
+    letter-spacing: 1px;
+    font-family: sans-serif;
+    font-size: .7rem;
+}
+
+td, th {
```
wbamberg: Formatting nit: each selector should be on its own line.

PR #826, opened 2018-04-25
Add interactive example for dl element

a2sheppy:
Simple interactive example for `<dl>`.  Probably will use modifications of this same code for `<dt>` and `<dd>` -- but not the exact same example, because that can mess up SEO.

### DIFF
--- a/live-examples/html-examples/text-content/css/dl.css
+++ b/live-examples/html-examples/text-content/css/dl.css
@@ -1,0 +1,14 @@
+@font-face {
+    font-family: "Fira Sans";
+    src: local("FiraSans-Regular"),
+    url("/media/fonts/FiraSans-Regular.woff2") format("woff2");
+}
+
+p, dl {
+    font: 1rem "Fira Sans", sans-serif;
+}
+
+dl > dt {
+    font-weight: bold;
+    text-decoration: underline;
+}
--- a/live-examples/html-examples/text-content/css/dl.css
+++ b/live-examples/html-examples/text-content/css/dl.css
@@ -1,11 +1,12 @@
 @font-face {
-    font-family: "Fira Sans";
-    src: local("FiraSans-Regular"),
-    url("/media/fonts/FiraSans-Regular.woff2") format("woff2");
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
+    url('/media/fonts/FiraSans-Regular.woff2')
+    /* format("woff2"); */
 }
 
 p, dl {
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/text-content/css/dl.css
```diff
@@ -0,0 +1,14 @@
+@font-face {
+    font-family: "Fira Sans";
+    src: local("FiraSans-Regular"),
+    url("/media/fonts/FiraSans-Regular.woff2") format("woff2");
```
schalkneethling: Nit: Code style: Use single quotes. Here and above

live-examples/html-examples/text-content/css/dl.css
```diff
@@ -0,0 +1,14 @@
...
+}
+
+p, dl {
```
schalkneethling: Nit: Code style: Each selector on its own row

live-examples/html-examples/text-content/css/dl.css
```diff
@@ -0,0 +1,14 @@
...
+    font: 1rem "Fira Sans", sans-serif;
```
schalkneethling: Nit: Code style: Use single quotes

live-examples/html-examples/text-content/css/dl.css
```diff
@@ -0,0 +1,14 @@
...
+}
+
+dl > dt {
+    font-weight: bold;
+    text-decoration: underline;
+}
```
schalkneethling: Nit: Missing empty line at end of file

live-examples/html-examples/text-content/css/dl.css
a2sheppy: Eep. That'll take reprogramming muscle memory, as in 12 years of documenting at Mozilla I've always used double quotes.  I'll change this. :)

live-examples/html-examples/text-content/css/dl.css
a2sheppy: This line was copied from another example (`<ol>`), so that one should be updated too.

live-examples/html-examples/text-content/css/dl.css
a2sheppy: Weird, my editor is configured to alert me if I do that. Fixing it.

live-examples/html-examples/text-content/css/dl.css
a2sheppy: Ignore this push. Something very weird went wrong, and the wrong stuff went up.

PR #828, opened 2018-04-26
[issue #767] [HTML Examples] Inline text semantics ~ br 

atopal:
Adds interactive example for https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br


live-examples/html-examples/inline-text-semantics/meta.json
```diff
@@ -10,6 +10,16 @@
...
+        "br": {
+                "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+                "exampleCode":
+                    "live-examples/html-examples/inline-text-semantics/br.html",
+                "cssExampleSrc":
+                    "live-examples/html-examples/inline-text-semantics/css/br.css",
+                "fileName": "br.html",
+                "title": "HTML Demo: <br>",
+                "type": "tabbed"
+            },
```
wbamberg: Indentation is off for these.

live-examples/html-examples/inline-text-semantics/css/br.css
```diff
@@ -0,0 +1,5 @@
+p {
+    font-size: 1rem;
+    font-family: sans-serif;
```
wbamberg: Total nit, but is there a reason to not combine these, like:

```
font: 1rem/24px sans-serif;
```

PR #839, opened 2018-04-27
Add <dd> example

a2sheppy:
Adds a simple example for the `<dd>` element.

### DIFF
--- a/live-examples/html-examples/text-content/meta.json
+++ b/live-examples/html-examples/text-content/meta.json
@@ -18,6 +18,16 @@
             "title": "HTML Demo: <dl>",
             "type": "tabbed"
         },
+        "dd": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/text-content/meta.json
```diff
@@ -18,6 +18,16 @@
...
+        "dd": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/text-content/dd.html",
+            "cssExampleSrc":
+                "live-examples/html-examples/text-content/css/dd.css",
+            "fileName": "dd.html",
+            "title": "HTML Demo: <dd>",
+            "type": "tabbed"
+        },
```
wbamberg: Please keep this file in alpha order.

a2sheppy: I see no reason why this should be confusing; yeah, it has info about HTML in it, but it's also correct, so it's not a problem per se, and being in the box should make it pretty clear that it's an example.

PR #840, opened 2018-04-27
Add input type=email example

wbamberg:
I've tried to show the most important and interesting attributes, without making things too complicated.


wbamberg:
> I'd make the first input have the required attribute too — there is no use submitting a contact form with no email entered at all! The secondary addresses can be optional.

Makes sense. I've done this, although it looks a bit weird because that gives it the `invalid` class, so it gets the red border. But maybe that's OK.

> The (fake) form feels a bit weird to me without a submit button. It is just something that I'd expect to see. But perhaps not a big deal.

I've not done this. It didn't come up in any other reviews, and it's not a complete form, just a part of one. And if it did have "Submit", what could it do? I think it would be more confusing to have a "Submit" button that did nothing.

> The legend looks wrongly placed inside Firefox (not vertically centered on the fieldset line), but it looks fine in Chrome. Weird.

This was apparently due to the `margin-bottom` it got. I've changed things a bit now and it has gone away. Good catch!

chrisdavidmills:
All looks good, thanks Will! r+

PR #844, opened 2018-04-29
Add <del> example

### LINKED ISSUE #735: [HTML Examples] Demarcating edits ~ del
Need an HTML example for the `del` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del

## Contributing

2alin:
Fixes #735


wbamberg:
This is a very nice example! You don't need `line-height: 1.5;`, because that's already set on`.output`
And I would make the text a little bigger with something like `font-size: 1rem;`, what do you think?

2alin:
It works for me. I prefer small text :sweat_smile: but I do agree that `1rem` will suit better here because of the short example.
And it's updated :wink:

PR #846, opened 2018-04-29
Add <hgroup> example

### LINKED ISSUE #732: [HTML Examples] Content sectioning ~ hgroup
Need an HTML example for the `hgroup` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup

## Contributing

2alin:
Fixes #732

### DIFF
--- a/live-examples/html-examples/content-sectioning/css/hgroup.css
+++ b/live-examples/html-examples/content-sectioning/css/hgroup.css
@@ -23,6 +23,7 @@ p {
 
 p::first-letter {
     font-size: 3.2rem;
+    line-height: .7;
     float: left;
     padding-right: 8px;
     padding-top: 4px;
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/content-sectioning/css/hgroup.css
```diff
@@ -0,0 +1,29 @@
...
+    font-family: Georgia, serif;
+    font-size: .95rem;
+}
+
+p::first-letter {
+    font-size: 3.2rem;
+    float: left;
+    padding-right: 8px;
+    padding-top: 4px;
+}
```
wbamberg: This looks weird in Chrome - I think perhaps you need to set `line-height`? https://css-tricks.com/snippets/css/drop-caps/#article-header-id-1

2alin: You are right, chrome needs `line-height` to make it look right.
This is how it looks now in my computer's chrome, after updating it. Could you check it again? Thanks!! 
screenshot from 2018-04-30 18-15-13

2alin:
Thank you for reviewing and merging it @wbamberg

PR #847, opened 2018-04-29
Add <ins> example

### LINKED ISSUE #736: [HTML Examples] Demarcating edits ~ ins
Need an HTML example for the `ins` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ins

## Contributing

2alin:
Fixes #736.
CSS got a little longer (35 lines) mainly because of the `+/-` decoration. Previously, in the `<del>` example, I used them as inline level and here I want them to use as block level.
I could remove the  `+/-` prefixes (which uses 15 lines of css), but it won't look as nice as it is now. Let me know what do you think.

### DIFF
--- a/live-examples/html-examples/demarcating-edits/css/ins.css
+++ b/live-examples/html-examples/demarcating-edits/css/ins.css
@@ -1,0 +1,35 @@
+del,
+ins {
+    display: block;
+    text-decoration: none;
+    position: relative;
+}
+
+del {
+    background-color: #fbb;
+}
+
+ins {
+    background-color: #d4fcbc;
+}
+
+del::before,
+ins::before {
+    position: absolute;
+    left: .5rem;
+    font-family: monospace;
+}
+
+del::before {
+    content: '−';
+}
+
+ins::before {
+    content: '+';
+}
+
+p {
+    margin: 0 1.8rem 0;
+    font-family: Georgia, serif;
+    font-size: 1rem;
+}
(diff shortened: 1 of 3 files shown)


wbamberg:
I'll let @schalkneethling review this one :), but just on this point:

> CSS got a little longer (35 lines) mainly because of the +/- decoration.

I think this is OK. The main point of the examples is the HTML page, and as long as that is short enough, I don't mind so much having to scroll in the CSS tab. As long as the CSS isn't unnecessarily long.

schalkneethling:
@2alin I wonder, it might be beneficial to also show the use of the `cite` and `datetime` attributes? I know these do not produce a visual indication but, might get folks to read about the attributes and note their value. Thanks again for all your work on this.

2alin:
@schalkneethling I agree with you. After looking at the example the reader will probably wonder why such attributes somehow don't show up , research a little and get to know the reason. The only think I would suggest is a reference or a comment in the guide article that indicates that those attributes are hidden data. Otherwise the reader will search in other place.

Now, I'm not sure if the `datetime` value I added: `TA-3001` is appropriate. I know it's not a valid date but I wanted to be consistent with the fictional story. If you think I should put a valid one, let me know and I will change it to a simple `2018-05` value.

Thank you for  the review! :smiley:

schalkneethling:
@2alin It would be better to change the `datetime` to a valid date as per the spec:
https://www.w3.org/TR/2011/WD-html5-20110525/common-microsyntaxes.html#valid-date-string-with-optional-time

2alin:
too bad :sweat_smile:, but it's changed :wink:

Issue #693, opened 2018-04-06
HTML editor size adjustments


wbamberg:
HTML is pretty wordy. The `<table>` example we currently have overflows the editor. I think we should  do the same thing we do for JS examples, with support for a `taller` class: https://github.com/mdn/interactive-examples/issues/408.

23 lines or about that should be enough, and if it turns out not to be we could revisit. @schalkneethling , what do you think?

schalkneethling:
Makes sense to me. Thanks for logging this @wbamberg

schalkneethling:
I am introducing three classes that will be usable on the editor itself, as well as via the macro:

```
.tabbed-shorter # 300px
.tabbed-standard # 370px
.tabbed-tall # 670px
```

PR #848, opened 2018-04-30
adding thead, tbody, and tfoot examples

chrisdavidmills:


wbamberg:
These examples are great, but I have a couple of concerns with them.

1) they overflow horizontally even at quite generous widths. This is what they would look like embedded on an MDN page, full-screen on my laptop (1440 px):

screen shot 2018-04-30 at 3 28 27 pm

2) The HTML is very long. Ideally I want these examples to work for someone who doesn't want to interact with them - meaning that they should be able to see the HTML for the element without having to scroll the window. HTML is pretty wordy, and we will probably need to support a "taller" variant of the editor (see https://github.com/mdn/interactive-examples/issues/693) - but even so I think we should try hard to keep the example to under 30 lines. These are 51 lines, which I think is too much. Apart from the scrolling thing, I think a long example makes it harder for someone to see and digest the element that's the subject of the demo. The mission for these examples is to be short, simple examples that give someone a quick understanding of the basics of the element.

chrisdavidmills:
Points taken on board. I've cut down the amount of HTML in the example to under 30 lines.

PR #864, opened 2018-05-01
Add html image

wbamberg:
This just steals the image from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image. I think that is OK because we will presumably replace the live example at the top of the page with this one.


wbamberg:
This all makes sense to me. I made all these changes except, again, the one about reordering the CSS.

Let me know what you think.

chrisdavidmills:
Yup, looks great. r+

PR #865, opened 2018-05-01
Add html input month

wbamberg:
This one also lifts various things from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month. 

In particular using some CSS on the `:valid` and `:invalid` pseudo-classes. Although that page also shows how to use `pattern`, I didn't use that because it doesn't seem to work very well (as that page also says):

> The best way to deal with dates in forms in a cross-browser way (until all of the major browsers have supported them for a while) is to get the user to enter the month and year in separate controls 

...but of course to demo that needs JS.


wbamberg:
> The only thing I'd suggest adding maybe is a pattern attribute on the input,

As I said in the comment, I did consider this, but didn't do it because our own docs page says that the best way to provide cross-browser support is to use separate controls:

> The best way to deal with dates in forms in a cross-browser way (until all of the major browsers have supported them for a while) is to get the user to enter the month and year in separate controls

The pattern as suggested in the docs page doesn't work that well - it doesn't stop people entering obviously dates like 2828-99, and it doesn't stop people entering dates that fail to meet the `min` attribute. So you're going to need JS validation anyway.

chrisdavidmills:
@wbamberg  OK, good reasoning. I forgot that I'd written that bit ;-) Ignore that comment then.

wbamberg:
So I just removed the `input + span {` rule here. Let me know if this is OK!

chrisdavidmills:
Perfect, r+

PR #866, opened 2018-05-01
Add html input number

wbamberg:
I've included a smattering of attributes, as well as the `:valid`/`:invalid` trick from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Validation.

I didn't use `min` or `max` because I think it would get complicated to express that in `placeholder` as well as `step`. But we could have a minimum number of eyes instead of a multiple, if that would be a better attribute to demo.


wbamberg:
Makes sense to have an example showing min and max, so I added that. Then I thought I should probably remove the "name" one, so there are no br elements left, so I didn't wrap them in divs.

I also tweaked the CSS a little so it works better at narrower widths.

Let me know what you think.

chrisdavidmills:
Yup, looks better. I still think an increment of 0.1 would be better on the height field, but I'm not that fussed about it.

So r+

wbamberg:
I didn't use 0.1 because it felt like it was too difficult to enter a number that was invalid. Is there a particular reason why you like 0.1 better?

chrisdavidmills:
I just felt that it was a more realistic increment for a height measurement, plus it is useful to show people that the value doesn't have to be an integer.

But your point about seeing invalid values more easily is a good one.

wbamberg:
Fair enough, I've changed it to "0.1". Thanks for the reviews Chrid!

PR #834, opened 2018-04-26
Add input-color example

wbamberg:
Most of the same comments apply as for https://github.com/mdn/interactive-examples/pull/832.

PR #867, opened 2018-05-02
Add html input password

wbamberg:


wbamberg:
> In the CSS, maybe put the input-related rulesets up at the top so they are easier to access?

This is the opposite of a previous review comment: https://github.com/mdn/interactive-examples/pull/834#discussion_r184638807 :).

wbamberg:
> It is not really good practice to break up the label/input pairs using <br>s — <div>s would be better. But I appreciate that you are probably trying to cut down on the number of lines.

And the amount of indentation too, given the horizontal constraints. You are right though. I will have a think about this. I'll need to go back and change the existing already-merged examples, like https://interactive-examples.mdn.mozilla.net/pages/tabbed/input-color.html.

wbamberg:
Fixed br->div, left the ruleset order alone.

PR #868, opened 2018-05-02
Add html input radio

wbamberg:
Just some dinosaur choices.


wbamberg:
> Looking good, same comments apply about putting input rulesets first, and using divs, not brs.

I've changed to divs here (and also changed from dinosaurs to drones, just to keep the lines a bit shorter) but not changed the CSS order, as Schalk already suggested the current order in a previous review. I'm happy to reconsider (I did originally have the input rules first) but think we should all agree on that, if that's what we want to do.

@chrisdavidmills , please take another look.

schalkneethling:
@chrisdavidmills @wbamberg I get what you are saying here with regards to putting the input elements first. My thinking is that the focus of these examples are on the HTML so, the CSS is pretty secondary. There fore context and an additional learning experience.

With that, I felt that we might as well follow best practice. Putting your `@font-face` declarations at the top is how it is done. Well, I have never seen them anywhere else, unless they are in a completely separate CSS files of course ;)

Thanks for helping out on the review side @chrisdavidmills - Much appreciated.

PR #875, opened 2018-05-02
Add iframe example.

### LINKED ISSUE #740: [HTML Examples] Embedded content ~ iframe
Need an HTML example for the `iframe` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

## Contributing

https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md#contributing-an-html-example

sheeeng:
Fix #740.

### DIFF
--- a/live-examples/html-examples/embedded-content/css/iframe.css
+++ b/live-examples/html-examples/embedded-content/css/iframe.css
@@ -1,0 +1,15 @@
+footer {
+    font-size: .8rem;
+    position: absolute;
+    bottom: 10px;
+    right: 10px;
+}
+
+iframe {
+    border: 1px solid black;
+}
+
+.output {
+    background: #eee;
+    position: relative;
+}
--- a/live-examples/html-examples/embedded-content/iframe.html
+++ b/live-examples/html-examples/embedded-content/iframe.html
@@ -1,0 +1,8 @@
+<iframe id="inlineFrameExample"
+    title="Inline Frame Example"
+    width="360"
+    height="240"
+    src="https://www.openstreetmap.org/export/embed.html?bbox=-0.004017949104309083%2C51.47612752641776%2C0.00030577182769775396%2C51.478569861898606&amp;layer=mapnik">
+</iframe>
+
+<footer><a href="https://www.openstreetmap.org">Map</a> by OpenStreetMap Foundation (OSMF)</footer>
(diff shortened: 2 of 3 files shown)


live-examples/html-examples/embedded-content/iframe.html
```diff
@@ -0,0 +1,8 @@
+<iframe id="inlineFrameExample"
+    title="Inline Frame Example"
+    width="360"
```
wbamberg: This is going to overflow too easily, I think 300 would be better here.

live-examples/html-examples/embedded-content/iframe.html
```diff
@@ -0,0 +1,8 @@
...
+    height="240"
+    src="https://www.openstreetmap.org/export/embed.html?bbox=-0.004017949104309083%2C51.47612752641776%2C0.00030577182769775396%2C51.478569861898606&amp;layer=mapnik">
+</iframe>
+
+<footer><a href="https://www.openstreetmap.org">Map</a> by OpenStreetMap Foundation (OSMF)</footer>
```
wbamberg: I don't think we need the footer here, as the embedded OSM includes their copyright statement (they are quite particular about copyright: https://www.openstreetmap.org/copyright).

But I would also welcome @schalkneethling 's opinion here.

If we can omit the footer then of course we can also remove the footer CSS and also the `position: relative` statement.

schalkneethling: @wbamberg @sheeeng I agree, we do not need the additional footer here.

live-examples/html-examples/embedded-content/css/iframe.css
```diff
@@ -0,0 +1,15 @@
+footer {
+    font-size: .8rem;
+    position: absolute;
+    bottom: 10px;
+    right: 10px;
+}
+
+iframe {
+    border: 1px solid black;
```
schalkneethling: Suggestions: I would set the width of the `iframe` here to 100%

schalkneethling:
Thanks for your contributions @sheeeng 🎉

live-examples/html-examples/embedded-content/iframe.html
sheeeng: Fixed.

live-examples/html-examples/embedded-content/iframe.html
sheeeng: Footer removed.

live-examples/html-examples/embedded-content/css/iframe.css
sheeeng: Added `width: 100%` to `iframe`.

PR #876, opened 2018-05-02
Add object and param examples.

### LINKED ISSUE #741: [HTML Examples] Embedded content ~ object
Need an HTML example for the `object` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object

## Contributing

https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md#contributing-an-html-example

### LINKED ISSUE #742: [HTML Examples] Embedded content ~ param
Need an HTML example for the `param` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/param

## Contributing

https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md#contributing-an-html-example

sheeeng:
Fix #741.
Fix #742.


wbamberg:
Thanks again for an excellent contribution! I'd love to have an example for `<object>` that doesn't use Flash, but I wasn't sure what would be good. @chrisdavidmills , any ideas?

chrisdavidmills:
This is difficult — our options are quite limited here, as our stance is quite firmly anti-plugin. We don't want to be seen to be advocating Flash usage, and my article on the subject explains this fairly clearly:

https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Other_embedding_technologies

How about a PDF, as seen in this example?

https://mdn.github.io/learning-area/html/multimedia-and-embedding/other-embedding-technologies/object-pdf.html

You wouldn't do this very often, but I see this as less politically difficult than using Flash.

sheeeng:
PDF example works for `<object>`. But, none of the `<param>` works for PDF using below code.

```
<object type="application/pdf"
    data="/media/examples/pdf_open_parameters.pdf"
    width="300"
    height="200">
    <param name="page" value="5">
    <param name="scrollbar" value="1">
</object>

<footer><a href="https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_open_parameters.pdf">PDF</a> by Adobe</footer>
```

The Flash example [works](https://helpx.adobe.com/flash/kb/flash-object-embed-tag-attributes.html) with both `<object>` and `<param>` elements.

sheeeng:
What is the recommended way to resolve merge conflicts? Manually fix it here?

wbamberg:
> What is the recommended way to resolve merge conflicts? Manually fix it here?

Yes, that's how I have been doing it. It's usually only meta.json that's conflicted, and the fix is quite simple. You are welcome to fix merge conflicts yourself, otherwise when the PR is ready to merge I will resolve them :-).

PR #832, opened 2018-04-26
Add <input type="checkbox"> example

wbamberg:
This is pretty much modeled on the existing [`<legend>`](https://interactive-examples.mdn.mozilla.net/pages/tabbed/legend.html) example.

I've tried to arrange line breaks in the source so it stays readable at reasonable window widths.

The input examples are a bit different from the others because they're not of course separate elements. So I'm proposing we give them titles like `<input type="checkbox">` (reflective of the page title) and a filename like "input-checkbox.html".

Thanks for your feedback!

PR #877, opened 2018-05-03
Replace <br> with <div> in input examples

wbamberg:
This updates the existing input examples to use div instead of br to arrange elements.


live-examples/html-examples/input/checkbox.html
```diff
@@ -1,16 +1,22 @@
 <fieldset>
     <legend>Choose some monster features</legend>
 
-    <input type="checkbox"
-           id="scales" name="feature" value="scales" checked />
-    <label for="scales">Scales</label><br/>
+    <div>
+       <input type="checkbox" id="scales" name="feature"
```
chrisdavidmills: Lines 5-7 only have a 3 space indentation; should be 4.

live-examples/html-examples/input/checkbox.html
```diff
@@ -1,16 +1,22 @@
...
+              value="scales" checked />
+       <label for="scales">Scales</label>
+    </div>
 
-    <input type="checkbox"
-           id="horns" name="feature" value="horns" />
-    <label for="horns">Horns</label><br/>
+    <div>
+        <input type="checkbox" id="horns" name="feature"
+               value="horns" />
```
chrisdavidmills: One query here — I've just noticed that you're using the empty element trailing slash on the input elements. Is that part of your code guidelines for interactive elements? Generally on MDN, I don't think we do this. Something we should probably decide on.

wbamberg: This was also a previous review comment: https://github.com/mdn/interactive-examples/pull/832#discussion_r184636430. Unless one or other is clearly bad practice, I don't really mind which we do, but we should probably be consistent.

wbamberg:
Fixed the indentation, please take another look.

PR #878, opened 2018-05-03
Add html input range

wbamberg:
I didn't show ticks, due to dodgy browser support. But let me know if you'd like it.


wbamberg:
> Can we please have the volume control going from 0 to 11? not joking ;-)

I was totally going to do this! Then I forgot :).

wbamberg:
I also wanted to have labels for the cowbell like "more" and "even more". Gutted that this is not supported by browsers yet.

wbamberg:
Updated! Please take another look.

chrisdavidmills:
perfect! r+

PR #885, opened 2018-05-03
Add source example.

### LINKED ISSUE #744: [HTML Examples] Embedded content ~ source
Need an HTML example for the `source` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

## Contributing

sheeeng:
Fix #744.


live-examples/html-examples/embedded-content/meta.json
```diff
@@ -7,6 +7,14 @@
...
+        "source": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode": "live-examples/html-examples/embedded-content/source.html",
+            "cssExampleSrc": "live-examples/html-examples/embedded-content/css/source.css",
+            "fileName": "source.html",
+            "title": "HTML Demo: <source>",
+            "type": "tabbed"
         }
     }
 }
```
wbamberg: Files should end with a newline.

live-examples/html-examples/embedded-content/source.html
```diff
@@ -0,0 +1,12 @@
+<video controls
+    width="300"
+    height="200"
+    autoplay="true"
+    loop="true"
```
wbamberg: I think we should omit the `autoplay` and `loop` attributes. I appreciate that you want to demo them, but I think it would be distracting for people who don't want to interact with the example, and are just visiting the page to look something up (we could omit just `autoplay`, but I'd prefer to omit both I think).

live-examples/html-examples/embedded-content/source.html
```diff
@@ -0,0 +1,12 @@
...
+    muted="true">
+    <source src="/media/examples/stream_of_water.webm" type="video/webm">
```
wbamberg: If we break this line and indent so the attributes are aligned, then it will be more readable at the most likely widths:

```
<source src="/media/examples/stream_of_water.webm"
        type="video/webm">
```

live-examples/html-examples/embedded-content/source.html
```diff
@@ -0,0 +1,12 @@
...
+    <source src="/media/examples/stream_of_water.mp4" type="video/mp4">
```
wbamberg: Same comment as above.

live-examples/html-examples/embedded-content/source.html
```diff
@@ -0,0 +1,12 @@
...
+    This browser does not support HTML5 video element.
```
wbamberg: I think this should be:

"This browser does not support the HTML5 video element."

PR #890, opened 2018-05-05
Add html input reset

wbamberg:
Added an example for `<input type="reset">`.

There are a couple of issues here: the grid layout is dreamy in Firefox, but seems to be broken in Chrome. Do you have any idea what I'm doing wrong?

Also, in Chrome, clicking "Submit" reloads the page, which is annoying. I don't know if we can fix this behaviour in the editor, but if not perhaps we'll just have to do without a "Submit' button, which would look a bit weird in this case.


wbamberg:
Thanks @chrisdavidmills , I made those changes (well, mostly - `gap` rather than `grid-gap` and `.controls` rather than `.wrapper`).

Please take another look.

schalkneethling:
Thanks @chrisdavidmills - What an interesting bug, is this widely known?

chrisdavidmills:
@schalkneethling Yeah, looks like it is known. This SO post has some interesting info (it's to do with fieldset being a replaced element, and it's not the only element that suffers):

https://stackoverflow.com/questions/28078681/why-cant-fieldset-be-flex-containers

We fixed this for fieldset back in Fx46:

https://bugzilla.mozilla.org/show_bug.cgi?id=1230207

They tried to fix it in Chromium, but it got reverted due to bustage:

https://bugs.chromium.org/p/chromium/issues/detail?id=262679

PR #891, opened 2018-05-05
Add Input ~ <input type="button">

nnupoor:
Example added for Issue #818


live-examples/html-examples/input/button.html
```diff
@@ -0,0 +1,8 @@
+<fieldset>
+    <legend>Button</legend>
+    <form>
+      <p> These buttons work with access key modifiers too. For Chrome, try `Ctrl + Alt + p`(Mac) / `Alt + p`(Others).</p>
```
schalkneethling: @nnupoor I would recommend we do without these for now. We are going to add JavaScript support to the editor in another round of iteration. You can keep the `accesskey` attribute but, remove this sentence and then the `onclick` JS parts as well.

live-examples/html-examples/input/css/button.css
```diff
@@ -0,0 +1,21 @@
+@font-face {
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
+    url('/media/fonts/FiraSans-Regular.woff2') format('woff2');
+}
+
+legend {
+    background-color: #000;
+    color: #ffffff;
```
schalkneethling: Nit: Code Style: Prefer short form `#ffffff` => `#fff`

live-examples/html-examples/input/css/button.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.btn {
+    font: 1rem 'Fira Sans', sans-serif;
+    padding: 1rem;
+    background: #86d1f0;
+    border: none;
+    border-radius: 4px;
+    margin: 4px;
+    outline: none;
```
schalkneethling: I would not recommend we add this as it hinders accessibility.

live-examples/html-examples/input/button.html
```diff
@@ -0,0 +1,8 @@
...
+      <input class="btn" type="button" value="Click Me" accesskey="p" onclick="alert('I work with Javascript attached to me. But `button` is preferred over me these days. :( ')">
+      <input class="btn" type="button" value="I am disabled" onclick="alert('I won\'t work' )" disabled>
```
schalkneethling: Perhaps add another button showing styling of the different states of a button element? See here for example: https://github.com/mdn/interactive-examples/blob/master/live-examples/html-examples/forms/button.html

nnupoor:
@schalkneethling I'll get these done and raise a PR. Thank you for the suggestions.

nnupoor:
@schalkneethling Fixed the issues. Let me know if I need to do something further. I'll be glad to help. :)

PR #897, opened 2018-05-07
Add html input search

wbamberg:
I've mostly lifted this example from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#Search_form_labels_and_accessibility.


schalkneethling:
The search field and buttons seems a bit small to me.

screen shot 2018-05-09 at 11 27 09

Perhaps add `font-size: 1.2rem;`

screen shot 2018-05-09 at 11 29 21

wbamberg:
Good call @schalkneethling , I've made that change. This wasn't the most exciting example, I guess I'm low on inspiration at the moment :).

PR #912, opened 2018-05-09
fix(tests): fix failing css example test

schalkneethling:


__tests__/puppeteer-css-editor.test.js
```diff
@@ -5,7 +5,7 @@ describe('CSS Editor', () => {
 
     it('sets the example-element\'s style attribute to the correct value', async () => {
         const expectedStyleAttr =
-            'animation: slidein 3s linear 1s infinite running;';
+            'animation: slidein 3s linear 1s infinite normal none running;';
```
wbamberg: Oh, so this is because it resets any properties that weren't explicitly specified?

schalkneethling: Yup, seems to be a new thing in the latest Chromium build that it uses.

schalkneethling: I am going to raise an issue for this though as it does make testing hard, if your expectations break because of changes in the browser engine. Might be that this is not the best way to test it. Perhaps getting computed style and checking for specific properties is more apt.

wbamberg: Or even just not using a shorthand property to text with?

PR #914, opened 2018-05-09
[HTML Examples] Text Content ~ main

a2sheppy:
Need an HTML example for the dt element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

## Contributing


live-examples/html-examples/text-content/main.html
```diff
@@ -0,0 +1,7 @@
+<header>
+    <img class='logo' src='/media/examples/lizard.png'>
```
schalkneethling: Nit: Code Style: Use double quotes in HTML. Also, explicitly close tags such as `img`. I know it is not required in HTML5 but, from a consistency perspective, I encourage that usage. Thanks @a2sheppy

live-examples/html-examples/text-content/main.html
```diff
@@ -0,0 +1,7 @@
...
+    <span>Gecko Goodness</span>
+</header>
+<main>
```
schalkneethling: For the semantic value of this to be correctly identified in IE11 and lower one needs to add `role="main"`. I reckon it would be good to illustrate that here.

live-examples/html-examples/text-content/css/main.css
```diff
@@ -0,0 +1,18 @@
...
+  	padding: 0.25rem 0.75rem 0.25rem 0;
+  	float: left;
+}
+
+main {
+  	margin-top: 0.2em;
+  	padding: 0 1em;
+    border: 1px solid #000;
+    box-shadow: 0 4px 3px #777;
+}
```
schalkneethling: Nit: Code style: Ensure a consistent indentation of 4 spaces.

live-examples/html-examples/text-content/css/main.css
```diff
@@ -0,0 +1,18 @@
+header {
+    height: 3.75rem;
+  	font: bold 2.5rem 'Arial', sans-serif;
+  	border-bottom: 1px solid #000;
+}
+
+header > .logo {
+    height: 3.5rem;
+  	padding: 0.25rem 0.75rem 0.25rem 0;
```
schalkneethling: Nit: Code Style: For 0 based fractional values, omit the leading 0

wbamberg:
@schalkneethling , I fixed the review comments. The layout also seemed to break at narrower widths, so I had a go at fixing it. I also changed the content a bit - I appreciate that this is entirely subjective, but since I was in there anyway, I thought, why not. I also added the `height` property.

PR #920, opened 2018-05-13
Add <area> example

### LINKED ISSUE #758: [HTML Examples] Image and multimedia ~ area
Need an HTML example for the `area` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area

## Contributing

2alin:
Fixes #758 
This one, was full of fun :grin:.
Because of the length taken by the coordinates of a polygon area, I decided to not put an                            `alt` property in the anchors (but I did in the image). If I need to specify that anyway, I could write `alt="go to ..."` and it will take at least 16 more characters of horizontal space, please let me know.

The whole graphic, was made from scratch by me in inkscape (I only took the dinosaur head from an external source), so there won't be any kind of copyright problem :wink:.


wbamberg:
This is really cool! Nice job :)

I think you do need `alt`, because it's mandated. I checked using the W3C validator, and it gave me the expected error:

screen shot 2018-05-16 at 9 05 12 pm

I take your point about horizontal space though. What I do is use hard line breaks to make sure that the markup will be readable when the browser window is 1000px (preferably less, more like 900px). I've done this a lot with the `input` examples, which typically use lots of attributes (e.g. https://interactive-examples.mdn.mozilla.net/pages/tabbed/input-password.html).

Then you get the other problem of course, which is lots of lines in the example. In this case I would probably suggest having fewer hotspots in the image. For example you could have just three - HTML, CSS, JS. Or maybe four, and you could put MDN in the middle, then you could demo `circle` as well as `poly`. I'm sorry to put you to the trouble of reworking the image though :(. 

What do you think about having the links go somewhere, using something like:

```
<area shape="poly" coords="207,241,54,241,72,217,189,217"
      href="https://developer.mozilla.org/en-US/docs/Web/JavaScript" target="_blank">
```

2alin:
And because the example is for `<area>`, I'm trying to cover all three different `shapes`: `circle`, `poly`, and `rect`. But if it's too much, I can make the figure without the recangle at the corner. What do you think is best?

wbamberg:
Actually, thinking some more about this, I think you can keep the image and the example won't be too long. Let's say you split each `area` element across three lines, like:

```
    <area shape="circle" coords="130,136,60"
          href="https://developer.mozilla.org/docs/Web/JavaScript"
          target="_blank" alt="some alt text" />
```

...then I think the example would be about 25 lines long, which is fine.

One thing I missed before: even when you don't need it we are preferring to use `/>` to close th element, so:

```
<area some stuff />
```

not:
```
<area some stuff >
```

2alin:
Done!
Could you review it again? please. Code ended up quite dense, mainly because of coords and href values, but as you said, it's only 24 lines :grin:.

wbamberg:
eek! I wondered what had happened to it! :-)

2alin:
thank to you for reviewing it and all the corrections you pointed out :)

PR #921, opened 2018-05-14
exclusive rare example for empty array

nikhilganji:
Empty array returns true when false is expected.

### DIFF
--- a/live-examples/js-examples/array/array-every.html
+++ b/live-examples/js-examples/array/array-every.html
@@ -7,5 +7,10 @@ var array1 = [1, 30, 39, 29, 10, 13];
 
 console.log(array1.every(isBelowThreshold));
 // expected output: true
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your contribution! I do appreciate it. But I think this aspect of `every()` is more of an edge case - in the interactive example we want to cover the basic use of an API, not all the details. If we do want to have an example, it would be better to have a static example in the "examples" section of the page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every#Examples.

nikhilganji:
Well yeah, thanks for letting me know that. I will better add it to the
examples section.




*Best Regards,*
Nikhil Ganji
(361) 228-6802
ganjinikhil@gmail.com

PR #924, opened 2018-05-15
Add HTML example for <details>

Regaddi:
No github issue created so far.
I added some more styles to the element to make clear that it is a fully customizable element.

### DIFF
--- a/live-examples/html-examples/interactive-elements/css/details.css
+++ b/live-examples/html-examples/interactive-elements/css/details.css
@@ -1,24 +1,13 @@
 details {
-    border: 1px solid #aaaaaa;
+    border: 1px solid #aaa;
     border-radius: 4px;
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/interactive-elements/css/details.css
```diff
@@ -0,0 +1,41 @@
+details {
+    border: 1px solid #aaaaaa;
```
wbamberg: We use short form hex codes when we can, so this should be `#aaa`.

Regaddi:
@wbamberg Thank you very much for the detailed review, I really appreciate your input!

I adjusted styles and content accordingly. Using the `::-webkit-details-marker` was truly insufficient. I removed it to engage simplicity in the example.

PR #926, opened 2018-05-16
Add HTML example for <summary>

Regaddi:
The merge will potentially fail, when #924 is merged before, because I created a new directory for the interactive elements.
I created a slightly simpler example compared to my `<details>` example.
I can surely do a rebase and fix the merge conflicts if #924 is merged before.


Regaddi:
@wbamberg I copied the styles over from #924 and changed the content.

PR #927, opened 2018-05-16
fix(tabbed-editor): issue #860, isolate CSS to output

schalkneethling:
This has been tested in Chrome(and other Webkit/Blink browsers) as well as Firefox. Still need to confirm that this works in Edge as well. The VM is pretty large so, while it download, this is good for review.


schalkneethling:
@wbamberg Thanks for the review. I do agree with you but, at this stage I am not sure who to ping for it, and I fear that this hangs around turning stale while I wait on someone.

When I was asking around for help, I could not find anyone that had experience with ShadowDOM in more specifically with the polyfills.

The code I am using now for ShadowDOM means that browsers that does not support ES6 will not work so, I need to add Babel to compile the code down to a support matrix we target. I have wanted to do this for some time here and on MDN so, no time like the present ;)

I will ask for review, but I am also going to go ahead and merge this so I can work on the next piece. I can always open up pull requests later should someone point out ways that this can be improved upon.

PR #936, opened 2018-05-18
added fantasy font-family for font-family example

ifeature:
Improved existing example - `font-family.html` which relates to [font-family article on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family).

- Added `font-family: fantasy;` option.

### DIFF
--- a/live-examples/css-examples/fonts/font-family.html
+++ b/live-examples/css-examples/fonts/font-family.html
@@ -42,6 +42,14 @@
 </div>
 </section>
 
+<div class="example-choice">
+<pre><code class="language-css">font-family: fantasy;</code></pre>
+<button type="button" class="copy hidden" aria-hidden="true">
+    <span class="visually-hidden">Copy to Clipboard</span>
+</button>
+</div>
+</section>
+
 <div id="output" class="output hidden">
     <section id="default-example">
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your contribution @ifeature ! But we don't really have the vertical space in the CSS examples for more than 6 options, so for cases like `font-family` we have to choose which 6 options best represent the property. I think the 6 we already have are pretty representative.

wbamberg:
BTW if you are interested in contributing interactive examples, we need examples for all the JavaScript [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)  methods, which got missed when we were adding JS examples. I will be filing bugs for them when I get time, but in the meantime, for example,

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt

As a guide to what they look like, I just merged examples for `String.trim()/trimEnd()/trimStart()`: https://github.com/mdn/interactive-examples/pull/922

PR #939, opened 2018-05-19
Add video example.

### LINKED ISSUE #763: [HTML Examples] Image and multimedia ~ video
Need an HTML example for the `video` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

## Contributing

sheeeng:
Fix #763.


wbamberg:
This looks great @sheeeng , except it looks like `type` is not an attribute of `video`: https://html.spec.whatwg.org/multipage/media.html#the-video-element.

Also if you wanted, you could use `muted`, and then you could use the video that has sound (and remove the HTML comment explaining about the 2 videos, to simplify the example a bit.

sheeeng:
Thanks for the correction, @wbamberg. I did not notice the mistake. Example updated.

PR #940, opened 2018-05-19
Add track example.

### LINKED ISSUE #762: [HTML Examples] Image and multimedia ~ track
Need an HTML example for the `track` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track

## Contributing

sheeeng:
Fix #762.


wbamberg:
This is a very nice example. I believe this is going to be the last of the "Image and multimedia" examples, really fantastic work!!

A couple of suggestions:

* as I understand it, `<source>` is mostly used where you have multiple different possible sources for a video. Since here you just have one, why not just use the `src` attribute of `<video>`? It would simplify the markup...

* according to the W3C validator, you need to quote the `href` in the footer:

```
<footer><a href="https://www.youtube.com/watch?v=Z_4TlZCvYL4">Video</a> by Leonard Lee</footer>
```
...because it contains `=`

* the subtitles are very small. Why not add something like this to the CSS:

```
video::cue {
    font-size: 1rem;
}
```
...this would also demo how to style subtitles...

sheeeng:
Thanks for the feedback, @wbamberg. Example updated.

PR #941, opened 2018-05-19
Implemented example for <figcaption> tag

dagolinuxoid:
Take a look. I used few lines of HTML and 22 lines of CSS — a little bit of scrolling won't hurt.

### DIFF
--- a/live-examples/html-examples/text-content/css/figcaption.css
+++ b/live-examples/html-examples/text-content/css/figcaption.css
@@ -1,0 +1,21 @@
+figure {
+    border: thin silver solid;
+    max-height: 240px;
+    padding: 5px;
+    display: flex;
+    flex-direction: column;
+}
+
+img {
+    width: 100%;
+    display: block;
+    max-height: 220px;
+}
+
+figcaption {
+    font: italic smaller sans-serif;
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/text-content/css/figcaption.css
```diff
@@ -0,0 +1,21 @@
...
+    max-height: 220px;
+}
+
+figcaption {
+    font: italic smaller sans-serif;
+    text-align: center;
+    background: #222;
+    color: #fff;
+    padding: 3px 0;
+}
```
wbamberg: Files should end with a newline, here and in the other cases.

live-examples/html-examples/text-content/figcaption.html
```diff
@@ -0,0 +1,4 @@
+<figure>
+    <img src="/media/examples/hamster.jpg" alt="hamster">
```
wbamberg: In these examples we prefer to close tags even when it's not needed:

`hamster`

dagolinuxoid:
@wbamberg, thanks for a thorough review. I made some fixes so now it looks decent... I guess :)

PR #949, opened 2018-05-26
Update math-trunc.html

goodwin64:
Fix: proper output

image


schalkneethling:
Thanks @goodwin64, that is indeed what the expected output should be but, the editor seems to be throwing away the `-`. This is most likely a bug that we need to look into.

schalkneethling:
I have opened an issue for it https://github.com/mdn/interactive-examples/issues/952 - Will update this once the related issue has been resolved.

goodwin64:
@schalkneethling thanks for clarifying!

PR #958, opened 2018-06-03
Add rtc example.

### LINKED ISSUE #783: [HTML Examples] Inline text semantics ~ rtc
Need an HTML example for the `rtc` element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rtc

## Contributing

sheeeng:
Fix #783.

### DIFF
--- a/live-examples/html-examples/inline-text-semantics/rtc.html
+++ b/live-examples/html-examples/inline-text-semantics/rtc.html
@@ -1,6 +1,16 @@
 <ruby>
-    <rb>旧</rb><rt>jiù</rt>
-    <rb>金</rb><rt>jīn</rt>
-    <rb>山</rb><rt>shān</rt>
-    <rtc>San Francisco</rtc>
+    <ruby xml:lang="zh-Hant">
+        <rbc>
+        <rb>大</rb><rp>(</rp><rt>dà</rt><rp>)</rp>
+        <rb>馬</rb><rp>(</rp><rt>mǎ</rt><rp>)</rp>
+        <rb>女</rb><rp>(</rp><rt>nǚ</rt><rp>)</rp>
+        <rb>子</rb><rp>(</rp><rt>zǐ</rt><rp>)</rp>
+        <rb>籃</rb><rp>(</rp><rt>lán</rt><rp>)</rp>
+        <rb>球</rb><rp>(</rp><rt>qiú</rt><rp>)</rp>
+        <rb>隊</rb><rp>(</rp><rt>duì</rt><rp>)</rp>
+        </rbc>
+    </ruby>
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/inline-text-semantics/rtc.html
```diff
@@ -0,0 +1,6 @@
+<ruby>
+    <rb>旧</rb><rt>jiù</rt>
+    <rb>金</rb><rt>jīn</rt>
+    <rb>山</rb><rt>shān</rt>
+    <rtc>San Francisco</rtc>
+</ruby>
```
wbamberg: I'm not very familiar with the HTML ruby elements, but I'm not sure this is what we want. As I understand it the idea of `<rbc>`/`<rtc>` is that you can apply Ruby annotations to different segments of the text. In this case I think the whole text "旧 金 山" means "San Francisco", so we want the "San Francisco" to run all along the text, which we could achieve by adding it to its own `<rbc>` container:

```
<ruby>
    <rbc>
        <rb>旧</rb><rt>jiù</rt>
        <rb>金</rb><rt>jīn</rt>
        <rb>山</rb><rt>shān</rt>
    </rbc>
    <rtc><rt>San Francisco</rt></rtc>
</ruby>
```

..which gives us:

screen shot 2018-06-05 at 12 07 01 pm

Like I say though I'm not very familiar with this (and our docs are unfortunately not very helpful), so I'm very happy to learn if this isn't correct.

The best article I found on this was by @huijing : https://www.chenhuijing.com/blog/html-ruby/, and it includes a neat example of `<rtc>`:

```
<ruby>
  <ruby xml:lang="zh">
    <rbc>
      <rb>大</rb><rp>(</rp><rt>dà</rt><rp>)</rp>
      <rb>马</rb><rp>(</rp><rt>mǎ</rt><rp>)</rp>
      <rb>女</rb><rp>(</rp><rt>nǚ</rt><rp>)</rp>
      <rb>子</rb><rp>(</rp><rt>zǐ</rt><rp>)</rp>
      <rb>篮</rb><rp>(</rp><rt>lán</rt><rp>)</rp>
      <rb>球</rb><rp>(</rp><rt>qiú</rt><rp>)</rp>
      <rb>队</rb><rp>(</rp><rt>duì</rt><rp>)</rp>
    </rbc>
  </ruby>
  <rtc xml:lang="en" style="ruby-position: under;">
  <rp>(</rp><rt>Malaysia Women's Basketball Team</rt><rp>)</rp>
  </rtc>
</ruby>
```

sheeeng:
Thanks @wbamberg for correcting it to `Hant`. Copy paste error once again on my side.

PR #960, opened 2018-06-04
Fix console util to support negative zero

### LINKED ISSUE #952: Possible bug with negative numbers in console.log
In PR https://github.com/mdn/interactive-examples/pull/949 @goodwin64 pointed out that the last `console.log` statement actually output `-0` in the browsers native console. The JS editor output `0` though so, there is a bug somewhere in the output formatting that discards the `-`

kenrick95:
Fixes #952 

The bug is caused by using [`String(value)` function](https://www.ecma-international.org/ecma-262/6.0/#sec-string-constructor-string-value), which per the spec calls [`ToString(value)`](https://www.ecma-international.org/ecma-262/6.0/#sec-tostring). _ToString(value)_ for number as per defined [at the spec](https://www.ecma-international.org/ecma-262/6.0/#sec-tostring-applied-to-the-number-type) states that

> Number m to String format as follows:
> 2.  If m is +0 or −0, return the String "0".

So, this PR adds the case where if we meet negative zero at `formatOutput`, then return `-0`, else call `String(input)`. Also, more tests are added.

### DIFF
--- a/__tests__/console-utils.test.js
+++ b/__tests__/console-utils.test.js
@@ -2,6 +2,50 @@ const consoleUtils = require('../js/editor-libs/console-utils');
 
 describe('console utils', () => {
     describe('formatOutput', () => {
+        test('String', function() {
+            expect(consoleUtils.formatOutput('lorem ipsum')).toBe(
+                '"lorem ipsum"'
+            );
+        });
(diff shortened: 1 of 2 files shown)


goodwin64:
👍

PR #961, opened 2018-06-05
fix(tabbed-editor): Apply output class to output container

schalkneethling:
Fix https://github.com/mdn/interactive-examples/issues/881 

Tested in Firefox, Firefox Nightly and Chrome


schalkneethling:
Yup, the package-lock was intentional as I had to add a new dependency `"jest-environment-node": "23.1.0"` - Thanks for the review @wbamberg

PR #962, opened 2018-06-07
Issue#693 add sizing util classes tabbed editor

schalkneethling:
@wbamberg Ok so, turns out it was a lot more involved than adding a couple of CSS classes 😄  I separated the work into two commits for easier review. Commit one is the main front end changes, and commit two contains the build and meta changes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,8 @@ module.exports = {
     env: {
         browser: true,
         es6: true,
-        jest: true
+        jest: true,
+        node: true
     },
     extends: 'eslint:recommended',
     parserOptions: {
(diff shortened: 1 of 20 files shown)


wbamberg:
Thanks Schalk! This seems to work great, except it looks strange if I omit the `"height"` property:

screen shot 2018-06-07 at 1 40 58 pm

If it is mandatory (which I think it probably should be) we should document that it is and throw an error if it is omitted.

I guess there will be a corresponding `tabbed-taller` etc argument to the KumaScript macro, as we did for the JS ones? Or will the existing `"taller"` and `"shorter"` (and nothing for "standard") arguments be used?

-> https://github.com/mdn/kumascript/blob/master/macros/EmbedInteractiveExample.ejs#L8 (I seem to remember "shorter" got implemented, but it's not documented in the macro).

schalkneethling:
Yup, I agree that we should throw an error at build time. I will add that to this PR. In terms of the classes for the macro, those landed here:
https://github.com/mozilla/kuma/pull/4842#pullrequestreview-126878628

schalkneethling:
This one has been updated.

PR #967, opened 2018-06-08
fix(performance): send loadEventEnd to parent via postMessage

schalkneethling:
@tkadlec Here is the other side of the coin 😄 

*NOTE:* This PR should only be merged after https://github.com/mozilla/kuma/pull/4846 has been merged and pushed to production.

### DIFF
--- a/js/editor.js
+++ b/js/editor.js
@@ -1,6 +1,7 @@
 (function() {
     'use strict';
 
+    var mceEvents = require('./editor-libs/events.js');
     var shadowOutput = require('./editor-libs/shadow-output');
     var templateUtils = require('./editor-libs/template-utils');
     var tabby = require('./editor-libs/tabby');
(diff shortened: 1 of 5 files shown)


jwhitlock:
https://github.com/mozilla/kuma/pull/4846 is deployed to production.

js/editor.js
```diff
@@ -100,4 +101,22 @@
     document.addEventListener('WebComponentsReady', function() {
         render(templateUtils.getTemplateOutput());
     });
+
+    /* Ensure that performance is supported before
+       gathering the performance metric */
+    if (performance !== undefined) {
+        document.addEventListener('readystatechange', function(event) {
```
jwhitlock: Sorry for the message on the old PR. 

Would it make more sense to listen for the [load event](https://developer.mozilla.org/en-US/docs/Web/Events/load) rather than the ``readystatechange`` event? This may avoid the 0.1 second wait for ``performance.timing.loadEventEnd`` to be populated.

schalkneethling: Definitely worth trying, thanks for the suggestion @jwhitlock

Issue #976, opened 2018-06-19
Inline perf.js for performance marks and measure


schalkneethling:
We now have a `perf.js` script that set’s certain marks and measure during document load time. The best way to utilise this is in the `head` of the document so it is non blocking and can start capturing measurements as soon as possible.

jwhitlock:
I _think_ this is closed, based on #977 being deployed.

PR #977, opened 2018-06-19
fix(perf): adds and inlines perf.js ~ Fix issue #976

schalkneethling:
@tkadlec r?


js/editor-libs/perf.js
```diff
@@ -0,0 +1,39 @@
...
+     * process, and exposes the following measurements:
+     * 1. Time to interactive
+     * 2. Time to DOM complete
+     */
+    if (performance.mark && performance.measure) {
+        performance.mark('interactive-editor-loading');
+
+        document.addEventListener('readystatechange', function(event) {
+            switch (event.target.readyState) {
+            case 'interactive':
```
tkadlec: Both of these seem to be working really well!

But since they come from the iframe instead of the parent window, they won't be accessible in the parent's `performance` object. To see these, and track them, we need to do one of two things.

We could post their durations (`performance.getEntriesByName('ie-time-to-interactive')[0].duration`, for example) back to the parent frame and then beacon them back to GA.

Our other option is to use the `readystatechange` event to post a message back to the parent, and then have the parent set a `mark` and beacon it to GA. You'll lose a little bit of accuracy because of the message posting, but this does give us the benefit of being able to track this metric in SpeedCurve, WebPageTest, etc. I think the trade-off is small enough that it's likely worthwhile sense.

schalkneethling: That was indeed my concern. As I believe the intent is to see these in Speedcurve as apposed to GA, I will opt for the second of two options. Thanks Tim

schalkneethling:
@tkadlec Ok updated. I am now simply posting to Kuma with the labels for the various timing events. Setting the marks and measures and beaconing to GA will be handled on the Kuma side.

Let me know if this looks good to you. Thanks!

PR #979, opened 2018-06-19
Add example for html <figure> element

wbamberg:
This mostly clones the existing example for `<figcaption>` but uses a different image, just for a bit of variety.

I thought it best to reuse an existing image though, so I used the frog one, but then I needed to rename that, so also needed to rename the other image used in the `<picture>` example (they were called "picture-narrow" and "picture-wide", which doesn't make sense in other contexts).


live-examples/html-examples/embedded-content/picture.html
```diff
@@ -1,9 +1,9 @@
...
-    <source srcset="/media/examples/picture-wide.png"
+    <source srcset="/media/examples/frog.png"
             media="(min-width: 1000px)">
-    <img src="/media/examples/picture-narrow.png">
+    <img src="/media/examples/fish.png">
 </picture>
 
-<footer>Images by <a href=https://unsplash.com/@davidclode>David Clode</a> on Unsplash</footer>
\ No newline at end of file
+<footer>Images by <a href=https://unsplash.com/@davidclode>David Clode</a> on Unsplash</footer>
```
schalkneethling: Nit: Code style: Wrap the value of the `href` in double quotes.

live-examples/html-examples/text-content/css/figure.css
```diff
@@ -0,0 +1,15 @@
+figure {
+    border: thin #C0C0C0 solid;
```
schalkneethling: Nit: Code style: lower case letters in colour values;

live-examples/html-examples/text-content/css/figure.css
```diff
@@ -0,0 +1,15 @@
...
+    display: flex;
+    flex-flow: column;
+    margin: 0;
+    padding: 5px;
+}
+
+figcaption {
+    background: #222;
```
schalkneethling: Nit: Code style: Because you are only specifying a colour, rather be explicit and use the `background-color` property.

PR #980, opened 2018-06-20
Add example for html <div> element

wbamberg:


wbamberg:
Thanks @chrisdavidmills , this makes sense. There are a couple of difficulties which I have tried to wrestle with:

* generally vertical space is difficult when there are so many elements. I would like to have a little more space, but I think this is OK.

* although it's not required, we like to attribute Unsplash images, but having that in the output would look weird with the Adams attribution as well. So I've gone for attribution in the source instead.

chrisdavidmills:
Looks much better; r+

PR #1000, opened 2018-06-27
Correct "expected output" mistake

Arkangus:


live-examples/js-examples/array/array-reverse.html
```diff
@@ -3,7 +3,7 @@
 var reversed = array1.reverse(); 
 
 console.log(array1);
-// expected output: Array ['three', 'two', 'one']
+// expected output: Array ['one', 'two', 'three']
 
 console.log(reversed);
 // expected output: Array ['three', 'two', 'one']
```
schalkneethling: @Arkangus This is actually a little trickier and perhaps we should rewrite the example a bit. The `reverse` function is actually destructive i.e. it changes in place so, effects the original array as well. So, the output that it shown, is what is printed but, might be unexpected.

I therefore propose we change the example to the following:

```
var array1 = ['one', 'two', 'three'];
console.log('array1: ', array1);
// expected output: Array ['one', 'two', 'three']

var reversed = array1.reverse(); 
console.log('reversed: ', reversed);
// expected output: Array ['three', 'two', 'one']

/* Because `reverse`'s changes are done in place it
   also changes the original array, as can be seen here */
console.log('array1: ', array1);
// expected output: Array ['three', 'two', 'one']
```

Thoughts @wbamberg ?

PR #1001, opened 2018-06-27
fix(performance): send post to kuma for custom ie-load-event-end metric

schalkneethling:
@tkadlec This will post a message to Kuma to set a custom mark at the same time as we send `performance.timing.loadEventEnd` for GA beaconing.

On the Kuma side I will then also set a new custom mark in the head of the pages similar to https://github.com/mdn/interactive-examples/blob/master/js/editor-libs/perf.js#L18

I will then also set a new measure which will use the mark from the head in Kuma as `startMark` and then the mark we are creating here as `endMark`. That should then give us the duration equal to `ie-load-complete`

Let me know your thoughts. Thanks!

### DIFF
--- a/js/editable-css.js
+++ b/js/editable-css.js
@@ -103,6 +103,8 @@
                         'CSS editor load time',
                         performance.timing.loadEventEnd
                     );
+                    // Posts mark to set on the Kuma side and used in measure
+                    mceUtils.postToKuma({ markName: 'css-ie-load-event-end' });
                 }, 100);
             }
         });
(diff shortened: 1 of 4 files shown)


tkadlec:
> I will then also set a new measure which will use the mark from the head in Kuma as startMark and then the mark we are creating here as endMark. That should then give us the duration equal to ie-load-complete

I may not be following, but why do we need a custom mark in the head of Kuma for startMark? I think if we use `navigationStart` as our `startMark` and this mark as the `endMark` we should have an accurate picture of what's going on.

schalkneethling:
@tkadlec As I posted that, I was thinking exactly that :) Other than that, does this look good to you?

PR #1004, opened 2018-06-28
fix(address): add name to the provided address

ro-ka:
The description for the [`<address>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address) element says:

> … It should include the name of the person, people, or organization to which the contact information refers.

This PR adds a name to make it clear that one should be added.

### DIFF
--- a/live-examples/html-examples/content-sectioning/address.html
+++ b/live-examples/html-examples/content-sectioning/address.html
@@ -1,4 +1,5 @@
 <address>
+  James Rockford
   2354 Pacific Coast Highway<br>
   California<br/>
   USA<br/>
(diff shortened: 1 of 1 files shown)


live-examples/html-examples/content-sectioning/address.html
```diff
@@ -1,4 +1,5 @@
 <address>
+  James Rockford
```
wbamberg: Should have `<br/>` after it perhaps?

ro-ka:
Argh, sorry.  🤦‍♂️   Added the line breaks!

PR #1008, opened 2018-06-29
Add color-adjust property example

a2sheppy:
Adds color-adjust interactive example.

### DIFF
--- a/live-examples/css-examples/color/color-adjust.css
+++ b/live-examples/css-examples/color/color-adjust.css
@@ -1,0 +1,11 @@
+#example-element {
+    padding: 1em;
+    font-size: 1.5em;
+    background-color: black;
+    background-image: linear-gradient(rgba(0, 0, 180, 0.5), rgba(70, 140, 220, 0.5));
+    color: #333;
+    text-align: center;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/color/color-adjust.css
```diff
@@ -0,0 +1,11 @@
+#example-element {
+    padding: 1em;
+    font-size: 1.5em;
```
schalkneethling: Super Nit: Code style: Prefer `rem` over `em` for font sizes.

live-examples/css-examples/color/color-adjust.css
```diff
@@ -0,0 +1,11 @@
...
+    background-color: black;
+    background-image: linear-gradient(rgba(0, 0, 180, 0.5), rgba(70, 140, 220, 0.5));
```
schalkneethling: Nit: Code style: Omit `0` from fractional values

a2sheppy:
Remove the leading "0" from fractional numbers? That's... strange. Leaving out the leading 0 makes the number harder and violates every style guide I've ever seen.

a2sheppy:
It's still weird; I've honestly never seen such a thing. I'll make the change, but I think it badly damages legibility of the code.

chrisdavidmills:
I've seen both in fairly equal measure, FWIW. I think I prefer "with leading 0s", for similar reasons to those sheppy brings up, but I can live with it.

schalkneethling:
@a2sheppy @chrisdavidmills Yeah, this is one of those points where there is often contention ;) As we have followed this guideline for the rest of this codebase, and have also been following it on Kuma since before my time, I felt that we should stick with it.

PR #1009, opened 2018-06-30
Issue#946 open links in new tab

stephanmax:
I added two helper functions that are called inside the editor's `render` method.

`openLinksInNewTab` prevents the default click event on external links (all anchors with an `href` attribute that starts with `http`) and opens them in a new tab.

While I was at it, I stumbled upon the issue that relative links (like `<a href="#someanchor">`) also didn't work. I didn't find a ticket for that so I thought it would make sense to fix in the scope of this ticket.

`scrollToAnchors` prevents the default click event on relative links and scrolls the targeted anchor into view.

I hope that the code structure adheres to your quality standards.

### DIFF
--- a/js/editor.js
+++ b/js/editor.js
@@ -55,35 +55,6 @@
     }
 
     /**
-     * Interrupts the default click event on external links inside
-     * the shadow dom and opens them in a new tab instead
-     * @param {Array} externalLinks - all external links inside the shadow dom
-     */
-    function openLinksInNewTab(externalLinks) {
-        externalLinks.forEach(function(externalLink) {
-            externalLink.addEventListener('click', function(event) {
-                event.preventDefault();
-                window.open(externalLink.href);
-            });
-        });
-    }
-
-    /**
-     * Interrupts the default click event on relative links inside
-     * the shadow dom and scrolls to the targeted anchor
-     * @param {Object} shadow - the shadow dom root
-     * @param {Array} relativeLinks - all relative links inside the shadow dom
-     */
-    function scrollToAnchors(shadow, relativeLinks) {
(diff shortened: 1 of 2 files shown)


js/editor.js
```diff
@@ -54,6 +54,35 @@
...
+     * @param {Array} relativeLinks - all relative links inside the shadow dom
+     */
+    function scrollToAnchors(shadow, relativeLinks) {
+        relativeLinks.forEach(function(relativeLink) {
+            relativeLink.addEventListener('click', function(event) {
+                event.preventDefault();
+                shadow.querySelector(relativeLink.hash).scrollIntoView();
+            });
+        });
+    }
```
schalkneethling: I would suggest adding these two functions to our utilities module:
https://github.com/stephanmax/interactive-examples/blob/e669e9f9f637684d371f97ac77ff7dbe27bf7145/js/editor-libs/mce-utils.js

js/editor.js
```diff
@@ -80,6 +109,8 @@
 
         shadow.appendChild(document.importNode(content, true));
         setOutputHeight(shadow.querySelector('div'));
+        openLinksInNewTab(shadow.querySelectorAll('a[href^="http"]'));
+        scrollToAnchors(shadow, shadow.querySelectorAll('a[href^="#"]'));
```
schalkneethling: You will then call these with:

```
mceUtils.openLinksInNewTab(shadow.querySelectorAll('a[href^="http"]'));
mceUtils.scrollToAnchors(shadow, shadow.querySelectorAll('a[href^="#"]'));
```

stephanmax:
That makes sense, thanks for the fast review, @schalkneethling. I updated the PR accordingly.

Issue #887, opened 2018-05-03
@font-face doesn't work with Shadow DOM?


wbamberg:
Related to https://github.com/mdn/interactive-examples/issues/881: the `input` examples generally use `.output` to set the font to one that we loaded using `@font-face`. But changing this seems to indicate that the custom font specified using `@font-face` isn't being loaded any more.

For example, in https://interactive-examples.mdn.mozilla.net/pages/tabbed/input-file.html, changing 

```
.output {
    font: 1rem 'Fira Sans', sans-serif;
}
```

to

```
legend, label {
    font: 1rem 'Fira Sans', sans-serif;
}
```
... you can see that the rules are having an effect, but `'Fira Sans'` is not being used.

This is in both Firefox(with the shadow DOM pref on) and Chrome.

wbamberg:
Weirdly this seems to work in production: https://interactive-examples.mdn.mozilla.net/pages/tabbed/input-email.html, just not in the local testing server.

wbamberg:
It looks like this is still not loading the custom fonts. In this example "He-Man" should use [molot](https://www.fontsquirrel.com/fonts/Molot), and "Skeletor" should use [rapscallion](https://www.dafont.com/rapscallion.font).

<img width="442" alt="screen shot 2018-07-12 at 2 13 39 pm" src="https://user-images.githubusercontent.com/432915/42659671-d5577bdc-85dd-11e8-81c8-6e2ce8addef6.png">

schalkneethling:
I tested this on production at:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

Tested in Chrome, Fx Release, Developer, Nightly, as well as Safari. They all display as below:

<img width="1056" alt="screen shot 2018-07-13 at 11 51 25" src="https://user-images.githubusercontent.com/10350960/42685520-74177e2e-8693-11e8-8b0c-a3573b8b82f2.png">

Perhaps cache because of the experiment going on at the moment with the longer lived cache? Try accessing and see if that busts the cache:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption?v=2

wbamberg:
Oh yes! Works great. Must have been the cache :).

drochgenius:
Hi,

I just found out the `@font-face` doesnt load the fonts if it's within a Shadow DOM. Looking at this thread, and in the Shadow DOM v1 spec I couldn't find any reference to this limitation.

[...]

PR #1015, opened 2018-07-05
fix(tabbed): load fonts as part of the editor css as @font-face does not work inside shadow dom

schalkneethling:
Turns `@font-face` and a few other such declarations does not work in ShadowDOM. So, the solution is to load up all the fonts we use in the examples as part of the editor bundle.

Fix issue #887


schalkneethling:
Thanks for the doc update @wbamberg completely slipped my mind.

PR #1016, opened 2018-07-05
adding rb element example

chrisdavidmills:
Noticed we were missing a page on `<rb>` while looking at our Ruby docs. This is part if it.

Related to https://github.com/mdn/sprints/issues/247

### DIFF
--- a/live-examples/html-examples/inline-text-semantics/rb.html
+++ b/live-examples/html-examples/inline-text-semantics/rb.html
@@ -1,5 +1,5 @@
 <ruby>
-  <rb>上<rb>手
-  <rt>じよう<rt>ず
-  <rtc><rt>jou<rt>zu
+    <rb>上</rb><rb>手</rb>
+    <rt>じよう</rt><rt>ず</rt>
+    <rtc><rt>jou</rt><rt>zu</rt></rtc>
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/inline-text-semantics/rb.html
```diff
@@ -0,0 +1,5 @@
+<ruby>
+  <rb>上<rb>手
+  <rt>じよう<rt>ず
+  <rtc><rt>jou<rt>zu
```
wbamberg: Nit: we use 4-space indent.

wbamberg: I'm not sure about all this omitting of closing tags. It is correct I think, but it makes Code Mirror sad:

screen shot 2018-07-11 at 5 20 36 pm

@schalkneethling , what do you think?

schalkneethling: Agree, please close all void, and non-void tags. Thanks!

chrisdavidmills:
OK, example updated.

I omitted those closing tags because that generally seems to be how people write ruby examples. But hey, it is probably easier to see what's going on if you include them.

PR #1025, opened 2018-07-09
Change <select> examples to use “placeholder label option” pattern

tigt:
https://html.spec.whatwg.org/multipage/form-elements.html#placeholder-label-option

### DIFF
--- a/live-examples/html-examples/forms/option.html
+++ b/live-examples/html-examples/forms/option.html
@@ -1,13 +1,13 @@
 <label for="pet-select">Choose a pet:</label>
 
 <select id="pet-select">
+    <option value="">--Please choose an option--</option>
     <option value="dog">Dog</option>
     <option value="cat">Cat</option>
     <option value="hamster">Hamster</option>
     <option value="parrot">Parrot</option>
     <option value="spider">Spider</option>
     <option value="goldfish">Goldfish</option>
-    <option value="none" selected>--Please choose an option--</option>
 </select>
 
 <footer>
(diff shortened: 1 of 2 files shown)


live-examples/html-examples/forms/option.html
```diff
@@ -1,13 +1,13 @@
 <label for="pet-select">Choose a pet:</label>
 
 <select id="pet-select">
+    <option value="">--Please choose an option--</option>
```
schalkneethling: Thank you for your pull request @tigt but, the specification for this seems to be related to `select` with a `required` attribute:

"If a select element has a required attribute specified,...."

Neither of these do, and because the "--Please choose an option--" `option` does have the `selected` attribute it will be visible by default to the user. Not sure if there is an a11y benefit to having it first in the options list?

If an assistive technology would read all the elements in order, then I can see there being a benefit to reading "--Please choose an option--" first. Although, then I wonder if having the dashes is a god idea?

schalkneethling: Just tested this with VoiceOver on MacOS. Whether the "--Please choose an option--" option is first, or last(with the `selected` attribute) it is read first by VoiceOver. The dashes are also not read.

@wbamberg @chrisdavidmills Thoughts on this change?

wbamberg: I think I agree with you @schalkneethling , that this change should not be needed. But I'm not particularly expert on this matter.

tigt:
It's not necessary, I just figured it was slightly less code and was more consistent with expectations if one *does* want a `required` attribute on there. Putting the "placeholder option" on top is a long-standing tradition from the early days of Windows forms, if not earlier.

PR #1027, opened 2018-07-11
Remove ending `/` that is not necessary

enguerran:
W3C and the examples on the bottom of the page https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Alternative_text tell us the ending `/` is not necessary anymore.
It is confusing to have it on the top of the page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img

### DIFF
--- a/live-examples/html-examples/image-and-multimedia/img.html
+++ b/live-examples/html-examples/image-and-multimedia/img.html
@@ -1,4 +1,4 @@
 <img src="/media/examples/frog.png"
-     alt="Frog"/>
+     alt="Frog">
 
 <footer><a href=https://unsplash.com/photos/2slBHG3HtdA>Image</a> by David Clode on Unsplash</footer>


wbamberg:
Thanks for your contribution!

We don't unfortunately have written coding standards for HTML yet :( but we've intentionally been using `/>` even where it's not needed in these examples.

Personally I'm less bothered about what a coding standard says (unless it is clearly absurd :), as long as there is one and we try to follow it consistently.

cc @schalkneethling in case he has something to add, but otherwise we should close this PR.

pieqq:
Hi @wbamberg,

I'm not sure I understand your rationale here.

The current sample code for the [img element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) is:

```html
<img src="/media/examples/frog.png"
     alt="Frog"/>
```

Note the trailing slash `/` in that example. This is useless in HTML5 (it is only useful in XHTML Strict, I believe). Moreover, as mentioned by @enguerran, none of the examples at the bottom of the same page have the trailing slash... it's very confusing for a newcomer.

This is why I think (and @enguerran too, I guess) the sample at the top of the page should match the ones at the bottom. What do you think?

enguerran:
I opened this PR to add homogeneity to the page to avoid confusion.

As I can see in https://www.w3.org/TR/html5/syntax.html#start-tags
> Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single U+002F SOLIDUS character (/). This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

`` is a void element and the `` is interpreted like ``. We used to need the ending `\` in previous HTML standards, but that is not true anymore. Newcomers could be confused and that is why I suggest to update the interactive example at the top of the page.

schalkneethling:
Hey All,

I agree that at the moment this can get confusing. We are standardising our rules around HTML, CSS and JavaScript. This is going to take some time to filter down the codebase, and onto the MDN pages.

As you can imagine, some of this content has been around for some time. I do feel that in this case, we need to make a call. Either we accept this PR as it is in conflict with the rest of the examples on the page or, we need to update the page.

@wbamberg In order to move us forward, I would suggest latter, but welcome your input.

PR #1029, opened 2018-07-11
Add example for <link>

stephanmax:
This addresses issue #737 with an example loading an external stylesheet via `<link>`.

### DIFF
--- a/live-examples/html-examples/document-metadata/meta.json
+++ b/live-examples/html-examples/document-metadata/meta.json
@@ -1,5 +1,16 @@
 {
     "pages": {
+        "link": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/document-metadata/link.html",
+            "cssExampleSrc":
+                "live-examples/html-examples/document-metadata/css/link.css",
+            "fileName": "link.html",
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/document-metadata/meta.json
```diff
@@ -1,5 +1,16 @@
...
+        "link": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/document-metadata/link.html",
+            "cssExampleSrc":
+                "live-examples/html-examples/document-metadata/css/link.css",
+            "fileName": "link.html",
+            "title": "HTML Demo: <link>",
+            "type": "tabbed",
+            "height": "tabbed-standard"
```
wbamberg: We should use "tabbed-shorter" here, as this is a short example.

media/examples/external.css
```diff
@@ -0,0 +1,3 @@
+p {
```
wbamberg: I think it would be helpful to give this CSS file a more explicit name, like "link-element-example.css" or something.

Even consider including a comment in the file, too, like:
```
/*
This CSS file is used in the example for the <link> element.
*/
```

...that way someone stumbling across it will understand what it's for :)

Issue #999, opened 2018-06-26
Finish HTML element examples


wbamberg:
There are still 7 HTML elements missing an example:

```
<h1>-<h6>
<link>
<style>
<a>
<bdi>
<bdo>
<figure>
```

## Acceptance criteria

All HTML elements have interactive examples, except those we have explicitly marked as not suitable for examples.

wbamberg:
Sounds great @stephanmax . It would probably be good to have separate issues for each example, so more than one person can contribute examples. There are already issues for four of these examples, some of which have been started:

* [`<bdo>`](https://github.com/mdn/interactive-examples/issues/766) - this is currently unassigned, although @Elchi3 [started a PR for it](https://github.com/mdn/interactive-examples/pull/880), but it needs rework before it can be merged.

* [`<bdi>`](https://github.com/mdn/interactive-examples/issues/765) - same story as for `<bdo>`.

* [`<link>`](https://github.com/mdn/interactive-examples/issues/737) - currently unassigned, some discussion in the bug but no PR.

* [`<h1>–<h6>`](https://github.com/mdn/interactive-examples/pull/911) - currently unassigned, a PR filed with some outstanding review comments.

I think it would be fine to take any of the above, and either reuse any existing work or start again from scratch.

For the others:

* I just checked, and `<figure>` has actually been done.
* `<style>` has also been done, but has a bug that's preventing us adding it to the page. I just filed https://github.com/mdn/interactive-examples/issues/1013 for that.
* I just filed [an issue for `<a>`](https://github.com/mdn/interactive-examples/issues/1014), which hasn't been started at all. I think we were waiting for https://github.com/mdn/interactive-examples/issues/946 to be fixed, so thanks for fixing that :).

chrisdavidmills:
`<bdi>` and `<bdo>` examples done: https://github.com/mdn/interactive-examples/pull/1041

chrisdavidmills:
`<link>` example also done: https://github.com/mdn/interactive-examples/pull/1029

stephanmax:
Cool, that only leaves `<h*>`, I think. I know there is a PR and I wasn't sure what the usual process is for those scenarios, but I am happy to jump in and finish that example.

chrisdavidmills:
@stephanmax if you'd be willing to jump in and finish off the h1-h6 example, that would be much appreciated! Feel free to put me as a reviewer once you're done. Sheppy was working on it, but he has been pulled on to other things now.

[...]

PR #1041, opened 2018-07-24
bdi and bdo interactive examples

chrisdavidmills:
As part of the final set of HTML interactive examples (https://github.com/mdn/interactive-examples/issues/999), I've written some examples for `<bdi>` and `<bdo>`.

Note that Florian already attemped these in https://github.com/mdn/interactive-examples/pull/880. I wasn't sure if there was a good way for me to add onto his PR, so I've just created a new one.

### DIFF
--- a/live-examples/html-examples/inline-text-semantics/bdi.html
+++ b/live-examples/html-examples/inline-text-semantics/bdi.html
@@ -1,0 +1,8 @@
+<h1>World wrestling championships</h1>
+
+<ul>
+   <li><bdi>الرجل القوي إيان</bdi>: 1st place</li>
+   <li><bdi>Evil Steven</bdi>: 2nd place</li>
+   <li><bdi>François fatale</bdi>: 3rd place</li>
+   <li>تیز سمی: 4th place (algorithm confused without BDI!)</li>
+</ul>
--- a/live-examples/html-examples/inline-text-semantics/bdi.html
+++ b/live-examples/html-examples/inline-text-semantics/bdi.html
@@ -1,8 +1,9 @@
 <h1>World wrestling championships</h1>
 
 <ul>
-   <li><bdi>الرجل القوي إيان</bdi>: 1st place</li>
-   <li><bdi>Evil Steven</bdi>: 2nd place</li>
-   <li><bdi>François fatale</bdi>: 3rd place</li>
-   <li>تیز سمی: 4th place (algorithm confused without BDI!)</li>
+   <li><bdi class="name">Evil Steven</bdi>: 1st place</li>
+   <li><bdi class="name">François fatale</bdi>: 2nd place</li>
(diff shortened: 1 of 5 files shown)


live-examples/html-examples/inline-text-semantics/bdi.html
```diff
@@ -0,0 +1,8 @@
+<h1>World wrestling championships</h1>
+
+<ul>
+   <li><bdi>الرجل القوي إيان</bdi>: 1st place</li>
+   <li><bdi>Evil Steven</bdi>: 2nd place</li>
+   <li><bdi>François fatale</bdi>: 3rd place</li>
+   <li>تیز سمی: 4th place (algorithm confused without BDI!)</li>
```
wbamberg: I think this is a bit confusing, because the order in the source is the same as the order in the output, so it's easy to think "so what".

What do you think about enclosing the name in a span, so you can really see the difference? e.g.

```
<h1>World wrestling championships</h1>

<ul>
   <li><span class='name'>Evil Steven</span>: 1st place</li>
   <li><span class='name'>François fatale</span>: 2nd place</li>
   <li><span class='name'>تیز سمی</span>: 3rd place</li>
   <li><bdi class='name'>الرجل القوي إيان</bdi>: 4th place</li>
   <li><span class='name' dir='auto' >تیز سمی</span>: 5th place</li>
</ul>
```

(this is probably closer to actual usage, since AFAIK this usually comes up when you want to add a string dynamically and don't know its directionality ahead of time, and in that case you might want a way to target the elements.)

(I've also added a version using `dir='auto'`, and changed the ordering a bit to show the error case before the fixed case. Not sure these are better, that's just how I would have done this.)

To further emphasise it you could add something like:

```
.name {
    color: #f00;
}
```

...in the CSS:

screen shot 2018-07-24 at 3 08 10 pm

live-examples/html-examples/inline-text-semantics/bdo.html
```diff
@@ -0,0 +1,7 @@
+<h1>Famous seaside songs</h1>
+
+<p>The English song "Oh I do like to be beside the seaside"</p>
+
+<p>Looks like this in Hebrew: <span dir="rtl">אה, אני אוהב להיות ליד חוף הים</span></p>
+
+<p>In the computer's memory, this is stored as <bdo>אה, אני אוהב להיות ליד חוף הים</bdo></p>
```
wbamberg: According to https://www.w3.org/International/articles/inline-bidi-markup/#override, `dir` is mandatory for `<bdo>`, and the example accompanying that article uses it: https://www.w3.org/International/articles/inline-bidi-markup/index-data/bdo.html.

(Our docs should say this, too.)

chrisdavidmills:
I agree with your suggestions on this, so have made the necessary changes. Thanks!

PR #1042, opened 2018-07-24
adding accesskey global attribute interactive example

chrisdavidmills:
As per https://github.com/mdn/sprints/issues/292, I have started adding interactive examples for HTML global attributes. This one covers accesskey.

Although there is one problem with this example — when I tested it locally, it worked fine in Firefox, Chrome, Safari, and Opera. 

In the interactive example version, it doesn't work in Firefox only. Very strange.

### DIFF
--- a/live-examples/html-examples/global-attributes/css/accesskey.css
+++ b/live-examples/html-examples/global-attributes/css/accesskey.css
@@ -1,0 +1,4 @@
+.output {
+    font: 1rem 'Fira Sans', sans-serif;
+    letter-spacing: 1px;
+}
--- a/live-examples/html-examples/global-attributes/css/accesskey.css
+++ b/live-examples/html-examples/global-attributes/css/accesskey.css
@@ -1,3 +1,11 @@
+@font-face {
+    font-family: 'Fira Sans';
+    src: local('FiraSans-Regular'),
+         url('/media/fonts/FiraSans-Regular.woff2') format('woff2');
+    font-weight: normal;
+    font-style: normal;
+}
+
 .output {
     font: 1rem 'Fira Sans', sans-serif;
     letter-spacing: 1px;
--- a/live-examples/html-examples/global-attributes/accesskey.html
+++ b/live-examples/html-examples/global-attributes/accesskey.html
@@ -1,0 +1,4 @@
+<p>If you need to relax, press the
+<strong><u>S</u></strong>tress reliever!</p>
+
+<button accesskey="s">Stress reliever</button>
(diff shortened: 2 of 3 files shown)


schalkneethling:
> In the interactive example version, it doesn't work in Firefox only. Very strange.

@chrisdavidmills You mean when it is embedded in an MDN page? It might be caused by it being in an `iframe`?

chrisdavidmills:
> @chrisdavidmills You mean when it is embedded in an MDN page? It might be caused by it being in an iframe?

Sorry, yes that's exactly what I mean. Hrm, so if it won't work inside an iframe, this example is kind of kaput ;-)

schalkneethling:
> Sorry, yes that's exactly what I mean. Hrm, so if it won't work inside an iframe, this example is kind of kaput ;-)

Well, you mentioned

> In the interactive example version, it doesn't work in Firefox only.

So, I assume it worked in all of the other browsers? Also, if you ensure focus is on the `iframe`, does it then work in Firefox? Not ideal indeed but, if Firefox is the only one being problematic, I would still say it is worth having this example out there.

chrisdavidmills:
> So, I assume it worked in all of the other browsers?

Duh, yes, sorry, I remember now ;-) Yes, it works in other browsers when in the iframe, even if it isn't specifically focused. But in Firefox, it doesn't seem to work even when it IS specifically focused.

schalkneethling:
> But in Firefox, it doesn't seem to work even when it IS specifically focused.

Might be bug file worthy? Seeing that, that is the only problem, I would merge and add this one to MDN.

chrisdavidmills:
Yup, let's merge then. I'll see if there is already a bug, and file one if not.

live-examples/html-examples/global-attributes/css/accesskey.css
```diff
@@ -0,0 +1,12 @@
+@font-face {
```
schalkneethling: I just noticed this. I think the other pull requests also had it. You "should" not need to do this as we are adding fonts here:

https://github.com/mdn/interactive-examples/blob/master/css/editor-libs/shadow-fonts.css

There was an issue where custom fonts would not work inside the showDOM output, and after some research I found info that required the above change. Does it not work for you without the `@font-face` block? If not, does it not work in all browsers or only a specific one?

chrisdavidmills: I just added it because I thought it was still needed. I'll not do that again then. ;-)

PR #1045, opened 2018-07-26
label for-attribute should reference id.

Row:


live-examples/html-examples/forms/datalist.html
```diff
@@ -1,5 +1,5 @@
 <label for="ice-cream-choice">Choose a flavor:</label>
-<input list="ice-cream-flavors" name="ice-cream-choice" />
+<input list="ice-cream-flavors" id="ice-cream-choice" />
```
schalkneethling: Nice catch! One suggestion though. The `input` still needs a `name` attribute so, perhaps add the `id` in addition to the name? They can safely share the same name. Thanks @Row

PR #1048, opened 2018-07-26
adding data-* global attribute interactive example

chrisdavidmills:


schalkneethling:
@chrisdavidmills I wonder if we should add the following to avoid scrolling in the output area?

```
h1 {
 margin: 0; 
}
...
ul {
 margin: 10px 0 0; 
}
```

chrisdavidmills:
Great idea man! Added.

Issue #729, opened 2018-04-19
[HTML Examples] Content sectioning ~ <h1>–<h6>: The HTML Section Heading elements


schalkneethling:
Need an HTML example for the HTML section heading elements
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

## Contributing

https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING-HTML.md

schalkneethling:
Note: This issue is being worked on by a community member.

MrBojangles:
like this?
<div>
  <h1>H1 Beetles</h1>
  <div style="margin-left: 10px;">
    <h2>H2 Etymology</h2>
  </div>
  <div style="margin-left: 10px;">
    <h2>H2 Distribution and Diversity</h2>
  </div>
  <div style="margin-left: 10px;">
    <h2>H2 Evolution</h2>
    <div style="margin-left: 10px;">
      <h3>H3 Late Paleozoic</h3>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Jurassic</h3>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Cretaceous</h3>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Cenozoic</h3>
    </div>
  </div>
  <div style="margin-left: 10px;">
    <h2>H2 External Morphology</h2>
    <div style="margin-left: 10px;">
      <h3>H3 Head</h3>
      <div style="margin-left: 10px;">
        <h4>H4 Mouthparts</h4>
      </div>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Thorax</h3>
      <div style="margin-left: 10px;">
        <h4>H4 Prothorax</h4>
      </div>
      <div style="margin-left: 10px;">
        <h4>H4 Pterothorax</h4>
      </div>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Legs</h3>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Wings</h3>
    </div>
    <div style="margin-left: 10px;">
      <h3>H3 Abdomen</h3>
    </div>
  </div>
</div>

chrisdavidmills:
Just to be clear, and so that noone's time is wasted, I think @stephanmax is working on this one. I'll assign him.

chrisdavidmills:
Ah, I can't assign him specifically...but at least we know now ;-)

[...]

PR #1053, opened 2018-07-26
nested heading html example

MrBojangles:
added example of nested heading with inline styles. Don't think I'm correct but would appreciate feedback. regards

### DIFF
--- a/live-examples/html-examples/content-sectioning/css/headings.css
+++ b/live-examples/html-examples/content-sectioning/css/headings.css
@@ -1,3 +1,0 @@
-.poo {
-  background-color: black;
-}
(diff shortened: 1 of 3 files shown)


wbamberg:
I'm sorry @MrBojangles , but @stephanmax is working on this one, as per the conversation in https://github.com/mdn/interactive-examples/issues/729. I'll leave this PR open for now in case that doesn't work out, but otherwise I think we should close it.

Sorry for not being clearer on assignments, and for you wasting your time as a result. I'm sure there will be more interactive examples coming in the future ...

MrBojangles:
No worries

PR #1059, opened 2018-07-26
adding style global attribute interactive example

chrisdavidmills:

### DIFF
--- a/live-examples/html-examples/global-attributes/meta.json
+++ b/live-examples/html-examples/global-attributes/meta.json
@@ -32,6 +32,17 @@
             "title": "HTML Demo: dir",
             "type": "tabbed",
             "height": "tabbed-standard"
+        },
+        "style-attr": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/global-attributes/style-attr.html",
+            "cssExampleSrc":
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/global-attributes/meta.json
```diff
@@ -32,6 +32,17 @@
...
             "type": "tabbed",
             "height": "tabbed-standard"
+        },
+        "style-attr": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/global-attributes/style-attr.html",
+            "cssExampleSrc":
+                "live-examples/html-examples/global-attributes/css/style-attr.css",
+            "fileName": "style-attr.html",
```
wbamberg: Nice catch to avoid the clash!

But I think we should do this consistently for all attributes, since they are in effect different namespaces. Only doing it in cases where we know they actually clash means it's easier to make mistakes. Also, I think we should call it "attribute-style" - this makes it clearer that they are a separate category and is consistent with what we have done for CSS functions: https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/transforms/meta.json#L13-L32

wbamberg:
Filed https://github.com/mdn/interactive-examples/issues/1065 for the follow-up.

PR #1060, opened 2018-07-26
adding tabindex global attribute interactive example

chrisdavidmills:

### DIFF
--- a/live-examples/html-examples/global-attributes/css/tabindex.css
+++ b/live-examples/html-examples/global-attributes/css/tabindex.css
@@ -1,0 +1,19 @@
+.output {
+    font: 0.9rem 'Fira Sans', sans-serif;
+}
+
+form {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 100px;
+}
+
+div,
+label {
+    letter-spacing: 0.5px;
+}
+
+div:focus {
+    font-weight: bold;
+}
--- a/live-examples/html-examples/global-attributes/css/tabindex.css
+++ b/live-examples/html-examples/global-attributes/css/tabindex.css
@@ -1,5 +1,5 @@
 .output {
-    font: 0.9rem 'Fira Sans', sans-serif;
+    font: .9rem 'Fira Sans', sans-serif;
 }
 
 form {
@@ -11,9 +11,13 @@ form {
 
 div,
 label {
-    letter-spacing: 0.5px;
+    letter-spacing: .5px;
 }
 
 div:focus {
     font-weight: bold;
+}
+
+.instructions {
+    font-style: italic;
 }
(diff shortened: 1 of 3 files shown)


live-examples/html-examples/global-attributes/tabindex.html
```diff
@@ -0,0 +1,5 @@
+<form>
+    <label>First in tab order: <input type="text"></label>
+    <div tabindex="0">Tabbable due to tabindex.</div>
+    <label>Third in tab order: <input type="text"></label>
+</form>
```
wbamberg: I found this one a little hard to use, because you have to know to focus the output pane before tabbing.

I am not generally that keen on including instructions in the example, but I think it might help here. I also think it would help to have a div that doesn't use tabindex and is therefore not tabbable. For example:

```
<p class="instructions">Click anywhere in this pane, then try tabbing through the elements.</p>

<form>
    <label>First in tab order: <input type="text"></label>
    <div tabindex="0">Tabbable due to tabindex.</div>
    <div>Not tabbable: no tabindex.</div>
    <label>Third in tab order: <input type="text"></label>
</form>
```

...with possible styling of the instructions:

```
.instructions {
    font-style: italic;
}
```

WDYT?

live-examples/html-examples/global-attributes/css/tabindex.css
```diff
@@ -0,0 +1,19 @@
+.output {
+    font: 0.9rem 'Fira Sans', sans-serif;
```
wbamberg: Omit leading zero please.

live-examples/html-examples/global-attributes/css/tabindex.css
```diff
@@ -0,0 +1,19 @@
...
+form {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 100px;
+}
+
+div,
+label {
+    letter-spacing: 0.5px;
```
wbamberg: Also here.

live-examples/html-examples/global-attributes/meta.json
```diff
@@ -32,6 +32,17 @@
...
+        "tabindex": {
+            "baseTmpl": "tmpl/live-tabbed-tmpl.html",
+            "exampleCode":
+                "live-examples/html-examples/global-attributes/tabindex.html",
+            "cssExampleSrc":
+                "live-examples/html-examples/global-attributes/css/tabindex.css",
+            "fileName": "tabindex.html",
+            "title": "HTML Demo: tabindex",
+            "type": "tabbed",
+            "height": "tabbed-standard"
```
wbamberg: Should also be "tabbed-shorter". I feel like I missed a bunch of these in my first reviews today. Oh well.

chrisdavidmills:
Thanks for the review @wbamberg ; I agree with all your comments here, and have made all the changes.

PR #1062, opened 2018-07-27
Update date-toutcstring.html

AugustinusVonHippo:
the new date is clearly no event, because of that the variable should not be named "event". I changed it to "date".

### DIFF
--- a/live-examples/js-examples/date/date-toutcstring.html
+++ b/live-examples/js-examples/date/date-toutcstring.html
@@ -1,7 +1,7 @@
 <pre>
-<code id="static-js">var event = new Date('14 Jun 2017 00:00:00 PDT');
+<code id="static-js">var date = new Date('14 Jun 2017 00:00:00 PDT');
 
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your contribution!

It's not an [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event), for sure, but I think the idea is that it's an "event" - like a birthday party or something like that. All the other Date examples follow a similar pattern (e.g. https://interactive-examples.mdn.mozilla.net/pages/js/date-tolocaletimestring.html).

But I can see this is obviously confusing :). I'd accept this PR if it changed the other Date examples from `event` to `date`.

wbamberg:
> I'd accept this PR if it changed the other Date examples from event to date.

On second thoughts, `date` is also not the best name for this variable (since it's just the same as the name of the type, but lowercased).

wbamberg:
Closing this one, given the lack of activity.

PR #1063, opened 2018-07-27
Added examples for String.prototype[@@iterator]()

irenesmith:


live-examples/js-examples/string/meta.json
```diff
@@ -1,5 +1,19 @@
 {
     "pages": {
+        "stringIterator": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/string/string-iterator.html",
+            "fileName": "string-iterator.html",
+            "title": "JavaScript Demo: String.prototype[@@iterator]()",
+            "type": "js"
+        },
+        "stringTrim": {
```
wbamberg: This should be stringIteratorForeach or something (but I don't think you actually need this example).

live-examples/js-examples/string/string-iterator-foreach.html
```diff
@@ -0,0 +1,8 @@
+<pre>
+<code id="static-js">var str = "The quick brown fox jumped over the lazy dog's back."
```
wbamberg: I think, since these show an ES6 feature (I think...) we should be able to use `const ` and `let` here, as per the style guide (https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6).

Here `const` seems like a good choice.

Also, please use single quotes for strings: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#single-quotes.

live-examples/js-examples/string/string-iterator-foreach.html
```diff
@@ -0,0 +1,8 @@
...
+
+for (var ch of str) {
```
wbamberg: Perhaps `let` instead of `var` here.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var str = "The quick red fox jumped over the lazy dog's back.";
```
wbamberg: Same comment as in the other example about `var` and single quotes.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
...
+
+  var iterator = str[Symbol.iterator]();
```
wbamberg: Again `const` might be a better fit here.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
...
+
+  do {
+    var theChar = iterator.next();
```
wbamberg: `let` or `const`

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
...
+    // output each character in the string, one per line
+    console.log(theChar);
```
wbamberg: To get the character this needs to be `theChar.value`.

live-examples/js-examples/string/string-iterator.html
wbamberg: This line and subsequent lines is getting 2 spaces of extra indentation that shouldn't be there.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
...
+  } while(!theChar.done)
```
wbamberg: With this logic the last time round will output `Object { value: undefined, done: true }` (or just `undefined` if you call `theChar.value`). I'm not sure if this is intentional.

If not you could avoid it by doing something like:

```
let theChar = iterator.next();

while (!theChar.done) {
  // output each character in the string, one per line
  console.log(theChar.value);
  theChar = iterator.next();
} 
```

PR #1067, opened 2018-07-30
Estelle/stringmethods

estelle:
7 new live examples for string methods, and updated meta.json to go with.


estelle:
Thanks for the super helpful comments. Updates have been made and committed.  @wbamberg

live-examples/js-examples/string/string-tolocaleuppercase.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var cedilla = '\u00E7';
+
+console.log(cedilla.toLocaleUpperCase('en-US'));
+// expected output: "Ç"
+
+console.log(cedilla.toLocaleUpperCase('TR'));
+// expected output: "Ç"
+</code>
```
wbamberg: This seems to give the same output for both locales to me (and the same output as toUpperCase as well) so it doesn't seem to be a good demonstration of the API. We could do the same as toLocaleLowerCase, but in reverse:

```
var city = 'istanbul';

console.log(city.toLocaleUpperCase('en-US'));
// expected output: "ISTANBUL"

console.log(city.toLocaleUpperCase('TR'));
// expected output: "İSTANBUL"
```
What do you think? I might be missing some subtlety here though...

wbamberg:
Thanks for the updates @estelle ! This looks great, except I had a question about the example for `toLocaleUpperCase`.

live-examples/js-examples/string/string-tolocaleuppercase.html
estelle: Changed to your example in my other pull request. Much better. Thank you.

live-examples/js-examples/string/meta.json
```diff
@@ -20,6 +20,55 @@
...
+            "type": "js"
+        },
+        "stringToLowerCase": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/string/string-tolowercase.html",
+            "fileName": "string-tolowercase.html",
+            "title": "JavaScript Demo: String.toLowerCase()",
+            "type": "js"
+        },
+        "stringToUpperCase": {
```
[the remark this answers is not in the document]
estelle: fixed.

wbamberg:
I think we can close this one since these examples are now in https://github.com/mdn/interactive-examples/pull/1095.

PR #1068, opened 2018-07-31
Add String concat, ends/startswith, padend/start live examples

meowwwls:
Adds 5 new live examples for string methods, along with an updated `meta.json`:

- `String.concat()`
- `String.padEnd()`
- `String.padStart()`
- `String.endsWith()`
- `String.startsWith()`


live-examples/js-examples/string/string-concat.html
```diff
@@ -0,0 +1,11 @@
...
+<code id="static-js">var str1 = 'Hello';
+var str2 = 'World';
+
+console.log(str1.concat(' ', str2));
+// expected output: 'Hello World'
+
+console.log(str2.concat(', ', str1));
+// expected output: 'World, Hello'
+</code>
+</pre>
```
wbamberg: Files should end with a newline, here and for the other examples too.

live-examples/js-examples/string/string-endswith.html
```diff
@@ -0,0 +1,12 @@
...
+<code id="static-js">var str1 = 'Cats are the best';
+
+console.log(str1.endsWith('best'));
+// expected output: true
+
+var str2 = 'Is this a question';
+
+console.log(str2.endsWith('?'));
+// expected output: false
+</code>
```
wbamberg: Do you think it is worth showing the `length` parameter too, using something like:

```
console.log(str1.endsWith('best', 16));
// expected output: false
```

live-examples/js-examples/string/string-concat.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">var str1 = 'Hello';
+var str2 = 'World';
+
+console.log(str1.concat(' ', str2));
+// expected output: 'Hello World'
```
wbamberg: Very nit-picky, but expected output should exactly match the actual output, which here (and also below) means double quotes:

```
 // expected output: "Hello World" 
```

Sorry to be so nit-picky, but we have considered using "expected output:" for testing, and if we ever do that it will help for the content to match exactly.

live-examples/js-examples/string/string-padstart.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">var str1 = '5';
+
+console.log(str1.padStart(2, '0'));
+// expected output: '05'
+
+var last4Digits = '2034399002125581'.slice(-4);
+var maskedNumber = last4Digits.padStart(16, '*');
```
wbamberg: I think it might be better to have a variable for the full number, and use its length instead of a hardcoded 16 here. That way the length is only represented in one place:

```
var fullNumber = '2034399002125581';
var last4Digits = fullNumber.slice(-4);
var maskedNumber = last4Digits.padStart(fullNumber.length, '*');
```

meowwwls: Thank you so much for the comments! I can't make the fixes until late Monday or early Tuesday, but as soon as I can, I will make all of the suggested changes!

meowwwls:
I was actually able to make the changes quickly right now, so hopefully, everything is OK now!

meowwwls:
> Sorry I wasn't clear about the package-lock - I think the changes should have been omitted from the PR, not that the file should be deleted. I've tried to revert that change in your PR, hopefully I didn't break anything :).

I'm sorry about that! I should have asked for clarification.

PR #1069, opened 2018-07-31
Added method for String[@@iterator]()

irenesmith:


live-examples/js-examples/string/meta.json
```diff
@@ -1,5 +1,19 @@
 {
     "pages": {
+        "stringIterator": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
+            "exampleCode": "live-examples/js-examples/string/string-iterator.html",
+            "fileName": "string-iterator.html",
+            "title": "JavaScript Demo: String.prototype[@@iterator]()",
+            "type": "js"
+        },
+        "stringTrim": {
```
wbamberg: I think this entry needs to be removed.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">const str = "The quick red fox jumped over the lazy dog's back.";
```
wbamberg: Use single quotes for strings, please: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#single-quotes

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,13 @@
...
+
+let iterator = str[Symbol.iterator]();
+let theChar = iterator.next();
+
+while(!theChar.done) {
+  // output each character in the string, one per line
+  console.log(theChar);
```
wbamberg: This doesn't output the character. To output the character you need `theChar.value`

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,13 @@
...
+  theChar = iterator.next();
+}
```
wbamberg: It feels a bit odd to use a while loop here - in this situation (where we want to iterate through every value) it would be more natural to use `for...of`. But I understand the value of showing `next()`, `done` and so on.

What do you think about something like:

```
while(!theChar.done && theChar.value !== ' ') {
  // output each character in the first word, one per line
  console.log(theChar.value);
  theChar = iterator.next();
} 
```

...this is a slightly less natural fit for `for...of`, because you'd have to break out of the loop. What do you think?

live-examples/js-examples/string/meta.json
irenesmith: Yep. Don't know where it came from.

live-examples/js-examples/string/string-iterator.html
irenesmith: Done

live-examples/js-examples/string/string-iterator.html
irenesmith: I had been showing the whole value to make it clear that theChar is actually an object with value and done fields, but I've changed this to just show the value.

live-examples/js-examples/string/string-iterator.html
irenesmith: Made the change. I know it seemed contrived, but I wanted to make it clear what was going on.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,14 @@
+<pre>
+<code id="static-js">const str = 'The quick red fox jumped over the lazy dog's back.';
```
wbamberg: The example fails with "Error: unexpected token: identifier" because you need to escape the apostrophe in "dog's".

irenesmith:
I saw that. Thank you!

PR #1070, opened 2018-07-31
Created an example for the String.raw() method

irenesmith:


live-examples/js-examples/string/meta.json
```diff
@@ -1,5 +1,19 @@
 {
     "pages": {
+        "stringIterator": {
```
wbamberg: It looks like meta.json hasn't been updated - it included entries from another PR, but not one for `raw`

live-examples/js-examples/string/string-iterator-foreach.html
```diff
@@ -0,0 +1,8 @@
+<pre>
```
wbamberg: This example should not be included here.

live-examples/js-examples/string/string-iterator.html
```diff
@@ -0,0 +1,12 @@
+<pre>
```
wbamberg: This example should not be included here.

live-examples/js-examples/string/string-raw.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">const name = 'Bob';
+
+// The next line returns the template with the
+// value of the name variable inserted.
+console.log(string.raw`<p>Good morning, {name}</p>`);
```
wbamberg: This logs "Error: string is not defined" - I think it needs `String` not `string`. Also it would need to be ${name} for the substitution to work.

live-examples/js-examples/string/string-raw.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">const name = 'Bob';
+
+// The next line returns the template with the
+// value of the name variable inserted.
```
wbamberg: This isn't exactly right - the next line logs a string to the console.

If possible, please use the "expected output:" format to label console output, as per https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#indicating-consolelog-output:

```
console.log(String.raw`<p>Good morning, ${name}</p>`);
// expected output: "<p>Good morning, Bob</p>"
```

live-examples/js-examples/string/string-raw.html
```diff
@@ -0,0 +1,9 @@
...
+console.log(string.raw`<p>Good morning, {name}</p>`);
+</code>
+</pre>
+
```
wbamberg: Please remove the spaces from this line, and all files should end with a new line.

live-examples/js-examples/string/string-raw.html
wbamberg: But also, I'm not sure this is an ideal example of `raw`. With this string, just console.log gives exactly the same output:

```
console.log(`<p>Good morning, ${name}</p>`);
```

(I'm also not sure what the `<p>`/`</p>` tags are for here)

As I understand it, the point of `raw` is that it doesn't process escapes:

```
console.log(String.raw`Good morning,\n ${name}`);
// expected output: "Good morning,\n Bob"

console.log(`Good morning,\n ${name}`);
// expected output: "Good morning,
//                  Bob"
```

irenesmith:
Yes, and of course I didn't push it.

live-examples/js-examples/string/string-raw.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">const name = 'Bob';
+
```
wbamberg: This `name` declaration and the blank line should be removed.

live-examples/js-examples/string/string-raw.html
```diff
@@ -0,0 +1,11 @@
...
+// Create a variable that uses a Windows
+// path without escaping the backslashes:
+const filePath = String.raw`C:\Development\profile\aboutme.html`;
+console.log('The file was uploaded from: ' + filePath);
+// expected output: "The file was uploaded from: C:\Development\profile\aboutme.html"
+</code>
+</pre>
+
```
wbamberg: Files should end with a new line, and please remove the spaces from this line.

irenesmith: Fixed. I think VS Code is adding the spaces because it expects the line to be indented. I will watch out for that in the future.

live-examples/js-examples/string/string-raw.html
irenesmith: Removed

PR #1071, opened 2018-07-31
Adding an example of String.split()

irenesmith:


live-examples/js-examples/string/meta.json
```diff
@@ -1,5 +1,19 @@
 {
     "pages": {
+        "stringIterator": {
```
wbamberg: It looks like this PR has the same issues with meta.json as https://github.com/mdn/interactive-examples/pull/1070.

live-examples/js-examples/string/string-iterator-foreach.html
```diff
@@ -0,0 +1,8 @@
+<pre>
```
wbamberg: Again this and `string-iterator.html` should not be included.

live-examples/js-examples/string/string-split.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">const str = "The quick brown fox jumped over the lazy dog.";
```
wbamberg: Please use single quotes for strings.

wbamberg: Since `split()` is not ES6, please don't use ES6 constructs like `const`: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6.

live-examples/js-examples/string/string-split.html
```diff
@@ -0,0 +1,7 @@
...
+
+words = str.split(" ");
+console.log(words[3]); // fox
```
wbamberg: Please use the `expected output:` format (https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#indicating-consolelog-output).

live-examples/js-examples/string/string-split.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">const str = "The quick brown fox jumped over the lazy dog.";
+
+words = str.split(" ");
```
wbamberg: This should be `var words =` not just `words =` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Declaring_variables).

live-examples/js-examples/string/string-split.html
```diff
@@ -0,0 +1,7 @@
...
+console.log(words[3]); // fox
+</code>
```
wbamberg: I think it would be good to show a couple of other examples of the syntax, e.g.

```
var result2 = str.split("");
console.log(result2[0]);
```

irenesmith:
Added two more examples and made sure that the expected output format for comments was used.

PR #1074, opened 2018-07-31
Adding an example for the String.substring() method

irenesmith:


live-examples/js-examples/string/string-substring.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">const str = 'Mozilla';
+console.log(str.substring(1,2)); // 'oz'
+console.log(str.substring(2,str.length)); 'zilla'
+</code>
+</pre>
+
```
wbamberg: Please remove this line containing just spaces, ad all files should end with a return character (that's what the 
screen shot 2018-08-10 at 12 22 07 pm is telling you).

live-examples/js-examples/string/string-substring.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">const str = 'Mozilla';
+console.log(str.substring(1,2)); // 'oz'
+console.log(str.substring(2,str.length)); 'zilla'
```
wbamberg: There's no point doing this: `str.substring(2,str.length)`, because it's just the same as `str.substring(2)`.

In fact let's use `str.substring(2)` instead on this line, then we can illustrate the optional parameter.

live-examples/js-examples/string/string-substring.html
irenesmith: Done

live-examples/js-examples/string/string-substring.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var str = 'Mozilla';
+
+console.log(str.substring(1, 2));
+// expected output "oz"
+
+console.log(str.substring(2));
+// expected output "zilla"
+</code>
+</pre>
```
wbamberg: This line still seems to end without a newline.

live-examples/js-examples/string/string-substring.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var str = 'Mozilla';
+
+console.log(str.substring(1, 2));
+// expected output "oz"
```
wbamberg: This actually returns "o" not "oz". Sorry, should have caught this last time round.

Also, sorry to be super-nitpicky, but the format is:

```
// expected output: thing
```

Note the colon. Sorry to be so nitpicky, but it's worth being exact about this I think.

irenesmith: I should have caught "oz" too. I fixed the code so it produces the intended result and also fixed the comments.

live-examples/js-examples/string/string-substring.html
irenesmith: There should be one now.

live-examples/js-examples/string/string-substring.html
```diff
@@ -0,0 +1,10 @@
+<pre>
+<code id="static-js">var str = 'Mozilla';
+
+console.log(str.substring(1, 2));
+// expected output: "oz"
```
wbamberg: This still seems to be wrong. It should be either `(1, 3)` or `expected output: "o"`

PR #1075, opened 2018-07-31
Created an example for the String.toLowerCase() method

irenesmith:


wbamberg:
This one is I think being handled in https://github.com/mdn/interactive-examples/pull/1067.

PR #1080, opened 2018-07-31
Added an example for String.length

irenesmith:


live-examples/js-examples/string/string-length.html
```diff
@@ -0,0 +1,5 @@
+<pre>
+<code id="static-js">const str='Life, the universe and everything. Answer:';
+console.log(str + ' ' + str.length); // 42
+</code>
+</pre>
```
wbamberg: Please end files with a newline.

live-examples/js-examples/string/string-length.html
```diff
@@ -0,0 +1,5 @@
+<pre>
+<code id="static-js">const str='Life, the universe and everything. Answer:';
```
wbamberg: Stick with ES5 for ES5 examples (https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6) - this means using `var` here.

wbamberg: Also please use spaces for readability: `var str = '...`

live-examples/js-examples/string/string-length.html
```diff
@@ -0,0 +1,5 @@
...
+console.log(str + ' ' + str.length); // 42
```
wbamberg: Please use the standard `expected output: ` format.

live-examples/js-examples/string/string-length.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">const str = 'Life, the universe and everything. Answer:';
```
wbamberg: This should be `var` not `const`.

live-examples/js-examples/string/string-length.html
```diff
@@ -0,0 +1,7 @@
...
+  
+console.log(str + ' ' + str.length);
+// expected result: 42
```
wbamberg: "expected output", not "expected result". Sorry to be picky, but we might as well be consistent here.

Also the output is `"Life, the universe and everything. Answer: 42"`, not `42`.

irenesmith:
You should be picky. Don't know why I said the wrong thing. Resolved both of these comments.

PR #1081, opened 2018-07-31
Adding an example for String.slice()

irenesmith:


live-examples/js-examples/string/string-slice.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">var str = 'The quick red fox jumped over the lazy dog\'s back.';
+console.log(str.slice(30)) // the lazy dog's back.
+console.log(str.slice(4,17)); // quick red fox
+</code>
+</pre>
+
```
wbamberg: Please remove these trailing spaces, and the file should end with a return character.

live-examples/js-examples/string/string-slice.html
```diff
@@ -0,0 +1,7 @@
+<pre>
+<code id="static-js">var str = 'The quick red fox jumped over the lazy dog\'s back.';
```
wbamberg: It would be good to have a blank line after this, for readability.

live-examples/js-examples/string/string-slice.html
```diff
@@ -0,0 +1,7 @@
...
+console.log(str.slice(30)) // the lazy dog's back.
```
wbamberg: Please use the standard `expected output:` format, and a blank line between each example, like:
```
console.log(str.slice(30));
// expected output: "the lazy dog's back."

console.log(str.slice(4,17));
// expected output: ...
```

wbamberg: Also this line is missing the semicolon.

live-examples/js-examples/string/string-slice.html
irenesmith: Done

live-examples/js-examples/string/string-slice.html
irenesmith: Done

live-examples/js-examples/string/string-slice.html
irenesmith: Fixed the missing semicolon and reformatted the comments.

irenesmith:
I also added two new examples that show using negative numbers.

PR #1087, opened 2018-08-02
Update <img> to use CC0 image

a2sheppy:
Updated the  example to use a CC0 image and
removed the attribution footer


schalkneethling:
@a2sheppy Thanks for the update. With this change, we also do not need the footer styles:

```
footer {
    font-size: .8rem;
    position: absolute;
    bottom: 10px;
    right: 10px;
}
```

The image is also a bit to tall for the output area causing scrolling so, we should size it smaller. Adding `width="250"` to the image tag solves the problem.

PR #1091, opened 2018-08-03
Update cc0 figcaption

a2sheppy:
Replaced the hamster image (which was free but not expressly labeled cc0) with a silhouetted elephant. Wrote a new caption, and updated the CSS to work with the shape of the image better.


live-examples/html-examples/text-content/css/figcaption.css
```diff
@@ -4,10 +4,10 @@ figure {
     flex-flow: column;
     margin: 0;
     padding: 5px;
+    width: 22em;
```
schalkneethling: I would suggest making this `width: 19em;` as `22` causes the output area to scroll vertically.

a2sheppy:
While it doesn't require scrolling for me, it's close, so I presume it's due to local font configuration differences. I will submit an update momentarily.

PR #1092, opened 2018-08-03
Update main element example to use cc0 image

a2sheppy:
Replaced the lizard image used here to a known CC0
one.


live-examples/html-examples/text-content/main.html
```diff
@@ -1,5 +1,5 @@
 <header>
-    <img class="logo" src="/media/examples/lizard.png" />
+    <img class="logo" src="/media/cc0-images/Gecko--320x213.jpg" />
```
schalkneethling: This image is sized really small. I know this is not introduced by this PR but, perhaps you can also go ahead and change the size from:

```
header > .logo {
    height: 2.5rem;
...
```

to

```
header > .logo {
    height: 4.5rem;
...
```

PR #1093, opened 2018-08-03
Update for...in example to use const

eszjay:


wbamberg:
I think according to the style guide (https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6) we need to use `var` for this example, don't we?

PR #1097, opened 2018-08-07
String endswith

irenesmith:
Updated the metadata and removed the two examples that don't belong in this branch.


wbamberg:
An example for this was already added in https://github.com/mdn/interactive-examples/pull/1068.

PR #1098, opened 2018-08-07
Added startsWith example

irenesmith:
Added the example and also updated the metadata.


wbamberg:
An example for this was already added in https://github.com/mdn/interactive-examples/pull/1068.

PR #1102, opened 2018-08-07
Remove background from option example

a2sheppy:
Removed image and footer.

### DIFF
--- a/live-examples/html-examples/forms/css/option.css
+++ b/live-examples/html-examples/forms/css/option.css
@@ -1,9 +1,5 @@
-label,
-footer {
-    font-family: sans-serif;
-}
-
 label {
+    font-family: sans-serif;
     font-size: 1rem;
     padding-right: 10px;
 }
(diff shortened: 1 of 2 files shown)


a2sheppy:
@wbamberg Yes, the hamster.jpg image is still being used elsewhere, though I think that's only until all the PRs are merged. I'm not certain. It's something to look at in a follow-up.

wbamberg:
> @wbamberg Yes, the hamster.jpg image is still being used elsewhere, though I think that's only until all the PRs are merged. I'm not certain. It's something to look at in a follow-up.

```
$ grep -r hamster ./live-examples/
./live-examples//html-examples/forms/select.html:    <option value="hamster">Hamster</option>
./live-examples//html-examples/forms/css/select.css:    background: center/cover no-repeat url('/media/examples/hamster.jpg');
./live-examples//html-examples/forms/option.html:    <option value="hamster">Hamster</option>
./live-examples//html-examples/text-content/figcaption.html:    <img src="/media/examples/hamster.jpg" alt="a cute hamster" />
```

PR #1111, opened 2018-08-16
Improve example code

aeinbu:
Both the class name and example were confusing.


wbamberg:
Yes, agreed. Thanks @aeinbu !

Issue #1119, opened 2018-08-31
`box-decoration-break` example output is hidden


hacknug:
As stated, the example for how the `box-decoration-break` property works, has the class `.hidden` on the output element. This is not good because [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break) readers can not see the live example and they need to rely on the screenshots.

https://github.com/mdn/interactive-examples/blob/cc5ae0a4dd019db6eda4dd60e88be19b94e4f437/live-examples/css-examples/fragmentation/box-decoration-break.html#L17

I can submit a PR if needed. Just let me know.

wbamberg:
Thanks for filing!

I think the problem is not the "hidden" class. I think you are using a browser that does not support this property, or that needs it prefixed (https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break#Browser_compatibility).

At the moment, if this is the case, the example editor displays something like this:

<img width="601" alt="screen shot 2018-08-31 at 7 59 21 am" src="https://user-images.githubusercontent.com/432915/44920040-cb248a00-acf3-11e8-8214-cc8470e9545b.png">

Try the page in Firefox, it should look like:

<img width="747" alt="screen shot 2018-08-31 at 8 00 04 am" src="https://user-images.githubusercontent.com/432915/44920104-e8595880-acf3-11e8-9f07-c644cad5c00b.png">

I know the way we handle unsupported properties is not ideal and would love to have a better way to handle it - this is discussed at some length here: https://github.com/mdn/interactive-examples/issues/532. We have bigger priorities at the moment though, mostly because most properties are in fact well supported across browsers.

In general we've handled this by just removing the examples for properties that don't have good cross-browser support (e.g. all the [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties)). So perhaps we should do the same for `box-decoration-break` - sad though, as it is a nice example.

hacknug:
I'm using vanilla Chrome (v68.0.3440.106) and it supports the property. If I remove the `.hidden` class from the output element this is what I see:

<img width="1063" alt="captura de pantalla 2018-08-31 a las 17 09 06" src="https://user-images.githubusercontent.com/1107521/44920595-a99dbf80-ad40-11e8-8c25-0c7dcc4d0b4d.png">

wbamberg:
> I'm using vanilla Chrome (v68.0.3440.106) and it supports the property

[...]

PR #1120, opened 2018-08-31
Support prefixes

wbamberg:
This was inspired by https://github.com/mdn/interactive-examples/issues/1119 and is intended to be a simple way to enable us to have examples for properties that are prefixed in some browsers.

It just changes `data-property` from a single value to a space-separated list of values. The editor then considers that the example will work if any of the given properties are supported in the browser.

The example author can then pass the unprefixed form and any prefixed forms into `data-property`, and then include prefixed and unprefixed variants in the example choices.

I've updated a couple of examples to show how it works:

* [`text-emphasis`](https://interactive-examples.mdn.mozilla.net/pages/css/text-emphasis.html)
* [`box-decoration-break`](https://interactive-examples.mdn.mozilla.net/pages/css/box-decoration-break.html) (this one is a bit gross - it appears as if Chrome does not automatically re-evaluate this property unless something else is change, so I'm also tweaking the `box-shadow` when we change choices. I'm not very happy about this and would be happy to learn of alternatives.)


js/editor-libs/mce-utils.js
```diff
@@ -26,16 +26,25 @@ module.exports = {
     isPropertySupported: function(dataset) {
         'use strict';
 
-        var property = dataset['property'];
         /* If there are no 'property' attributes,
            there is nothing to test, so return true. */
-        if (property === undefined) {
+        if (dataset['property'] === undefined) {
```
schalkneethling: Any specific reason you do not want to store this in a variable?

js/editor-libs/mce-utils.js
```diff
@@ -26,16 +26,25 @@ module.exports = {
...
             return true;
         }
 
+        /* 'property' may be a space-separated list of properties. */
+        var properties = dataset['property'].split(' ');
+        /* Iterate through properties: if any of them apply,
+        the browser supports this example. */
+        var supported = false;
         var tmpElem = document.createElement('div');
+        for (var i = 0; i < properties.length; i++) {
```
schalkneethling: Nit: These arrays will obviously never be large but, caching the length is generally a best practice:

```
for (var i = 0, l = properties.length; i < l; i++) {
```

js/editor-libs/mce-utils.js
wbamberg: Because then I'd have extra temporary state - I only need this variable so I can immediately slice it. So I think this:

```
var propertiesAttribute = dataset['property'];
if (propertiesAttribute === undefined) {
            return true;
        }

/* 'property' may be a space-separated list of properties. */
var properties = propertiesAttribute.split(' ');
```

is less clear than this:

```
if (dataset['property'] === undefined) {
    return true;
}

/* 'property' may be a space-separated list of properties. */
var properties = dataset['property'].split(' ');
```

But it's not my code, I'd be happy to change it of you prefer.

Issue #1115, opened 2018-08-24
CSS position example doesn't work in Safari


guillaumebreux:
**Related article:** https://developer.mozilla.org/en-US/docs/Web/CSS/position

After looking at this example on safari, you can see that there is no prefix on the sticky property causing the example to fail.

But, even if you set the prefix, it won't work because the element is a `inline-block` element and Safari seems to only apply this position value on a `block` element (see here: http://jsfiddle.net/hg6Lqk7n/8/)

I don't know if the solution is to wait for safari to fix the issue or if you could tweak the exemple a little to make it work or to warn the user that it will fail on safari (v11.1.2) ? 

Anyway, the prefix is still needed and I could find where to make the change to make a PR :(

wbamberg:
Thanks for filing!

> I don't know if the solution is to wait for safari to fix the issue or if you could tweak the exemple a little to make it work or to warn the user that it will fail on safari (v11.1.2) ?

There's some previous chat about how the editor should behave when properties aren't well supported across browsers: https://github.com/mdn/interactive-examples/issues/532. We have a proposal for how to handle this, but it's not a very high priority at the moment just because cross-browser support is in general pretty good. Still it would be great to have a fix for this in the editor.

In this case, though, I think it would be possible for the example to work in Safari, and it would be worth doing. We could add `position: -webkit-sticky;` to the [CSS for the `sticky` option](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/positioned-layout/position.html#L27-L28), and [make the box `display: block;`](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/positioned-layout/position.css#L11). We'd also need to adjust the markup, but it seems simple - we could just remove those 2 blue boxes I think, and then adjust the red dotted box so it overlaps with the yellow box.

If you want to file a PR, please do go ahead! Otherwise we can leave this open and perhaps someone will take it before I manage to fix it.

The changes would all be in [position.html](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/positioned-layout/position.html), plus changes to the positioning of `#example-element-placeholder` in [position.css](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/positioned-layout/position.css). The [general contributing guide](https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md) covers how to make and test changes.

schalkneethling:
[...]

PR #1121, opened 2018-08-31
Make position:sticky example work in Safari

wbamberg:
Fix for https://github.com/mdn/interactive-examples/issues/1115.

This adds the `-webkit` prefix for `position: sticky` and makes the element `display: block`. Those changes make it work in Safari.

Because Safari wants the element to be `display: block`, this broke the original layout, so I've ended up simplifying it quite a lot.

### DIFF
--- a/live-examples/css-examples/positioned-layout/position.css
+++ b/live-examples/css-examples/positioned-layout/position.css
@@ -7,13 +7,10 @@
     width: 100%;
 }
 
-.box {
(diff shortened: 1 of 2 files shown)


wbamberg:
@estelle , would you mind reviewing this?

wbamberg:
@chrisdavidmills , could you take a look at this PR please? Thanks!

PR #1123, opened 2018-09-02
rest parameters initial example

bobtherobot:
A simpler initial example.


wbamberg:
Sounds good. I'm going to close this PR now though, because we're planning to make a big disruptive change to the repo (merging this branch: https://github.com/mdn/interactive-examples/tree/webapi-editor) and it will create conflicts for all open PRs.

If we want to update the rest parameters example, it would be better ti file a fresh PR after that merge I think.

Thanks for your contribution!

PR #1124, opened 2018-09-03
Added more helpful technique for accessing promise

justsml:
Showing that a promise evaluates to the string descriptor: `[object Promise]` is not the **most helpful** thing to include in the _first_ example.

Let me know if I need to revise, thanks for any feedback! 💯

### DIFF
--- a/live-examples/js-examples/promise/promise-constructor.html
+++ b/live-examples/js-examples/promise/promise-constructor.html
@@ -2,22 +2,16 @@
 <code id="static-js">var promise1 = new Promise(function(resolve, reject) {
   setTimeout(function() { 
     resolve('foo');
-  }, 100);
+  }, 300);
 });
 
+promise1.then(function(value) { 
+  console.log(value)
+  // expected output: "foo"
+});
 
-// How to show the value of promise1: 
-
-promise1.then(value => console.log(value));
-//=> "foo"
-
-// A Promise's value can ONLY be accessed via a `.then()` callback function:
-// promise1.then(function(value) {/* handle value in function here */})
-
-// The following example **won't work**:
-// Reading Synchronously:
 console.log(promise1);
-//=> [object Promise]
+// expected output: [object Promise]
 
 </code>
 </pre>
--- a/live-examples/js-examples/promise/promise-constructor.html
+++ b/live-examples/js-examples/promise/promise-constructor.html
@@ -1,9 +1,23 @@
 <pre>
 <code id="static-js">var promise1 = new Promise(function(resolve, reject) {
-  setTimeout(resolve, 100, 'foo');
+  setTimeout(function() { 
+    resolve('foo');
+  }, 100);
 });
 
+
+// How to show the value of promise1: 
+
+promise1.then(value => console.log(value));
+//=> "foo"
+
+// A Promise's value can ONLY be accessed via a `.then()` callback function:
+// promise1.then(function(value) {/* handle value in function here */})
+
+// The following example **won't work**:
+// Reading Synchronously:
 console.log(promise1);
-// expected output: [object Promise]
+//=> [object Promise]
+
 </code>
 </pre>
--- a/live-examples/js-examples/promise/promise-constructor.html
+++ b/live-examples/js-examples/promise/promise-constructor.html
@@ -6,7 +6,7 @@
 });
 
 promise1.then(function(value) { 
-  console.log(value)
+  console.log(value);
   // expected output: "foo"
 });
 


live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
 <pre>
 <code id="static-js">var promise1 = new Promise(function(resolve, reject) {
-  setTimeout(resolve, 100, 'foo');
```
justsml: Using `setTimeout` like this, with a 3rd argument ('foo') -  is IMHO rather rare.

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
...
+  setTimeout(function() { 
+    resolve('foo');
+  }, 100);
 });
 
+
+// How to show the value of promise1: 
+
+promise1.then(value => console.log(value));
```
justsml: I know this is the `constructor` example- I feel this bit is needed because people see `constructor` in their search, then think assume they will find how to make (and perhaps read) promises there." Well... Not exactly on this page. 😿 

While there are good supporting links and contextual info, it can easily be missed, especially the first time you land on the example/page.

The `.then()` line is a far more common use case, and I hope will benefit many users. 🤞 

> Note: I wouldn't add anything extraneous like `.catch()`

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
 <pre>
 <code id="static-js">var promise1 = new Promise(function(resolve, reject) {
-  setTimeout(resolve, 100, 'foo');
+  setTimeout(function() {
```
wbamberg: I think it's weird to mix arrow functions and non-arrow functions, is there a reason you're doing this?

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
 <pre>
 <code id="static-js">var promise1 = new Promise(function(resolve, reject) {
-  setTimeout(resolve, 100, 'foo');
+  setTimeout(function() { 
+    resolve('foo');
+  }, 100);
```
wbamberg: If we want to do this we should bump this a bit so the delay is noticeable (but not annoying). 300 or 400 ms would work better.

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
...
 });
 
+
+// How to show the value of promise1:
```
wbamberg: This comment isn't needed.

live-examples/js-examples/promise/promise-constructor.html
wbamberg: Yeah, I can see the value of showing how to use the promise in this example.

(really I suspect about 95% of the people who land on this page should be reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises instead)

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
...
+  setTimeout(function() { 
+    resolve('foo');
+  }, 100);
 });
 
+
+// How to show the value of promise1: 
+
+promise1.then(value => console.log(value));
+//=> "foo"
```
wbamberg: We use the convention:

```
// expected output: "foo"
```

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
...
+
+// A Promise's value can ONLY be accessed via a `.then()` callback function:
+// promise1.then(function(value) {/* handle value in function here */})
+
+// The following example **won't work**:
+// Reading Synchronously:
```
wbamberg: I don't think these comments are needed. This is an example, it is not the documentation. We should be very sparing with comments in these examples IMO - we should aim to show not tell. There are plenty of words in the rest of the page for people to read.

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -1,9 +1,23 @@
...
 console.log(promise1);
-// expected output: [object Promise]
+//=> [object Promise]
```
wbamberg: Again, we use the convention:

```
// expected output: [object Promise]
```

justsml:
Thanks for the feedback @wbamberg !

live-examples/js-examples/promise/promise-constructor.html
justsml: Let me know if I need to do any other changes. Thanks for your time.

justsml:
Thanks @wbamberg !!!

PR #1125, opened 2018-09-03
Clarifying Promise.resolve() Example

justsml:
Thanks in advance & feedback always welcome 😇

### DIFF
--- a/live-examples/js-examples/promise/promise-resolve.html
+++ b/live-examples/js-examples/promise/promise-resolve.html
@@ -1,9 +1,9 @@
 <pre>
-<code id="static-js">var promise1 = Promise.resolve([1, 2, 3]);
+<code id="static-js">var promise1 = Promise.resolve(123);
 
 promise1.then(function(value) {
   console.log(value);
-  // expected output: [1, 2, 3]
+  // expected output: 123
 });
 
 Promise.resolve(42)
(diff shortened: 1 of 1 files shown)


wbamberg:
Sorry if this is obvious, but could you elaborate on why you want to make this change to the original example? That is, on how this helps explain `Promise.resolve()`? Thanks!

wbamberg:
Oh, so it's the array argument in particular that you think might be misleading? Then perhaps we could just change that, like:

```
var promise1 = Promise.resolve('foo');

promise1.then(function(value) {
  console.log(value);
  // expected output: "foo"
});
```

Do you think that would be enough? I do want to keep these examples as simple as possible (but no simpler).

live-examples/js-examples/promise/promise-resolve.html
```diff
@@ -1,9 +1,14 @@
...
 
 promise1.then(function(value) {
   console.log(value);
-  // expected output: Array [1, 2, 3]
+  // expected output: 123
 });
+
+Promise.resolve(42)
+  .then(console.log);
+// expected output: 42
```
schalkneethling: I reckon we can remove these last three lines. It might confuse especially newcomers who will then focus more on understanding this syntax than on `Promise.resolve`. So in the end the example should be just:

```
var promise1 = Promise.resolve(123);

promise1.then(function(value) {
  console.log(value);
  // expected output: 123
});
```

PR #1135, opened 2018-09-07
bug 1483072: handle new jenkins/aws services

escattone:
This PR allows this Jenkinsfile to run on both the new IT-owned Jenkins service as well as the existing MozMEAO-owned Jenkins service.

- Changes the name of the S3 bucket (to which the build artifacts will be sync'ed) depending upon the Jenkins service.
- Change the `bin/irc-notify.sh` script to use `RUN_DISPLAY_URL` instead of the fixed `BLUE_BUILD_URL`

I also took the liberty of making the indentation consistent (4 spaces). For easier review, you may want to check the `Hide whitespace changes` checkbox under the `Diff settings` button.

These changes have been successfully tested on both the new IT-owned Jenkins service and the existing MozMEAO-owned Jenkins service.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,33 +2,42 @@
 
 def buildSite() {
     stage ('build') {
-      try {
-        sh 'bin/build.sh'
-      } catch(err) {
-        sh "bin/irc-notify.sh --stage 'build " + env.BRANCH_NAME + "' --status 'failed'"
-        throw err
-      }
+        try {
+            sh 'bin/build.sh'
+        } catch(err) {
+            sh "bin/irc-notify.sh --stage 'build " + env.BRANCH_NAME + "' --status 'failed'"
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -2,33 +2,42 @@
...
 node {
     stage ('Prepare') {
-      checkout scm
+        checkout scm
     }
     if ( env.BRANCH_NAME == 'prod' ) {
-      buildSite()
-      syncS3('mdninteractive')
+        buildSite()
+        syncS3(is_mozmeao_pipeline() ? 'mdninteractive' : 'mdninteractive-b77d14bceaaa9ea4')
```
jwhitlock: I wish there was some way to conceal the nakedness of this constant (*emoji omitted*), but it only appears here, so 👍

PR #1140, opened 2018-09-11
Switch to https source for HTML audio example

jwhitlock:
The ``http://`` URL results in a mixed content warning when [embedded on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio).

Both links work:

* https://soundbible.com/mp3/Tyrannosaurus%20Rex%20Roar-SoundBible.com-807702404.mp3
* http://soundbible.com/mp3/Tyrannosaurus%20Rex%20Roar-SoundBible.com-807702404.mp3

### DIFF
--- a/live-examples/html-examples/image-and-multimedia/audio.html
+++ b/live-examples/html-examples/image-and-multimedia/audio.html
@@ -8,7 +8,7 @@
     <audio
         id="t-rex-roar"
         controls
-        src="http://soundbible.com/mp3/Tyrannosaurus%20Rex%20Roar-SoundBible.com-807702404.mp3">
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks @jwhitlock . It looks like this only changes the second link, not the one at line 11.

It's unfortunate that the local server doesn't show us this issue, and that we don't have tests for it, but that's a separate problem.

jwhitlock:
Thanks for the review @wbamberg, I've force-pushed to change both URLs.

I think CSP could help here. The example would break when it goes to production, so that it is fixed when someone embeds it. I'm not sure if CSP is applied for ``http://`` in local development, but if it is, we could find out even sooner.

PR #1141, opened 2018-09-11
Each radio button in the group should have distinct `value` attribute

Konrud:
If you omit the `value` attribute in the HTML tag, the submitted form data assigns the value "on" to the group of the radio buttons, it's not very helpful.


Konrud:
It's fine too, but I think that we should keep all three of them in order to persist the UI balance, 
as almost all examples out there for radio buttons have at least 3 of them.

PR #1142, opened 2018-09-12
Issue#1005 web api examples editor

schalkneethling:
@wbamberg This clean up the interactive examples editor and switching to using BoB for building the interactive examples. Work on the editor, and enabling building Web API type examples happens in the mdn/bob repo.


schalkneethling:
I will address the conflicts and update the PR

schalkneethling:
@wbamberg You will also notice that I am not proposing we merge this directly into master, but instead into a separate webapi-editor branch.

schalkneethling:
@hobinjk you should also have a look over these changes as it might impact your current work. Let me know your thoughts. Thanks!

wbamberg:
Thanks @schalkneethling ! This is a big change, and it would help to get a bit more detail on what the changes are, as I feel I'm lacking context a bit.

As I understand it, the main things here is separating out the source for the examples, which will stay in interactive-examples, from the code to build them and run the editor, which will be in bob. This PR doesn't provide a prototype for the Web API editor, so it should be functionally equivalent to what we had before.

## What's changed

In a bit more detail, I see the following changes in this PR:

* Remove editor code:
    * `_tests_*`
    * `css/*`
    * `js/*`
    * `lib/*`

* changes to `package.json` to build using `mdn-bob`

* changes to all `meta.json` files:
    * remove `baseTmpl`
    * change paths for `cssExampleSrc` and `exampleCode`

* one style change in the live examples source (named color to hex color) 

* `media/*` moved:
    * `media/fonts` to `live-examples/fonts`
    * the rest to `live-examples/media`, 

* Contribution docs moved from `./` to `/project-docs`, plus some doc updates (Markdown syntax changes, and mdn-bob mention added)

## Comments

I don't understand all the infra changes being made here, but I've tried running it and testing out a few examples, and everything seemed to work fine 👍 .

I think it's a little weird moving `media/*` under `live-examples`. Or at least, I think it might be good for the directory structure to distinguish between live example sources and the "assets" that they use, that might be shared across a number of examples. Does that make sense?

If we move the contributing docs we'd need to update the links to them from README. I also worry a bit that we'll break anywhere external that links to them (e.g. places like https://discourse.mozilla.org/t/html-interactive-examples-help-wanted/27966). Perhaps we should leave them in the root?

We'll also need to update the contributing docs to reflect the changes needed to meta.json and any relevant file moves (such as the `media/` move).

There are also some other infra cleanup bits. For example, `./eslintignore` references `./js/`. And some other things that I don't understand well enough to comment on.

What should we do with `all-contributorssrc`? Split it so code contributors are listed in mdn-bob and docs contributors in interactive-examples? Or not worry about it?

schalkneethling:
Thanks for adding all the details @wbamberg - This was an initial push so that I can:

1. Show where I am going with this
2. Get feedback in general
3. Ensure @hobinjk is aware of the changes as it may affect the work he is doing.

I have pushed some updates that adds a super simple WebAPI example. I will spend tomorrow adding the remainder, and also expand on what is changing and how the WebAPI examples might work.

NOTE: The latest here is still missing the console. That is part of the next iteration. This turned out to be a lot more complex than anticipated ;)

live-examples/js-examples/object/meta.json
```diff
@@ -1,148 +1,179 @@
...
-            "fileName": "object-values.html",
-            "title": "JavaScript Demo: Object.values()",
-            "type": "js"
-        }
-    }
-}
+{
+    "pages": {
+        "objectAssign": {
+            "baseTmpl": "tmpl/live-js-tmpl.html",
```
wbamberg: This commit seems to undo the changes that the last commit made to live-examples/js-examples/object/meta.json - so e.g. baseTmpl is reinstated here. Is that intentional?

schalkneethling: Nope, just need to rebase against master to get the latest changes ;)

schalkneethling:
> but maybe we can merge this and fix those as follow ups, if we want people to explore the new editor this week?

Yup, that sounds like a plan to me.

live-examples/js-examples/object/meta.json
schalkneethling: Hmmm, I did a diff between between the `meta.json` in this branch and the one on master and find no difference. Was `baseTmpl` removed here? Perhaps it is still in a PR that needs to be merged? @wbamberg

schalkneethling:
@wbamberg as mentioned via email

Learned a couple of things about ShadowDOM and JavaScript, and had to find a way to work around it for out examples. The tl;dr

Because the HTML is isolated and the CSS scoped to the shadow root, using something like `document.querySelector` will not return anything contained inside the shadowed area. i.e. ShadowDOM is not a document but a document fragment and so, document does not get mapped to the shadow root but, still refer to the parent document.

This means we need to first get at the shadow root, and from there we can then limit out query etc. to only the HTML inside the shadow tree. When I push the new code to the PR you will see what I mean in more detail but, I created a small util function, that is inject prior to the example JS, that simply return the shadow root. This does mean that, instead of for example doing:

```
document.querySelector('header');
```

It now is:

```
let documentRoot = getShadowRoot();
documentRoot.querySelector('header');
```

Also, to avoid accidentally overwriting variables that is in the scope of the entire page/window, we need to wrap our example code in a IIFE.

schalkneethling:
@wbamberg Another thing we need to think about is how to format the output when logging. When doing for example `console.log('header', rootDocument.querySelector('header'));`

I currently just output the `tagName` so:

```
> "header" HEADER
```

That is not extremely useful so, I also highlight the item that was selected as you will see in the current example. We could output the `outerHTML` but, that could fill up the console real quickly.

Would like your(and others') thoughts on this.

With regards to the console, it is still missing the following:

1. Scrolling for overflow content does not work for some odd reason
2. I reckon for the webapi examples it makes sense to set a taller height for the console
3. Missing a clear button to clear the console

live-examples/js-examples/object/meta.json
wbamberg: The first commit in this PR (https://github.com/mdn/interactive-examples/pull/1142/commits/95d94183e7eee55881489623463dd801b80ff0d4) removes `baseTmpl` from all meta.jsons, and adjusts some paths, as far as I can tell (e.g. https://github.com/mdn/interactive-examples/pull/1142/commits/95d94183e7eee55881489623463dd801b80ff0d4#diff-a19506ff399b51802f2918aefbf9d85c).

But the second commit (https://github.com/mdn/interactive-examples/pull/1142/commits/94dbbb194bdf72b5a185214665e38647d6050d90) seems to reinstate `baseTmpl`, and undo the path changes, for live-examples/js-examples/object/meta.json, presumably as a result of the catch-up integration for https://github.com/mdn/interactive-examples/pull/1127 . I assume that's not intentional, so this PR needs to re-remove `baseTmpl` and reapply the path changes for this file?

schalkneethling: You are indeed correct. Let me have a look at those. It was definitely a copy/paste manual merge thing. Thanks @wbamberg

schalkneethling:
@wbamberg Ok, this has been updated. I also found a bug in the JS examples build, so also pushed a new version of Bob.

wbamberg:
Thanks for the updates @schalkneethling ! 

> Another thing we need to think about is how to format the output when logging. When doing for example console.log('header', rootDocument.querySelector('header'));

Why not do the same as the JS console? In that case, for this example it would log:

```
> [object HTMLHeaderElement]
```

(or something). In this case I would probably make the example log `header.textContent` or something like that.

schalkneethling:
@wbamberg 

> Why not do the same as the JS console? In that case, for this example it would log:

Oh we could totally do that. I was thinking we might want that formatted a bit more. But having some best practices as to what to log, would be even better. Let me fix that in Bob.

wbamberg:
> When I push the new code to the PR you will see what I mean in more detail but, I created a small util function, that is inject prior to the example JS, that simply return the shadow root. 

I don't think I like having to have this in the example code. It obscures the point of the example and exposes the internal machinery of the editor. Is it possible somehow to have the editor take care of this, and let the example just use the global objects? 

> Also, to avoid accidentally overwriting variables that is in the scope of the entire page/window, we need to wrap our example code in a IIFE.

Similarly, could the editor take care of this? It seems weird to add the IIFE in the example, which should demonstrate the API as cleanly as possible.

wbamberg:
> Oh we could totally do that. I was thinking we might want that formatted a bit more. But having some best practices as to what to log, would be even better. Let me fix that in Bob.

Yes, it would be even better to be able to share the console code, which IIRC is a bit complicated.

schalkneethling:
@wbamberg Agree on both points.

1. One the first one, not sure really.... Let me ponder this for some time ;)
2. I can surely make this happen

schalkneethling:
> Yes, it would be even better to be able to share the console code, which IIRC is a bit complicated.

Yup, currently it is being shared.

schalkneethling:
Ok Bob updated to address the output. This now matches the JS editor. Also updated the current WebAPI example. Allowing `console.log` in eslint, just because all of the examples use it, and it gets annoying when the editor complains about it ;)

Todo:

1. Scrolling for overflow content does not work for some odd reason
2. I reckon for the webapi examples it makes sense to set a taller height for the console
3. Missing a clear button to clear the console
4. Let me ponder this for some time ;)
5. I can surely make this happen

schalkneethling:
@wbamberg Yup, this is fixed but I still need to update.

schalkneethling:
Updated to BoB 1.1.7 - Only items remaining are:

1. Let me ponder this for some time ;) aka

>I don't think I like having to have this in the example code. It obscures the point > of the example and exposes the internal machinery of the editor. Is it possible somehow to have the editor take care of this, and let the example just use the global objects?

2. I can surely make this happen, aka

> Similarly, could the editor take care of this? It seems weird to add the IIFE in the example, which should demonstrate the API as cleanly as possible.

schalkneethling:
@wbamberg The only item remaining here now is:

1. Let me ponder this for some time ;) aka

> I don't think I like having to have this in the example code. It obscures the point of the example and exposes the internal machinery of the editor. Is it possible somehow to have the editor take care of this, and let the example just use the global objects?

PR #1144, opened 2018-09-12
added path() example to basic shapes (CSS values)

estelle:
Adding path() as a basic shape.
It is not supported in clip-path. (but it is supported in offset-path in FF63, with motion path in Chrome)

specs: 
Definition of path()
https://drafts.csswg.org/css-shapes-2/#supported-basic-shapes

to be added to basic shapes at:
https://www.w3.org/TR/css-shapes/#basic-shape-functions

as per bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1246764

### DIFF
--- a/live-examples/css-examples/values-and-units/type-basic-shape.html
+++ b/live-examples/css-examples/values-and-units/type-basic-shape.html
@@ -35,7 +35,7 @@
         </button>
     </div>
     <div class="example-choice">
-        <pre><code class="language-css">clip-path: path('M 10,30 A 20,20 0,0,1 50,30 z');</code></pre>
+        <pre><code class="language-css">clip-path: path('M 20,60 A 40,40 0,0,1 100,60 z');</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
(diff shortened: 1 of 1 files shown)


wbamberg:
I'm not sure I understand the situation here. In https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path#Browser_compatibility it doesn't say that `clip-path: path(...)` is not supported, but it doesn't seem to work in this example and your comment:

> It is not supported in clip-path. 

obviously indicates that it isn't.

I don't think we should include options in the interactive examples that don't work yet (I think it's OK to include options that only work in some or most browsers, but not options that don't work in any, or that only work in one).

schalkneethling:
@wbamberg Same question here as in https://github.com/mdn/interactive-examples/pull/1145

schalkneethling:
@estelle Because of the following security vulnerability https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream that was revealed yesterday, we are reaching out to everyone that currently has an open PR on interactive examples.

schalkneethling:
I just tested and this works in Firefox, Safari, and Edge so I reckon is safe to merge now. @chrisdavidmills agree?

chrisdavidmills (approved):
Yup, let's get it done.

PR #1145, opened 2018-09-13
added path() value example, support in FF63 behind a flag.

estelle:
https://bugzilla.mozilla.org/show_bug.cgi?id=1246764

### DIFF
--- a/live-examples/css-examples/masking/clip-path.html
+++ b/live-examples/css-examples/masking/clip-path.html
@@ -19,6 +19,13 @@
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
     </div> 
+
+    <div class="example-choice">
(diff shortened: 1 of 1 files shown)


schalkneethling:
@wbamberg Should we merge and revisit later when support is better or, would you rather close?

wbamberg:
@schalkneethling is there a reason we can't leave this open? I appreciate that it would need rebasing after the big PR merge, but I don't think this one is ready to merge, but closing it seems a bit final.

schalkneethling:
@estelle Because of the following security vulnerability https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream that was revealed yesterday, we are reaching out to everyone that currently has an open PR on interactive examples.

schalkneethling:
Same here @chrisdavidmills, seems safe to merge now as it is supported in three browsers.

chrisdavidmills (approved):
Let's do it

PR #1170, opened 2018-10-03
Better naming for RegExp.exec example

Granjow:


wbamberg:
Thanks for your contribution!

The variable names in this example follow the style guide for JS examples (https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md) so we'd like to keep them.

PR #1189, opened 2018-10-18
Add "attribute-" prefix to HTML attribute examples.

### LINKED ISSUE #1065: Use disambiguating prefix for all HTML attribute examples
The example for the `style` attribute calls the resulting file "style-attr.html", to avoid a clash with the `<style>` element.

This is nice, but we should do this for all attribute examples, and use the form "attribute-style.html" to distinguish this group more clearly and to be consistent with what we have done for CSS functions: https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/transforms/meta.json#L13-L32

ShadowMitia:
Should close #1065.


ShadowMitia:
I've reused the css file as asked. Guess I was too busy renaming that I didn't notice I actually added that file!

ShadowMitia:
Glad I could help!

PR #1194, opened 2018-10-23
Fix #1115: position: sticky not working in Safari

### LINKED ISSUE #1115: CSS position example doesn't work in Safari

stephanmax:
This fixes #1115 by building on top of PR #1121 (using prefixed `-webkit-sticky` and making the boxes `display: block`).

The remaining issue was the fixed height of the inner container `.example-element-container` which Safari seems to use for calculating how long a sticky element is allowed to scroll (even if it is scrollable itself). Removing any height constraints on that `div` and instead making the outer section `overflow: auto` solved the problem (tested in Safari 12.0, FF, and Chrome).

I also recreated the blue boxes/placeholder behavior from before, removed the animation (was quite confusing for me, because you cannot animate `position` changes), and fixed the code formatting a bit.

### DIFF
--- a/live-examples/css-examples/positioned-layout/position.css
+++ b/live-examples/css-examples/positioned-layout/position.css
@@ -30,11 +30,3 @@
     border: 3px solid red;
     z-index: 1;
 }
-
-#example-element-placeholder {
-    background-color: transparent;
-    border: 3px dotted red;
-    position: absolute;
-    top: 56px;
-    left: 65px;
-}
--- a/live-examples/css-examples/positioned-layout/position.css
+++ b/live-examples/css-examples/positioned-layout/position.css
@@ -1,29 +1,16 @@
+.output section {
+    align-items: flex-start;
+    overflow: auto;
+}
 
 #example-element-container {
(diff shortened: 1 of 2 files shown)


stephanmax:
Great catch, thanks @wbamberg. I see two possibilities:
1. We put the placeholder into the first paragraph and position it relatively to it (which destroys a bit the semantics, but I'd say its okay for the sake of the visuals.)
2. We get rid of the placeholder.

What do you think?

PR #1196, opened 2018-10-23
Fix typo in String.prototype.normalize() example

vsemozhetbyt:


wbamberg:
Thanks @vsemozhetbyt but it looks like @ElForastero got to this one first :).

Issue #1138, opened 2018-09-08
Style Guide for Interactive Examples


atopal:
Interactive examples in my mind should allow you to instantly see how to use something to either remind you of the usage or give you a quick first insight. The rest of the page can handle exceptions and things to look out for. 

I was looking up the a[udio element ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio) and can’t even see the syntax without scrolling, that to me defeats the purpose of putting it on top of the page.

I'd suggest that we come up with a Style Guide that describes best practices and can be used to review submissions.

wbamberg:
-> https://github.com/mdn/interactive-examples/pull/1184

wbamberg:
Based on the draft guidelines and our discussions, I've been through the HTML elements and added an annotation (column C) saying whether they are OK as they are or should have some changes, and if so what the changes should be. The notes are a bit cryptic, but in general they are things like:

* simplifying and shortening where that seemed needed, to make the examples quicker to grasp and to fit without scrolling. To do this I've recommended things like:
    * removing extraneous elements that add extra complexity, where that doesn't seem needed
    * removing extra options (for example, when a <audio> example shows two <audio> elements, recommending we only show 1
    * removing whitespace where that seemed like the only way to get things to fit
* fixing some things that seem like accessibility issues (e.g. missing labels, use of title for explanatory text)
* improving formatting to make examples easier to scan
* using CC0 media so as not to have attributions in the examples
* setting a different height for the editor: either shorter for very short examples, or taller for longer ones that we can't reduce.

The trickiest I think were the table elements: the source there is always too long, but I think it's hard to get a table in the space, so I think we may need to use a taller editor for these.

atopal:
Thanks for doing this, Will!

wbamberg:
FYI - I estimated these updates at 2 days, laughably: it took me 6 days elapsed.

PR #1204, opened 2018-10-29
Issue 1138 content sectioning

wbamberg:
Part of https://github.com/mdn/interactive-examples/issues/1138: this part addressed the examples in the "content-sectioning" category. Generally minor updates to make the examples a bit simpler and more compact.

Note that I haven't changed indentation to 2 spaces as per the new style guidelines. I'd rather address this (and any other changes from that) separately.


wbamberg:
> I think the example should have something that makes this clear (the associated reference page should also be updated to make this clearer).

Yeah, I do agree. I think it is technically OK, but misleading.

>  Also, I don't see what was wrong with including the author's email address in the example? 

It makes the address wrap, which makes the source harder to read. In fact I can *just about* avoid wrapping (at 1000 pixels) with a shorter address.

What do you think of:

```
<p>Contact the author of this page:<p/>
<address>
  <a href="mailto:jim@example.com">jim@example.com</a><br/>
  <a href="tel:+311-555-2368">+311-555-2368</a><br/>
</address>
```

chrisdavidmills:
@wbamberg yup, that works for me!

wbamberg:
>     I'd suggest the following CSS would look better:

That looks like this on my computer:

screen shot 2018-10-30 at 1 44 26 pm

...which I think squashes the list items too much.

I think just changing:

```
footer p {
    flex: 2; /* from 1 in the original */
}
```

looks all right.

But then I got to thinking, this is for a footer, so it would be good to have something more like a real footer implementation. I found one at [this great website](https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers) - I'm a little sad about having an extra element in the example but think it's worth it. I simplified the footer contents to make up for it, so we end up with simpler CSS actually.

PR #1209, opened 2018-10-31
rename offset properties

rachelandrew:
The offset-* logical properties have been renamed. This makes that update to the interactive examples.

See the page for inset-block-end for details of the renaming: https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-end

### DIFF
--- a/live-examples/css-examples/logical-properties/inset-block-end.css
+++ b/live-examples/css-examples/logical-properties/inset-block-end.css
@@ -11,6 +11,7 @@
     background-color: yellow;
     border: 3px solid red;
     position: absolute;
-    offset-block-start: 50px;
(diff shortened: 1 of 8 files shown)


rachelandrew:
I am very jetlagged but I'm not sure what you are seeing that is different than what I am. I have four files and four related CSS files, all of which seem to be showing the right thing.

rachelandrew:
it looked a bit weird in GitHib, like it had confused which file had become what. I had to trash and recreate my fork as it was in a mess so I don't know if that had anything to do with it. Seems as if once a week I managed to cause some git related calamity somewhere!

chrisdavidmills:
No worries ... I am a recovering git destroyer too ;-)

PR #1216, opened 2018-11-01
Update image and multimedia examples

wbamberg:
Part of #1138: this part addressed the examples in the "images and multimedia" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

* [`audio`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio): big changes: only one element, simpler markup, made the mp3 local (it's [public domain](http://soundbible.com/1748-Tyrannosaurus-Rex-Roar.html))
* [`track`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track): change to use a [public domain clip](https://archive.org/details/his_girl_friday), and more plausible captions
* [`video`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video): change to use a public domain clip, and add a `track` element to describe it for a11y.

I wasn't sure what to do about a11y for the video clip, and found the [MDN page not terribly helpful](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#Accessibility_concerns). It says "Videos should provide both captions and transcripts" - but nowhere does it say how to provide a transcript, and of course it has no audio content so if captions "allow people who are experiencing hearing loss to understand a video's audio content as the video is being played" - then they are not appropriate here.

It feels that I want something like a label, that describes the video, so in the end I went with a `track` with a `kind` of `descriptions`.


wbamberg:
Thanks for the review, all good suggestions. I've made some updates which (I hope) cover it.

PR #1217, opened 2018-11-01
Fix logical properties

wbamberg:
https://github.com/mdn/interactive-examples/pull/1209 updated the names of some of the CSS logical properties but didn't update meta.json, which causes `npm run build` and `npm run start` to fail.

This hopefully fixes things.


chrisdavidmills:
Looks good @wbamberg , thanks for catching that.

PR #1218, opened 2018-11-01
Update inline text semantics examples

wbamberg:
Part of #1138: this part addressed the examples in the "inline text semantics" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

Note that this also include https://github.com/mdn/interactive-examples/pull/1217, since I needed that to test the examples.

Mostly these are quite small updates: simplifying the markup, changing the formatting for readability. I've also made most of them use the shorter editor since they tend to be very short examples.

### DIFF
--- a/live-examples/css-examples/logical-properties/meta.json
+++ b/live-examples/css-examples/logical-properties/meta.json
@@ -182,32 +182,32 @@
             "title": "CSS Demo: min-inline-size",
             "type": "css"
         },
-        "offset-block-end": {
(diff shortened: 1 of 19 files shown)


wbamberg:
Thanks! Updated.

wbamberg:
\o/ thanks for all the reviews @chrisdavidmills !

PR #1219, opened 2018-11-02
Update input examples

wbamberg:
Part of #1138: this part addressed the examples in the "input" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

Note that this also include #1217, since I needed that to test the examples.

This is quite a big change. I won't go through it in detail, but changes are:

* remove the generic `input` example, which had lots of different types. We'll just use some representative one on the main input page, probably `text`.

* lots of changes to try to fit the examples into the space available without needing to scroll. This is mostly removing extra markup. Sometimes I've gone with just a "bare" example with only label and input, sometimes (where it fit, or where it seemed needed (e.g. `reset`)) I've included other elements. The old examples used `fieldset` and `legend` pretty consistently: I like these, but they use a lot of space and add a lot of markup to scan. So I've mostly but not always removed this.

A few of the examples *still* don't fit, and I'd be happy to discuss suggestions for fixing this, but at some point I think you just have to accept it. I couldn't quite let go of Huey Dewey and Louie.

I'd also be interested in knowing if you think the "bare" examples are in general better, and if you think I should rewrite some of the other examples to be "bare" as well.

* a few changes to try to improve accessibility, including removing critical info from placeholders and including it in labels instead.


wbamberg:
> So for the parent <input> page, you are intending to just use the <input type="text"> one, or something equally simple?

Yes, that was the plan.

Other comments all look totally doable. I omitted `checked` in `checkbox` and `radio` to save space, but on reflection `checkbox` will still fit and `radio` doesn't fit anyway, so it will be fine ti add it.

wbamberg:
Oh and I need to keep the `form` in the `reset` example, or the reset button doesn't work.

PR #1220, opened 2018-11-02
correct the position based index at which to start copying elements from in the comment

ichthub:


ichthub:
Hi @wbamberg, yeah the wording is little confusing because the comment refers to the index as position at the beginning of the sentence and ends with using the same word "position"  to refer to value. So I tried to go with interpreting it as index of array rather than the value.

schalkneethling:
@ichthub Because of the following security vulnerability https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream that was revealed yesterday, we are reaching out to everyone that currently has an open PR on interactive examples.

PR #1222, opened 2018-11-02
Update table content examples

wbamberg:
Part of #1138: this part addressed the examples in the "table content" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

The problem with this category is, the examples look fine but it's not really possible to show these elements outside the context of a table. This has a couple of consequences:

* it's hard to see the target of an example right away, because there's so much markup
* it's very hard to write markup for a table that fits into 14 lines, which is all we get for the "standard" height editor.

I don't like using the "taller" editor (23 lines) but I don't really see an alternative here.

So what I've done is:
* trimmed the examples to fit inside 23 lines, and to simplify them as much as I could
* set them to "taller" in meta.json

I'm open to improvement suggestions, but don't really have a better idea here.

### DIFF
--- a/live-examples/html-examples/table-content/caption.html
+++ b/live-examples/html-examples/table-content/caption.html
@@ -6,6 +6,11 @@
         <th scope="col" class="skeletor">Skeletor</th>
     </tr>
     <tr>
+        <th scope="row">Role</th>
+        <td>Hero</td>
+        <td>Villain</td>
+    </tr>
+    <tr>
         <th scope="row">Weapon</th>
         <td>Power Sword</td>
         <td>Havoc Staff</td>
--- a/live-examples/html-examples/table-content/caption.html
+++ b/live-examples/html-examples/table-content/caption.html
@@ -1,14 +1,9 @@
 <table>
     <caption>He-Man and Skeletor facts</caption>
     <tr>
-        <td>&nbsp;</td>
-        <th scope="col"><span class="heman">He-Man</span></th>
-        <th scope="col"><span class="skeletor">Skeletor</span></th>
-    </tr>
-    <tr>
-        <th scope="row">Role</th>
-        <td>Hero</td>
-        <td>Villain</td>
+        <td> </td>
+        <th scope="col" class="heman">He-Man</th>
+        <th scope="col" class="skeletor">Skeletor</th>
     </tr>
     <tr>
         <th scope="row">Weapon</th>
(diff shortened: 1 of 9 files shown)


wbamberg:
> How about setting the table font-size to 0.6rem? That makes the table fit better in the output

What about:
```
td,
th {
    border: 1px solid rgb(190, 190, 190);
    padding: 7px 5px;  /* instead of 7px 15px */
}
```
This fits nicely and looks fine,to my eye:

screen shot 2018-11-05 at 8 25 55 am

> To stop the HTML source code wrapping, how about writing the header code like this:

While I think this is better than what we have now, I thought removing the `span` entirely made the markup simpler and avoided wrapping at sensible widths. Is there some reason the extra `span` is important, that I'm missing?

chrisdavidmills:
I agree with the CSS updates you suggest — this is much better and looks fine.

I don't think we need to extra `<span>`, no — just putting the classes on the `<th>` elements would work fine. something like this?

```
<tr>
    <td> </td>
    <th scope="col" class="he">He-Man</th>
    <th scope="col" class="sk">Skeletor</th>
</tr>
```

Making the class names shorter helps further.

wbamberg:
> I agree with the CSS updates you suggest — this is much better and looks fine.

OK, I just pushed this change.

> I don't think we need to extra <span>, no — just putting the classes on the <th> elements would work fine.

That was the [version in the original PR](https://github.com/mdn/interactive-examples/pull/1222/commits/d1f09c4122362796ff462e97fe4ab5475a24fe03#diff-8c7c731a62f1849bfbae84cb569c8ba3R5), so I've just kept it.

> Making the class names shorter helps further.

I didn't do this - just removing the extra `span` elements, without shortening class names, makes the example avoid wrapping at 850px. Shortening the class names would help more of course, but I think "he" and "sk" are a bit cryptic, so on balance I don't think it is worth doing.

My original PR here removed the first table row - "Role" - just to shorten the example source. I think the example (and more importantly, the joke) still work well without that row, but let me know if you think I should reinstate it.

wbamberg:
> I'd like it reinstated please. I think the joke works better with ;-)

Yeah, it does, really. Reinstated.

wbamberg:
Thanks @chrisdavidmills !

PR #1223, opened 2018-11-03
Update text content examples

wbamberg:
Part of #1138: this part addressed the examples in the "text content" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

Changes to simplify markup mostly. I also switched to a CC0 image for one of the examples, and switched to using shorter editors where that was possible.

I also changed a couple of dl/dt ones whose topic was web documentation, to make it something else.


wbamberg:
> <div> — this use case would be a much better semantic fit for `figure`/`figcaption`. Would it be better on this one to show just maybe an information box div with a couple of small paragraphs in it, set to about 300px wide and then centered in the output area with margin: 0 auto? 

Hmm, this wasn't what [the original review](https://github.com/mdn/interactive-examples/pull/980#pullrequestreview-130339464) had for this example.

I don't especially mind rewriting this example but I'm a bit reluctant to when we've gone round it before.

I don't actually think this is a good fit for `figcaption` - it feels to me that `<figure>`/`<figcaption>` is suitable where the image is primary, and the caption is an addendum to the image. In this case the warning text is primary and the image is an illustration.

chrisdavidmills:
> Hmm, this wasn't what the original review had for this example.

This is a different day, you know ;-)

> I don't actually think this is a good fit for figcaption - it feels to me that `<figure>`/`<figcaption>` is suitable where the image is primary, and the caption is an addendum to the image. In this case the warning text is primary and the image is an illustration.

This is a good point. OK, let's leave it as is.

wbamberg:
> `<figcaption>/<figure>` — lovely example; shame it doesn't fit into the example pane.

It's a good point, I didn't try at wider screen sizes. What do you think of:

```
figure {
    border: thin #c0c0c0 solid;
    display: flex;
    flex-flow: column;
    padding: 5px;
    max-width: 220px;
    margin: auto;
}

img {
    max-width: 220px;
    max-height: 150px;
}

figcaption {
    background-color: #222;
    color: #fff;
    font: italic smaller sans-serif;
    padding: 3px;
    text-align: center;
}
```
?

chrisdavidmills:
Yup, that looks better for the `<figure>`/`figcaption` example

wbamberg:
Updated `figure`,`figcaption`, `ol`, and `main`. I made `figure` and `figcaption` share a CSS file.

PR #1224, opened 2018-11-03
Update global attributes

wbamberg:
Part of #1138: this part addressed the examples in the "global attributes" category, following the notes in column C of https://docs.google.com/spreadsheets/d/1EC23mRNF7i-Bd29A4wEcFV_iiThL1P-NBtH4lzme8eM/edit#gid=0.

These are generally minor changes to simplify the markup a bit, format it so it's more readable, and use the shorter editor where it's possible.

This is the last one \o/. Thanks for all the reviews!


wbamberg:
> This doesn't work because most of the CSS is missing. 

Ah it looks like https://github.com/mdn/interactive-examples/pull/1189 introduced this error. I've added your CSS back for `data` and also reinstated the original CSS for `dir`.

Yes, I agree about `title`. It's not ideal but it's OK for now I think.

PR #1236, opened 2018-11-18
Provide valid and more semantic HTML with <audio>

karlstolley:
The old example used <label>, which while permitted in this context, the
value of `for` must be a labelable form element. In HTML 5.2, that list
is at https://www.w3.org/TR/html/sec-forms.html#labelable-element (W3C)
https://html.spec.whatwg.org/multipage/forms.html#category-label (TLS).
Because <audio> is not in that list, the previous example triggered
validation errors with both the W3C and WHATWG validators.

I have rewritten the example with a more accessible, semantic solution
using the <figure> element. The accompanying <figcaption> serves a more
semantic role than <label> in this context.

I've also updated the CSS for example, although it is kind of
superfluous.

### DIFF
--- a/live-examples/html-examples/image-and-multimedia/audio.html
+++ b/live-examples/html-examples/image-and-multimedia/audio.html
@@ -1,8 +1,9 @@
-<label for="t-rex-roar">Listen to the T-Rex:</label>
-
-<audio
-    id="t-rex-roar"
-    controls
-    src="/media/examples/t-rex-roar.mp3">
-    Your browser does not support the <code>audio</code> element.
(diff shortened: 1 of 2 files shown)


wbamberg:
Thank you, this looks great (and thanks also for the nice PR description :).

Just one thing, do you think setting `margin: 0` will make it look better in the output pane?

```
figure {
    margin: 0;
}
```

(I agree that the existing CSS is not needed).

karlstolley:
@wbamberg absolutely. Let me make that fix and push again.

PR #1245, opened 2018-11-28
Adding  Intl.RelativeDateFormat example

romulocintra:
Adding RelativeDateFormat example , according to  https://github.com/tc39/proposal-intl-relative-time


romulocintra:
@wbamberg  completely agree will do changes ... and align it with actual docs

PR #1257, opened 2018-12-06
Use foo-bar convention

elharony:
Generic the code by replacing the human names with `foo-bar` convention instead


elharony:
Thanks for merging it @wbamberg ! :+1:

PR #1266, opened 2018-12-11
Add example for window.alert

kevinsimper:
There is no interactive examples for window.alert which I found really sad because that is some of the most fun part to learn at the start about JavaScript 😄 

https://developer.mozilla.org/en-US/docs/Web/API/Window/alert

So I made this PR because I couldn't find any examples about `alert` either. So now I am trying to add it all the way.

Btw. couldn't docs inject a editor into for example the alert wiki page and make the static examples interactive? Like there is already a example, but you just can interact with it on the page: https://developer.mozilla.org/en-US/docs/Web/API/Window/alert

Is there any reason for that?

### DIFF
--- a/live-examples/js-examples/window/meta.json
+++ b/live-examples/js-examples/window/meta.json
@@ -1,0 +1,10 @@
+{
+    "pages": {
+        "stringCharAt": {
+            "exampleCode": "./live-examples/js-examples/window/window-alert.html",
+            "fileName": "window-alert.html",
+            "title": "JavaScript Demo: window.alert()",
+            "type": "js"
+        }
+    }
+}
--- a/live-examples/js-examples/window/window-alert.html
+++ b/live-examples/js-examples/window/window-alert.html
@@ -1,0 +1,4 @@
+<pre>
+<code id="static-js">alert('Hello');
+</code>
+</pre>


kevinsimper:
> But `alert` isn't strictly a JS API, it's a DOM API, so it's not included in the JS examples. We are working on adding DOM examples but they are a bit tricker as they generally want a full Web page environment (meaning that the example will usually want to include HTML and sometimes CSS sources as well as JS).

I am not sure I understand, I know the alert is triggered on window, but the code example I made works great.

https://imgur.com/a/rv9yFCG

And it doesn't need a full Web page environment and that was exactly why I wanted to add it, it is a very simple code snippet and it would improve the MDN docs for alert a lot that today only contains a non-interactive code and a very old windows screenshot 😄 

I would argue that putting it into the a folder called `window` and it only contains javascript makes for a great start and a easy way to move it if there exist a better place later.

But if this doesn't work, how can we make an interactive alert example?

kevinsimper:
Sorry, forgot to mention @wbamberg :)

kevinsimper:
@wbamberg I really appreciate the responses and respect it. I will look for a issue or create one for the editor that can handle both 👍

PR #1273, opened 2018-12-19
demonstration of mutation side-effects

dehuszar:
As the first example that the visitor will see, it's important to emphasize that the target argument will get mutated.  Those details get described a little later in the document, but it's easily missed at a glance.  This update to the example intends to demonstrate that.  Please advise if better language could be used.


wbamberg:
Thanks for the PR and sorry to be so slow to review. I don't think we need to go into this in the first example on the page. The interactive examples are supposed to be a quick demo of the basic behavior, and we can note extra details in more explanation and examples further on.

I don't like saying no to PRs, but I do think in this case we don't need this change in the interactive example. If you wanted to, perhaps it would be worth adding an example showing the mutation side-effect in the "Examples" section instead, where we have the space to highlight many different aspects of this API.

dehuszar:
Sorry for not getting back to this sooner, I'll move to a different example

wbamberg:
@dehuszar , I have https://github.com/mdn/sprints/issues/961, cross-filed from https://bugzilla.mozilla.org/show_bug.cgi?id=1514877, to add a non-interactive example demonstrating mutation to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Examples.

Let me know if you want to do this, if not I will. Thanks!

wbamberg:
@dehuszar , @schalkneethling , I've thought about this some more and think maybe we should after all have this in the interactive example, but maybe in a simpler form.

What do you think about:

```js
const target = { a: 1, b: 2 };
const source = { b: 4, c: 5 };

const returnedTarget = Object.assign(target, source);

console.log(target);
// expected output: Object { a: 1, b: 4, c: 5 }

console.log(returnedTarget);
// expected output: Object { a: 1, b: 4, c: 5 }
```
?

dehuszar:
That works for me!

dehuszar:
Apologies, I didn't see your previous comment till just now

wbamberg:
Thanks @dehuszar for your contribution, I think this is a better example for it!

PR #1235, opened 2018-11-10
Convert to string in toString override

Jonathan-Ironman:
The old example did not necessarily return a string.

PR #1279, opened 2019-01-08
Convert to string in toString override

Jonathan-Ironman:
@schalkneethling, @wbamberg I re-created PR #1235


wbamberg:
@Jonathan-Ironman , thanks for this. There's still this comment, from https://github.com/mdn/interactive-examples/pull/1235#pullrequestreview-180496483:

> I think on balance I'd prefer not to use class, and just to change the toString implementation to address the problem you've identified. But I don't feel very strongly about it, and maybe there a reason to use it here that I am missing.

...if only to fit this example inside 13 lines, so the user doesn't need to scroll to see it.

Jonathan-Ironman:
@wbamberg I've reverted the conversion to class and simply fixed the old example as you suggested.

Issue #1263, opened 2018-12-08
Allow targetting of a class rather than an ID in CSS examples


rachelandrew:
I have an issue where to make a CSS example for the CSS property `scroll-snap-align` behave consistently I need to be able to apply the selected value to all child nodes. 

Currently I believe I can only target one element, by giving it the ID `example-element`. I actually need to target that element and all of the siblings. Would it be possible to make it so that a class can be targeted instead as this would mean I could apply the property to all children. Otherwise I think the solution will be to use some JavaScript to do so.

In the example below, I want to be able to apply `scroll-snap-align` to all of the children and not just the one with the ID.

```
<section id="example-choice-list" class="example-choice-list large" data-property="scroll-snap-align">
    <div class="example-choice" initial-choice="true">
        <pre><code class="language-css">scroll-snap-align: none;</code></pre>
        <button type="button" class="copy hidden" aria-hidden="true">
            <span class="visually-hidden">Copy to Clipboard</span>
        </button>
    </div>
    <div class="example-choice">
        <pre><code class="language-css">scroll-snap-align: start;</code></pre>
        <button type="button" class="copy hidden" aria-hidden="true">
            <span class="visually-hidden">Copy to Clipboard</span>
        </button>
    </div>

    <div class="example-choice">
        <pre><code class="language-css">scroll-snap-align: end;</code></pre>
        <button type="button" class="copy hidden" aria-hidden="true">
            <span class="visually-hidden">Copy to Clipboard</span>
        </button>
    </div>

    <div class="example-choice">
        <pre><code class="language-css">scroll-snap-align: center;</code></pre>
        <button type="button" class="copy hidden" aria-hidden="true">
            <span class="visually-hidden">Copy to Clipboard</span>
        </button>
    </div>
</section>

<div id="output" class="output large hidden">
    <section id="default-example" class="default-example">
        <div class="scroller">
            <div>1</div>
            <div id="example-element">2</div>
            <div>3</div>
        </div>
    </section>
</div>
```

wbamberg:
I can't think of any reason not to do this, unless it's terribly complicated to implement.

bsmth:
This is possible following https://github.com/mdn/rari/pull/84 so this is solved. See https://github.com/orgs/mdn/discussions/782 for more information. Thank you :)

PR #1280, opened 2019-01-09
New PR for scroll snap examples

rachelandrew:
This is a new PR to try and fix the unresolvable mess I created in my last one!

Old PR (which I will close) https://github.com/mdn/interactive-examples/pull/1260

I have added some text to each example which I help demonstrates what people are supposed to do.

I have removed `scroll-snap-align` as that shouldn't be in there until we decide what to do about this issue: https://github.com/mdn/interactive-examples/issues/1263

### DIFF
--- a/live-examples/css-examples/scroll-snap/meta.json
+++ b/live-examples/css-examples/scroll-snap/meta.json
@@ -1,6 +1,153 @@
 {
     "pages": {
-        "scrollSnap": {
+        "scrollMargin": {
+            "cssExampleSrc": "./live-examples/css-examples/scroll-snap/scroll-margin.css",
+            "exampleCode": "./live-examples/css-examples/scroll-snap/scroll-margin.html",
+            "fileName": "scroll-margin.html",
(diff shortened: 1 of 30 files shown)


live-examples/css-examples/scroll-snap/meta.json
```diff
@@ -1,11 +1,158 @@
...
+            "type": "css"
+        },
+        "scrollMarginInlineStart": {
+            "cssExampleSrc": "./live-examples/css-examples/scroll-snap/scroll-margin.css",
+            "exampleCode": "./live-examples/css-examples/scroll-snap/scroll-margin-inline-start.html",
+            "fileName": "scroll-margin-inline-start.html",
+            "title": "CSS Demo: scroll-margin-inline-start",
+            "type": "css"
+        },
+        "scrollMarginInlineRight": {
```
wbamberg: I think this should be `scrollMarginRight`, shouldn't it?

rachelandrew: oops yes! Fixed.

PR #1284, opened 2019-01-10
added anywhere value to overflow-wrap property

estelle:
added 'anywhere' value to bcd for overflow-wrap property

based on https://bugzilla.mozilla.org/show_bug.cgi?id=1505786 (FF 65 supports 'anywhere') 

spec: https://drafts.csswg.org/css-text-3/#propdef-overflow-wrap

### DIFF
--- a/live-examples/css-examples/text/overflow-wrap.css
+++ b/live-examples/css-examples/text/overflow-wrap.css
@@ -1,6 +1,13 @@
-#example-element {
-    background-color: gold;
+.default-example {
+    display: flex;
+}
+
+.default-example > div {
+    background-color: lemonchiffon;
     padding: .75em;
     width: 200px;
-    height: 200px;
+}
+
+#example-element {
+    background-color: gold;
 }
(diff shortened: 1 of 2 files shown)


wbamberg:
As it stands this change doesn't show the different between `break-word` and `anywhere`: is it possible to change the example so it does? I see that the live example further down the page does this by adding extra elements in a flex layout, so you can see how the width calculation is affected (if I understand it properly). Is it possible to do something like that here?

e.g.

screen shot 2019-01-15 at 4 32 48 pm

and

screen shot 2019-01-15 at 4 33 03 pm

estelle:
@wbamberg That's an awesome example. Do you want to check it in?

wbamberg:
>  Do you want to check it in?

OK, let's use it.

CShepartd:
Using flex is unproperly with `overflow-wrap: anywhere`. For better understand this rule see [this example](https://jsfiddle.net/ofgd83um/80/)

PR #1289, opened 2019-01-15
text-transform: full-size-kana value

estelle:
added kana example and full-size-kana value. Works in FF only.
https://bugzilla.mozilla.org/show_bug.cgi?id=1498148
https://drafts.csswg.org/css-text-3/#propdef-text-transform
https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

### DIFF
--- a/live-examples/css-examples/text/text-transform.html
+++ b/live-examples/css-examples/text/text-transform.html
@@ -35,11 +35,20 @@
 </div>
 </section>
 
+<div class="example-choice">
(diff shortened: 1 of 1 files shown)


live-examples/css-examples/text/text-transform.html
```diff
@@ -35,11 +35,20 @@
 </div>
 </section>
```
wbamberg: This extra closing `</section>` is breaking the example.

PR #1292, opened 2019-01-17
Replace H1 with H2 in multiple sections example

OlivierJaquemet:
Having several H1 on the same page is not recommended. Citing MDN itself https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements
"You should consider avoiding using < h1 > more than once on a page. See Defining sections in Using HTML sections and outlines for more information."

Update this multiple section example to replace use of H1 elements with H2, as it does not reflect the best practice to have several H1.

### DIFF
--- a/live-examples/html-examples/content-sectioning/section.html
+++ b/live-examples/html-examples/content-sectioning/section.html
@@ -1,9 +1,9 @@
 <section>
-    <h1>Introduction</h1>
+    <h2>Introduction</h2>
     <p>People have been catching fish for food since before recorded history...</p>
 </section>
 
 <section>
-    <h1>Equipment</h1>
+    <h2>Equipment</h2>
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for the PR!

That doc is a little hard for me to understand, but as I read it, it's fine to have multiple h1 headings when each is in its own section:

> Each section can have its own heading hierarchy. Therefore, even a nested section can have an \<h1\>.

OlivierJaquemet:
@wbamberg thanks for your explanation. Bad understanding on my side.
Those spec are sometimes so hard to interpret, it's quite confusing! Sorry, I was trying to help, but did not ;) Next time maybe!

wbamberg:
No worries, I appreciate your engagement with the project 👍

PR #1298, opened 2019-01-22
Added for await... of example

jahzielv:
Contributing this example as part of a documentation effort for async iteration, which can be seen here: https://github.com/tc39/proposal-async-iteration/issues/132.


jahzielv:
Is there anything I can do to move this PR forward?

wbamberg:
Thanks for your contribution @jahzielv ! 

I'm going to ask Florian to take a look at when he digs into the async interation doc updates. My initial reaction though is that it is too complex for an interactive example, although I don't know this stuff well enough to know whether it's possible to simplify it.

Do you have to write your own async generator, or is it possible to use one that's already implemented in an API?

I'd also prefer to have examples that are self-contained and don't depend on any other sites, although I could see that this might conflict with making the example simpler.

jahzielv:
Hi @wbamberg ! Thanks for welcoming me and for your feedback; I'm excited to be contributing to this project!

I definitely agree that it's a little longer than most interactive examples, and I looked for a way to make it it more concise, but couldn't find any - all examples I can find online use custom generators. I'm not sure that there's any web APIs that currently implement async iterables, but I could be totally wrong about that. 

I also see your point about depending on external sites; it's not great from a maintainability POV since you don't control what's going on elsewhere. I can't think of a way to avoid it though, since async iteration seems to be aimed mostly at streaming data, and like you said, avoiding it would add more code. I'm down to make any changes you and @Elchi3 suggest to improve the example!

PR #1299, opened 2019-01-23
Add example for globalThis

goodforenergy:
Adds an interactive example for globalThis.


wbamberg:
Thanks @goodforenergy ! I've asked Florian to take a look at it.

PR #1303, opened 2019-01-28
fix: change Object.entries to make it clearer

esanzgar:
1. sort method had no effect as it was incorrectly subtracting arrays
2. calling Object.entries on an array was a bit contrived


wbamberg:
This PR seems to be stalled, so I'm closing it in favour of https://github.com/mdn/interactive-examples/pull/1350.

PR #1311, opened 2019-02-05
Revert "added anywhere value to overflow-wrap property"

CShepartd:
Reverts mdn/interactive-examples#1284

Nesting `overflow-wrap: anywhere` in `flex` element is unproperly. For better understand this rule see [this example](https://jsfiddle.net/ofgd83um/80/)

### DIFF
--- a/live-examples/css-examples/text/overflow-wrap.css
+++ b/live-examples/css-examples/text/overflow-wrap.css
@@ -1,13 +1,6 @@
-.default-example {
-    display: flex;
-}
-
-.default-example > div {
-    background-color: lemonchiffon;
-    padding: .75em;
-    width: 200px;
-}
-
 #example-element {
     background-color: gold;
+    padding: .75em;
+    width: 200px;
+    height: 200px;
 }
--- a/live-examples/css-examples/text/overflow-wrap.html
+++ b/live-examples/css-examples/text/overflow-wrap.html
@@ -7,14 +7,8 @@
         </button>
     </div>
 
-    <div class="example-choice">
+    <div class="example-choice" initial-choice="true">
         <pre><code class="language-css">overflow-wrap: break-word;</code></pre>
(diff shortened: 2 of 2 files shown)


wbamberg:
@estelle , would you mind looking at this?

estelle:
LGTM.

wbamberg:
Sorry @estelle , I should have been clearer. This PR reverts the example which we added in https://github.com/mdn/interactive-examples/pull/1284, on the grounds that `overflow-wrap: anywhere` should not be used with flex.

The first iteration of that PR didn't show the difference between `anywhere` and `break-word`, so I suggested using flex, although I got the idea from the other example in the page: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Example.

I'm not familiar enough with this property to know whether it's true that `overflow-wrap: anywhere` should not be used with flex, but I've got no reason to doubt @CShepartd .

So I think just merging this PR is not enough: if we do want an example for `overflow-wrap: anywhere`, then we should add one that does not use flex, but that still illustrates the difference between `anywhere` and the other values. We should also presumably fix the example in the page, that also uses flex.

Can you take this on, since you originally updated the page to add `anywhere`?

estelle:
OK. i need to research the flex thing. I thought you just wanted to revert the example at the top. As it also doesn't yet work in all browsers and you can't really tell the difference, i am fine with reverting it. I need to research the 'don't use with flex' because i missed seeing that in the spec.

schalkneethling:
> @schalkneethling , this PR is showing a snyk failure, although I don't see anything in the actual change to reflect that. What do you think?

It is one of our dependencies that needs to be updated. Not introduced here. This is fine to merge.

schalkneethling:
@wbamberg Probably a good idea to rebase this against master as I have updated a couple of the dependecies.

PR #1312, opened 2019-02-06
Changed to let instead of var due to Hoisting

WouterNoordhof:


WouterNoordhof:
I initially noticed that it was `var`, because in the example of the [`for...of loop`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), `let` does get used instead of `var`. 
To be honest i didn't look at the other variables/bindings in the example, but you are correct, if `let` would be used for `property1`, then `const` should also be used for `object1`. (And maybe `let` should also be used for `string1`)

wbamberg:
> I initially noticed that it was `var`, because in the example of the [`for...of loop`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), `let` does get used instead of `var`.

That's because `for...of` is ES6-ish, so the example gets to use other ES6-ish language features, as per https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6.

If we're strictly following that guideline then we should just use `var` in this example, and that would be consistent. As it goes I'm happy to use modern language features anyway, and we could reconsider that guideline, especially now that the overall rules for MDN examples allow/encourage them (https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Code_guidelines/JavaScript#Use_ES6_features).

But yes, in that case `object1` should use `const` and `string1` should use `let`.

wbamberg:
I'm closing this due to lack of activity.

PR #1317, opened 2019-02-18
Remove one element using splice

Rotimi-Best:
It wasn't clear in the docs that I could use splice to just remove an element and not only replace. So to show this I added a simple example.

### DIFF
--- a/live-examples/js-examples/array/array-splice.html
+++ b/live-examples/js-examples/array/array-splice.html
@@ -1,5 +1,10 @@
 <pre>
 <code id="static-js">var months = ['Jan', 'March', 'April', 'June'];
+months.splice(1, 1);
+// starting from index 1 remove 1 element or remove the second element
+console.log(months);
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your contribution @Rotimi-Best , and sorry to be slow reviewing.

I think this change makes sense and works well. But with these interactive examples our intention is to show simple basic usage of the item, to help people get a quick grasp of it, not to show all the things you can do with it. There are just too many, and it would make the interactive examples too long.

For more exhaustive/in-depth examples there's the "Examples" section further down the page, and in this page actually there's a great list of examples, including one of removing a single element: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#Remove_1_element_from_index_3.

So although I don't like to say no, I don't think we need this addition in the interactive example. But I do appreciate your contribution to MDN.

Rotimi-Best:
oh great, such example already exists, I just didn't notice it. Thanks

PR #1320, opened 2019-02-19
Repair <label> html example

HolgerJeromin:
The whole point in the example was to have one checkbox with and one without a label.

This was reverted in the cleanup PR #1210 and is now not in sync with the main text in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

@wbamberg @chrisdavidmills

### DIFF
--- a/live-examples/html-examples/forms/label.html
+++ b/live-examples/html-examples/forms/label.html
@@ -1,5 +1,5 @@
 <div class="preference">
-    <label for="cheese">Do you like cheese?</label>
+    <span>Do you like cheese?</span>
     <input type="checkbox" name="cheese" id="cheese">
 </div>
 


wbamberg:
I'm not sure how it is not in sync with the text?

> The HTML <label> element represents a caption for an item in a user interface.
> 
> In the examples above you can see two form input examples — both with a corresponding `<label>`.

HolgerJeromin:
@wbamberg 
Sorry, i only checked the German text, which is only partly translated:
https://developer.mozilla.org/de/docs/Web/HTML/Element/label
which is an old version of the original. Quite a mess 8-/

> In the examples above you can see two form input examples — the first one has no <label>, while the second one does have a <label>. The second example has some major advantages:

An [edit](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label$revision/1434231) from Nov 9, 2018 was the first english version which states that both rows are the same. Your PR was before that. The editor probably saw the HTML code not matching the English text. 

The fr and ja version appears (deepl.com and google translate :) to be updated, too. The other languages are not using the examples from here.

So even if i like the example with different meaning the translation has been moved forward, so the best solution would be to update the outdated German edition.

I will take care of that!

wbamberg:
Thank you @HolgerJeromin ! Indeed that explains it. Many of our non-en-US locales are very out of date unfortunately, and (for this reason I guess) I think many people prefer to use the en-US version even when it's not their native language. but it's definitely good to correct misleading text like this.

I'm on the fence about whether I like the example comparing no label to label. The history of this is that we added a bunch of HTML examples, then decided that many of them were too complex, and decided to simplify them. This was in general the right decision I think, and I like the idea that the interactive examples should be as simple as possible to grasp, to show the basic usage and syntax of the thing, and any nuances can be left to the rest of the page to explain.

PR #1321, opened 2019-02-21
Add example of string replace without regex

EddieOne:
Otherwise, it's easy to miss this use case. Also, shows the single operation limit in contrast to using regex.

### DIFF
--- a/live-examples/js-examples/string/string-replace.html
+++ b/live-examples/js-examples/string/string-replace.html
@@ -4,6 +4,11 @@
 var regex = /dog/gi;
 
 console.log(p.replace(regex, 'ferret'));
+console.log(p.replace('dog', 'monkey'));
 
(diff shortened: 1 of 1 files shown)


wbamberg:
@EddieOne , thanks ! I like this example (also that it shows the original string is not modified) but the convention is to have a separate `// expected output"` line for each console.log, so this should be:

```js
var p = 'The quick brown fox jumps over the lazy dog. If the dog reacted, was it really lazy?';

var regex = /dog/gi;

console.log(p.replace(regex, 'ferret'));
// expected output: "The quick brown fox jumps over the lazy ferret. If the ferret reacted, was it really lazy?"

console.log(p.replace('dog', 'monkey'));
// expected output: "The quick brown fox jumps over the lazy monkey. If the dog reacted, was it really lazy?"
```

EddieOne:
Cool, thanks!

PR #1322, opened 2019-02-23
Add comparison, that return true

username1565:
Add comparison example, that return - true.
I cann't edit this here: https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/RegExp


wbamberg:
Thanks for your contribution! But I'm not sure what this change does to improve the example.

(I do think the example currently in the page isn't very helpful, and one that shows actual usage of the regex we constructed would be more helpful.)

username1565:
So I leave old code. No any replacement, just add. See commit.

PR #1324, opened 2019-03-07
Add examples for matchAll

Elchi3:
Examples for:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/matchAll
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/@@matchAll

This will work in Chrome 73 and Firefox 67 (or the nightly versions of these browsers).

### DIFF
--- a/live-examples/js-examples/regexp/regexp-prototype-@@matchall.html
+++ b/live-examples/js-examples/regexp/regexp-prototype-@@matchall.html
@@ -1,0 +1,16 @@
+<pre>
+<code id="static-js">class MyRegExp extends RegExp {
+    [Symbol.matchAll](str) {
+        var result = RegExp.prototype[Symbol.matchAll].call(this, str);
+        if (!result) return null;
+        return Array.from(result);
+    }
+}
+var re = new MyRegExp('([0-9]+)-([0-9]+)-([0-9]+)', 'g');
+var result = '2016-01-02|2019-03-07'.matchAll(re);
+console.log(result[0]);
+// expected output: [ "2016-01-02", "2016", "01", "02" ]
+console.log(result[1]);
+// expected output: [ "2019-03-07", "2019", "03", "07" ]
+</code>
+</pre>
--- a/live-examples/js-examples/regexp/regexp-prototype-@@matchall.html
+++ b/live-examples/js-examples/regexp/regexp-prototype-@@matchall.html
@@ -1,16 +1,16 @@
 <pre>
 <code id="static-js">class MyRegExp extends RegExp {
     [Symbol.matchAll](str) {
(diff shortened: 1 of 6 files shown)


live-examples/js-examples/string/string-matchall.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">var regexp = /t(e)(st(\d?))/g;
+var str = 'test1test2';
+
+let array = [...str.matchAll(regexp)];
+
+console.log(array[0]);
+// expected output: ['test1', 'e', 'st1', '1']
```
wbamberg: `Array ["test1", "e", "st1", "1"]`

live-examples/js-examples/string/string-matchall.html
```diff
@@ -0,0 +1,13 @@
...
+
+console.log(array[1]);
+// expected output: ['test2', 'e', 'st2', '2']
```
wbamberg: `Array ["test2", "e", "st2", "2"]`

live-examples/js-examples/symbol/symbol-matchall.html
```diff
@@ -0,0 +1,9 @@
+<pre>
+<code id="static-js">var re = /[0-9]+/g;
+var str = '2016-01-02|2019-03-07';
+var result = re[Symbol.matchAll](str);
+
+console.log(Array.from(result, x => x[0]));
+// expected output: ["2016", "01", "02", "2019", "03", "07"]
```
wbamberg: `Array ["2016", "01", "02", "2019", "03", "07"]`

live-examples/js-examples/regexp/regexp-prototype-@@matchall.html
```diff
@@ -0,0 +1,16 @@
...
+    }
+}
+var re = new MyRegExp('([0-9]+)-([0-9]+)-([0-9]+)', 'g');
+var result = '2016-01-02|2019-03-07'.matchAll(re);
+console.log(result[0]);
+// expected output: [ "2016-01-02", "2016", "01", "02" ]
+console.log(result[1]);
+// expected output: [ "2019-03-07", "2019", "03", "07" ]
+</code>
+</pre>
```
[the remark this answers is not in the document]
Elchi3: Thanks Will, I think this is less dense and thus better.

Elchi3:
I've updated this and hopefully addressed your comments. Thanks for your review! 
Can you take another look and merge if this is good to go?

PR #1333, opened 2019-03-26
Correcting `contenteditable` attribute value in `spellcheck` example

gridlined:
As per the spec for `contenteditable`:

> This attribute is an enumerated one and not a Boolean one. This means that the
> explicit usage of one of the values `true`, `false` or the empty string is
> mandatory and that a shorthand like `<label contenteditable>Example Label</label>`
> is not allowed. The correct usage is `<label contenteditable="true">Example Label</label>`.

The `spellcheck` example was incorrect using it without a value.

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-spellcheck.html
+++ b/live-examples/html-examples/global-attributes/attribute-spellcheck.html
@@ -1,3 +1,3 @@
-<p contenteditable spellcheck="true">This exampull will be checkd fur spellung when you try to edit it.</p>
+<p contenteditable="true" spellcheck="true">This exampull will be checkd fur spellung when you try to edit it.</p>
 
-<p contenteditable spellcheck="false">This exampull will nut be checkd fur spellung when you try to edit it.</p>
+<p contenteditable="true" spellcheck="false">This exampull will nut be checkd fur spellung when you try to edit it.</p>


wbamberg:
I'm not sure about this. Our own docs page says:

> This attribute is an enumerated one and not a Boolean one. This means that the explicit usage of one of the values true, false or the empty string is mandatory and that **a shorthand like \<label contenteditable>Example Label\</label> is not allowed**. The correct usage is \<label contenteditable="true">Example Label\</label>.

This seems very clear: the current example for `spellcheck` is wrong, and this PR fixes it. But the W3C docs also say:

> Attributes can be specified in four different ways:
>
> 1. Empty attribute syntax
> 
> Just the attribute name. **The value is implicitly the empty string.**

(https://www.w3.org/TR/html/syntax.html#elements-attributes)

...so writing just `contenteditable` gives it the value `""` which makes it `true`.

Also the W3C's own example uses the shorthand form:

```html
<!doctype html>
<html lang=en>
<title>Live CSS editing!</title>
<style style=white-space:pre contenteditable>
html { margin:.2em; font-size:2em; color:lime; background:purple }
head, title, style { display:block }
body { display:none }
</style>
```

(https://html.spec.whatwg.org/multipage/interaction.html#attr-contenteditable)

So I wonder if our docs are wrong. @estelle , @chrisdavidmills , do you know the answer to this?

estelle:
https://www.w3.org/TR/html/editing.html#element-attrdef-global-contenteditable 

> The contenteditable content attribute is an enumerated attribute whose keywords are the empty string, true, and false. The empty string and the true keyword map to the true state. The false keyword maps to the false state. In addition, there is a third state, the inherit state, which is the missing value default (and the invalid value default).

So our docs are incorrect.

gridlined:
I was unsure myself, and thought the same, but created the PR nonetheless because it was either than this change was needed and the PR would be of use, or having created the PR would spur discussion just as it has and end up in a fix to the other documentation.

wbamberg:
Thanks @estelle  for the clarification and @gridlined for bringing this up. I've filed https://github.com/mdn/sprints/issues/1596 to fix the docs.

PR #1338, opened 2019-04-11
Fix RegExp test example

alexdachin:
Previously, the example was incorrect:

```
console.log(regex2.test(str1));
// expected output: true

console.log(regex2.test(str1));
// expected output: false
```

### DIFF
--- a/live-examples/js-examples/regexp/regexp-prototype-test.html
+++ b/live-examples/js-examples/regexp/regexp-prototype-test.html
@@ -1,18 +1,12 @@
 <pre>
-<code id="static-js" data-height="taller">var regex1 = RegExp('foo*');
-var regex2 = RegExp('foo*','g');
+<code id="static-js" data-height="taller">var regex = RegExp('foo*');
 var str1 = 'table football';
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your PR! But I think the current example is correct:

```js
var regex2 = RegExp('foo*','g');
var str1 = 'table football';

console.log(regex2.test(str1));
// expected output: true

console.log(regex2.test(str1));
// expected output: false
```

...the point being that, [in the words of the doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#Description), "test() called multiple times on the same global regular expression instance will advance past the previous match" - so calling it again does not find a match. Try it with, say `var str1 = 'table footballfoo';` and see the difference.

alexdachin:
Thank you for pointing that out and sorry for assuming stuff. I only read the description and didn’t notice the “Using test() on a regex with the global flag” section.

wbamberg:
No problem! I appreciate your willingness to help out with the project!

PR #1345, opened 2019-05-02
Update object-freeze.html

stevenvachon:
This was implying that Object.freeze() creates a copy, when it explicitly does not.

### DIFF
--- a/live-examples/js-examples/object/object-freeze.html
+++ b/live-examples/js-examples/object/object-freeze.html
@@ -1,14 +1,14 @@
 <pre>
-<code id="static-js">const object1 = {
-  property1: 42
+<code id="static-js">const obj = {
+  prop: 42
 };
 
-const object2 = Object.freeze(object1);
+Object.freeze(obj);
 
-object2.property1 = 33;
+obj.prop = 33;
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/object/object-freeze.html
```diff
@@ -1,14 +1,14 @@
 <pre>
-<code id="static-js">const object1 = {
+<code id="static-js">const object = {
   property1: 42
 };
 
-const object2 = Object.freeze(object1);
+Object.freeze(object1);
```
wbamberg: The variable declared at line 2 is renamed to `object` but this isn't, so we'll get an error here.

I think we should keep the name `object1` - I'm uncomfortable with a variable named `object`, and using the suffix is recommended in the style guide: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#numerical-suffixes.

stevenvachon: How about `obj` ?

wbamberg: Yes, I'd be happy with that too :).

PR #1346, opened 2019-05-15
Update regexp-prototype-test.html

ackvf:
Seeing two different return values for identical statements seemed hell confusing for an unsuspecting viewer, especially when it is not mentioned anywhere nearby. I think that exposing what is going under the hood is a good introduction and will suggest the viewer to continue reading.

### DIFF
--- a/live-examples/js-examples/regexp/regexp-prototype-test.html
+++ b/live-examples/js-examples/regexp/regexp-prototype-test.html
@@ -9,10 +9,10 @@ console.log(regex1.test(str1));
 console.log(regex1.test(str1));
 // expected output: true
 
-console.log(regex2.test(str1));
-// expected output: true
+console.log(regex2.test(str1), regex2.lastIndex);
+// expected output: true 9
 
-console.log(regex2.test(str1));
-// expected output: false
+console.log(regex2.test(str1), regex2.lastIndex);
+// expected output: false 0
 </code>
 </pre>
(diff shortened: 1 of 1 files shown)


ackvf:
Excellent!!

May I propose a further simplification? I find it highly distracting and hard to follow, when there are actual code statements mixed with comments. Note that the bare output is still consistent with other examples, like [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array).
```
var str1 = 'table football';

var regex1 = RegExp('foo*');
var globalRegex1 = RegExp('foo*','g');

console.log(regex1.test(str1));
// true

console.log(regex1.test(str1));
// true

console.log(globalRegex1.lastIndex);
// 0

console.log(globalRegex1.test(str1));
// true

console.log(globalRegex1.lastIndex);
// 9

console.log(globalRegex1.test(str1));
// false
```

wbamberg:
The `// expected output: foo` thing is a convention across all the JS interactive examples, and I'm a fan of consistency. I'm open to this suggestion, but the way to make it happen would be:

1. propose the change in https://discourse.mozilla.org/c/mdn
2. if there's consensus, change the [documented guidelines](https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#indicating-consolelog-output) and update all the examples.

I think this might be an improvement but would need to think more about whether it would work well across all our interactive examples or if there are places where it would be confusing. The nice thing about the current method is that it's very explicit (at the cost of being more wordy).

Also it might take a while to get done, even if everyone agrees, since people are busy :).

PR #1348, opened 2019-05-16
Update border-color.html

mbrookes:


wbamberg:
Thanks for filing this PR!

As I understand it this and the related https://github.com/mdn/interactive-examples/pull/1349 are about using the `rgba()` and `hsla()` forms for color rather than `rgb()` and `hsl()`, because we're passing an alpha component as the fourth argument.

According to https://developer.mozilla.org/en-US/docs/Web/CSS/color_value, in CSS Color Level 4 `rgb()` and `hsl()` behave the same as `rgba()` and `hsla()`. I can't find any good information about browser support for this. These examples work in Firefox and Chrome, but I just tested in Safari and they do not.

So we should merge this :).

I've also filed https://github.com/mdn/browser-compat-data/issues/4169 to have better compat data for this feature.

PR #1352, opened 2019-05-18
Improve <blockquote> example semantics

teddybradford:
- Related to #803


teddybradford:
I'm actually kind of unclear as to whether a title of a work inside a <cite> tag should be wrapped in <i> tags. Any thoughts?

teddybradford:
I removed the `<i>` tag after seeing #1342

PR #1353, opened 2019-05-18
Fix unclosed <p> tag in strikethrough example

teddybradford:


teddybradford:
Note to self: Don't try to make PRs late at night or you'll mess up your branching. :(

PR #1365, opened 2019-05-28
Add An Easy Demonstration Of 2D Matrices

anonyco:
The [setTransform](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setTransform), [transform](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/transform), and [clearRect](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clearRect) pages of the CanvasRenderingContext2D could really use a demonstration of how 2D matrices work for those among us who do not know the math yet. I believe that my code (perhaps with some modifications from other people) provides a good solid demonstration. My JSFiddle can be found at https://jsfiddle.net/m3etvgyp/

### DIFF
--- a/live-examples/webapi-examples/CanvasRenderingContext2D/meta.json
+++ b/live-examples/webapi-examples/CanvasRenderingContext2D/meta.json
@@ -1,0 +1,13 @@
+{
+    "pages": {
+        "setTransform": {
+            "exampleCode": "./live-examples/webapi-examples/CanvasRenderingContext2D/setTransform/canvas2D-setTransform.html",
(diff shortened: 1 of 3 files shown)


wbamberg:
@anonyco , thanks for your contribution!

But we're not yet ready to add Web API interactive examples at this point, and when we do we will need to redo the editor in use, since it's not really fit for purpose. And unfortunately the project to add interactive examples for Web APIs is on hold at the moment, so I can't let you know when we will be ready (most likely not in 2019 as far as I can tell).

anonyco:
@wbamberg Thank you  so much for your helpful explanation. Your explanation answers all of my questions and then some.

~ Happy Coding!

PR #1377, opened 2019-06-13
Create promise-allsettled.html

MarkBennett:
Add an example of the `Promise.allSettled()` method as it will be shipping soon.

The style is based off of the [`Promise.all()`](https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/promise/promise-all.html) example currently on MDN. See the TC39 Proposal for more examples of how [`Promise.allSettled()`](https://github.com/tc39/proposal-promise-allSettled) works.

This is my first PR for this project so please let me know if I've done anything wrong.

Thanks for your work on MDN. It rocks! 😄

### DIFF
--- a/live-examples/js-examples/promise/meta.json
+++ b/live-examples/js-examples/promise/meta.json
@@ -6,6 +6,12 @@
             "title": "JavaScript Demo: Promise.all()",
             "type": "js"
         },
+        "promiseAllSettled": {
+            "exampleCode": "./live-examples/js-examples/promise/promise-allsettled.html",
+            "fileName": "promise-allsettled.html",
+            "title": "JavaScript Demo: Promise.allSettled()",
+            "type": "js"
+        },
         "promiseCatch": {
             "exampleCode": "./live-examples/js-examples/promise/promise-catch.html",
             "fileName": "promise-catch.html",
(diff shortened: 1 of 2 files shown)


MarkBennett:
Thanks for the great feedback @wbamberg!

> You need to update the [meta.json file for promises](https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/promise/meta.json), to tell the builder about this new example and supply some metadata.

Will do! :)



I didn't realize that the example would get cut off and agree about not wanting to scroll. How does this look for a shorter example? It's the same as yours but I made it a bit shorter by using `forEach()` instead. Is using `for..of` preferred?

```js
const promise1 = Promise.resolve(3);
const promise2 = new Promise((resolve, reject) => setTimeout(reject, 100, 'foo'));
const promises = [promise1, promise2];

Promise.allSettled(promises).
  then((results) => results.forEach((result) => console.log(result.status)));

// expected output:
// "fulfilled"
// "rejected"
```

Let me know if that's ok and I'll update the PR.

MarkBennett:
Thanks for your review and help with this @wbamberg. I've updated MDN to include this sample! :)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

PR #1385, opened 2019-07-01
Fix incorrect expected output

BenAlderfer:


wbamberg:
Thanks for your PR @BenAlderfer !

When I try the [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) example in Chrome and Firefox I see the result that's noted in the comment:

```
> Object { a: 1, b: 4, c: 5 }
> Object { a: 1, b: 4, c: 5 }
```

...I think because `assign()` mutates the `target` parameter itself, and also returns it (so `target === returnedTarget`). There's more chat about this in https://github.com/mdn/interactive-examples/pull/1273.

So I think the existing example is correct. But if I have missed something here I'd be very happy to be corrected!

BenAlderfer:
Thanks for the clarification @wbamberg.

PR #1386, opened 2019-07-06
Intl.Locale.prototype.maximize() interactive example

jahzielv:
Creating an interactive example for `Intl.Locale.prototype.maximize()`. 

Example renders correctly and outputs correct values.

I've been writing the docs for `Intl.Locale`, and am now opening up PRs for interactive examples.

Thanks!

PR #1388, opened 2019-07-06
Intl.Locale.prototype.minimize() interactive example

jahzielv:
Added an interactive example for `Intl.Locale.prototype.minimize()`.

Part of an [MDN documentation effort](https://github.com/tc39/proposal-intl-locale/issues/70) for `Intl.Locale` started by Daniel Ehrenberg.

Thanks!


wbamberg:
Thanks for this @jahzielv !

As for https://github.com/mdn/interactive-examples/pull/1386#issuecomment-509409636, we could use `const` here instead of `let`. Also as for https://github.com/mdn/interactive-examples/pull/1386#issuecomment-509409636, we could have an example with another language, although it seems less important than with `maximize()` since its behavior is less interesting.

live-examples/js-examples/intl/meta.json
```diff
@@ -60,6 +60,12 @@
             "title": "JavaScript Demo: Intl.ListFormat",
             "type": "js"
         },
+        "intlLocalePrototypeMinimize": {
+            "exampleCode": "./live-examples/js-examples/intl/intl-locale-prototype-minimize.html",
+            "fileName": "intl-locale-prototype-minimize.html",
+            "title": "JavaScript Demo: Intl.Locale.prototype.minimize",
```
wbamberg: This should be "JavaScript Demo: Intl.Locale.prototype.minimize()".

PR #1389, opened 2019-07-06
Intl.Locale.prototype.toString() interactive example

jahzielv:
Added an interactive example for Intl.Locale.prototype.toString().

Part of an [MDN documentation effort](https://github.com/tc39/proposal-intl-locale/issues/70) for Intl.Locale started by Daniel Ehrenberg.

Thanks!


wbamberg:
As for the others, we could use `const` here and it would be illustrative to have more than one example:

```js
const myLoc1 = new Intl.Locale("fr-Latn-FR", {calendar: "gregory", hourCycle: "h12"});
console.log(myLoc1.toString());
// expected output: "fr-Latn-FR-u-ca-gregory-hc-h12"

const myLoc2 = new Intl.Locale("fr-Latn-FR", {calendar: "hebrew", hourCycle: "h24"});
console.log(myLoc2.toString());
// expected output: "fr-Latn-FR-u-ca-hebrew-hc-h24"
```

live-examples/js-examples/intl/meta.json
```diff
@@ -60,6 +60,12 @@
             "title": "JavaScript Demo: Intl.ListFormat",
             "type": "js"
         },
+        "intlLocalePrototypeToString": {
+            "exampleCode": "./live-examples/js-examples/intl/intl-locale-prototype-tostring.html",
+            "fileName": "intl-locale-prototype-tostring.html",
+            "title": "JavaScript Demo: Intl.Locale.prototype.toString",
```
wbamberg: As for the others, this should be "JavaScript Demo: Intl.Locale.prototype.toString()"

jahzielv: Changes made! :D

PR #1391, opened 2019-07-09
Offset Path (CSS Motion Path) properties

danwilson:
This PR adds demos for `offset-path` (only showing the `path()` syntax for now since that is the only supported function in browsers currently (Blink and Firefox Nightly)), `offset-rotate`, and `offset-distance`.

The offset properties are interesting that by themselves they don't really show much. For example, defining an `offset-path` but nothing else shows very little, as it simply defines a path that it will follow when adjusted by the other `offset` properties. Similarly `offset-distance` and `offset-rotate` are meaningless without an `offset-path` defined.

For `offset-distance` and `offset-rotate`, I added a background image that represents a common `offset-path` that the examples all use. I'm concerned that could be confusing since the path is technically invisible, but it felt more important to have a visual reference for the path it is following in those contexts.

For `offset-path` and `offset-rotate` I added a playback control, since they are easier to understand if you animate the `offset-distance` from `0%` to `100%`. I did not want to have the animation auto playing for accessibility reasons, and that followed a pattern I saw with other animating demos in this repository.

### DIFF
--- a/live-examples/css-examples/motion-path/meta.json
+++ b/live-examples/css-examples/motion-path/meta.json
@@ -3,6 +3,7 @@
     "offsetPath": {
         "cssExampleSrc": "./live-examples/css-examples/motion-path/offset-path.css",
         "exampleCode": "./live-examples/css-examples/motion-path/offset-path.html",
+        "jsExampleSrc": "./live-examples/css-examples/motion-path/offset-playback.js",
         "fileName": "offset-path.html",
         "title": "CSS Demo: offset-path",
         "type": "css"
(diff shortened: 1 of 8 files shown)


wbamberg:
I think these look great!

The only comment I have is that the CSS choices should have semicolons after them:

```css
offset-path: path('M-70,-40 C-70,70 70,70 70,-40');
```

> I added a background image that represents a common offset-path that the examples all use. I'm concerned that could be confusing 

I think this is fine and we do quite similar stuff with some other examples (e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin).

> For offset-path and offset-rotate I added a playback control, since they are easier to understand if you animate the offset-distance from 0% to 100%. I did not want to have the animation auto playing for accessibility reasons, and that followed a pattern I saw with other animating demos in this repository.

Yes, I think this is the right thing to do here as well.

These are very illustrative, well-thought out and good-looking examples that help a lot to explain these properties. Thanks!

danwilson:
Thanks very much! I've added the semicolons.

PR #1394, opened 2019-07-14
Interactive example for Intl.Locale

jahzielv:
Added an interactive example for Intl.Locale. Demonstrates several of Locale's properties.

Part of an [MDN documentation effort](https://github.com/tc39/proposal-intl-locale/issues/70) for Intl.Locale on behalf of ECMA 402.

Thanks!


live-examples/js-examples/intl/intl-locale.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">const korean = new Intl.Locale("ko", {script: "Kore", region: "KR", hourCycle: "h24", calendar: "gregory"});
```
wbamberg: Rather than have it on one line it would be easier to read like:

```js
const korean = new Intl.Locale("ko", {
  script: "Kore", region: "KR", hourCycle: "h24", calendar: "gregory"
});
```

wbamberg: Also here and elsewhere, single quotes for strings is our style.
(although note: use double for the expected output strings)

live-examples/js-examples/intl/intl-locale.html
```diff
@@ -0,0 +1,12 @@
...
+const japan = new Intl.Locale("ja-Jpan-JP-u-ca-japanese-hc-h12")
```
wbamberg: Here and elsewhere, end the line with a semicolon.

wbamberg: Also we should be consistent with variable names: so either:

"japan" and "korea"
or
"japanese" and "korean"

wbamberg: Also a blank space before this line (to separate the two declarations) would help readability I think.

live-examples/js-examples/intl/intl-locale.html
```diff
@@ -0,0 +1,12 @@
...
+console.log(korean.baseName)
+// expected output: "ko-Kore-KR"
+
+console.log(japan.baseName)
+// expected output: "ja-Jpan-JP"
```
wbamberg: Since my above comments make the example longer, why not compress this bit as you do below, with something like:

```js
console.log(korean.baseName, japan.baseName)
// expected output: "ko-Kore-KR" "ja-Jpan-JP"
```

live-examples/js-examples/intl/intl-locale.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">const korean = new Intl.Locale("ko", {script: "Kore", region: "KR", hourCycle: "h24", calendar: "gregory"});
+const japan = new Intl.Locale("ja-Jpan-JP-u-ca-japanese-hc-h12")
+console.log(korean.baseName)
```
wbamberg: Also a blank line before this (after the 2 declarations) would help readability.

jahzielv:
Requested changes made! :D

jahzielv:
Thanks @wbamberg! My pleasure :D

PR #1408, opened 2019-08-03
Fix #1404 - array filter example is not inline with the browser compatibility table

### LINKED ISSUE #1404: array filter example is not inline with the browser compatibility table
Array filter example is not inline with the browser compatibility table.
Using the arrow function is not compatible with IE11 and bellow.

Suggested fix:
``
words.filter(function(word) { return word.length > 6; });
``

joedarc:
Fixes #1404

According to https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#language-choice-es5es6

> For more established example content, such as Arrays, it is recommended that we stick with ES5.

This patch downgrades 4 array examples to ES5 to align with the compatibility chart for each of the APIs.

### DIFF
--- a/live-examples/js-examples/array/array-filter.html
+++ b/live-examples/js-examples/array/array-filter.html
@@ -1,7 +1,9 @@
 <pre>
 <code id="static-js">var words = ['spray', 'limit', 'elite', 'exuberant', 'destruction', 'present'];
 
-const result = words.filter(word => word.length > 6);
+var result = words.filter(function(word) {
+  return word.length > 6;
+});
 
 console.log(result);
 // expected output: Array ["exuberant", "destruction", "present"]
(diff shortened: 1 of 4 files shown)


wbamberg:
Thanks for filing @joedarc !

But I think, as noted in https://github.com/mdn/interactive-examples/issues/1404#issuecomment-517415773, that it might be better to update [our style guide](https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md) to point to [MDN's general JS examples style guide](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Code_guidelines/JavaScript). I can't see a good reason really for interactive-examples to have different rules from other examples on MDN.

joedarc:
Thanks @wbamberg, i'll file a PR to update the style guide instead.

PR #1415, opened 2019-08-13
Adding examples overflow-block and overflow-inline

rachelandrew:
These properties are shipping in Firefox 69, and the examples are essentially a copy of `overflow-x` and `overflow-y` as the flow relative properties map to these.

https://bugzilla.mozilla.org/show_bug.cgi?id=1470695


rachelandrew:
Ugh sorry! I've added that. And yes we probably don't add them yet, but I figured if I did them they would be there. I have added live examples to pages when going to update the BCD due to a feature shipping somewhere.

PR #1416, opened 2019-08-13
Added interactive example for dotAll

sabahat70:
Added interative example for dotAll in js-examples, regexp. Also changed the meta.json file.

### DIFF
--- a/live-examples/js-examples/regexp/meta.json
+++ b/live-examples/js-examples/regexp/meta.json
@@ -66,6 +66,13 @@
             "title": "JavaScript Demo: RegExp.prototype.global",
             "type": "js"
         },
+        "regexPrototypeDotAll":{
+            "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-dotall.html",
+            "fileName": "regexp-prototype-dotall.html",
+            "title": "JavaScript Demo: RegExp.prototype.dotall",
+            "type": "js"
+            
+        },
         "regexpPrototypeIgnoreCase": {
             "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-ignorecase.html",
             "fileName": "regexp-prototype-ignorecase.html",
--- a/live-examples/js-examples/regexp/meta.json
+++ b/live-examples/js-examples/regexp/meta.json
@@ -48,6 +48,12 @@
             "title": "JavaScript Demo: RegExp.prototype[@@split]",
             "type": "js"
         },
+        "regexPrototypeDotAll":{
+            "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-dotall.html",
+            "fileName": "regexp-prototype-dotall.html",
+            "title": "JavaScript Demo: RegExp.prototype.dotAll",
+            "type": "js"
+        },
         "regexpPrototypeExec": {
             "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-exec.html",
             "fileName": "regexp-prototype-exec.html",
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/regexp/meta.json
```diff
@@ -66,6 +66,13 @@
             "title": "JavaScript Demo: RegExp.prototype.global",
             "type": "js"
         },
+        "regexPrototypeDotAll":{
```
wbamberg: This should be `"regexpPrototypeDotAll"` and this file should be ordered, so it should go after `"regexpPrototype@@Split"`

live-examples/js-examples/regexp/meta.json
```diff
@@ -66,6 +66,13 @@
...
+            "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-dotall.html",
+            "fileName": "regexp-prototype-dotall.html",
+            "title": "JavaScript Demo: RegExp.prototype.dotall",
```
wbamberg: This should be "JavaScript Demo: RegExp.prototype.dotAll"

live-examples/js-examples/regexp/meta.json
```diff
@@ -66,6 +66,13 @@
...
+            "type": "js"
+
```
wbamberg: Please remove this blank line.

live-examples/js-examples/regexp/regexp-prototype-dotall.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">
+var regex1 = new RegExp('foo', 's');
```
wbamberg: We should use `const` for these variables.

wbamberg: There's a blank line at the start of this example. The source should be like:

    <code id="static-js">const regex1 = new RegExp('foo', 's');

Weird I know.

live-examples/js-examples/regexp/regexp-prototype-dotall.html
```diff
@@ -0,0 +1,13 @@
...
+
+console.log(regex1.dotAll);
+// expected output: true
+
+var regex2 = new RegExp('bar', 'i');
+
+console.log(regex2.dotAll);
+// expected output: false
+</code>
+</pre>
```
wbamberg: Please end files with a newline.

live-examples/js-examples/regexp/regexp-prototype-dotall.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js">
+var regex1 = new RegExp('foo', 's');
+
+console.log(regex1.dotAll);
+// expected output: true
+
+var regex2 = new RegExp('bar', 'i');
```
wbamberg: Also use `const` here. And I'm not sure why you are passing the `'i'` flag? It would be clearer I think if you omitted it, unless there's something I am missing here.

sabahat70: Hi, I used the 'i' flag to demonstrate the case of this function returning an output of 'false', if that makes sense.

sabahat70:
Hi,
I have made the requested changes you had suggested. Kindly, review.

live-examples/js-examples/regexp/regexp-prototype-dotall.html
wbamberg: But I don't think you need `'i'` for this example, do you? `dotAll` will be `false` if you just don't pass any flags.

sabahat70: I just checked and I see that it results in false without the flag as well. Should I update the example with no flag argument?

wbamberg: Yes, please! I think it would be clearer like that.

sabahat70:
Thank you all! I am glad I was able to make this contribution.

PR #1419, opened 2019-08-19
Updated img example to HTML5

TKirmani:
The example used '/>' which was used in XHTML.
Now it uses proper HTML5. i.e. '>'


TKirmani:
@wbamberg My pleasure

PR #1434, opened 2019-09-06
Clarified retval and added example of multiple params

aeinbu:
This PR adds:
- Clarified that calling  `push` returns a count.
- Showing how to call to `push` with multiple parameters, and the result of that.

### DIFF
--- a/live-examples/js-examples/array/array-push.html
+++ b/live-examples/js-examples/array/array-push.html
@@ -1,15 +1,18 @@
 <pre>
 <code id="static-js">var animals = ['pigs', 'goats', 'sheep'];
 
-console.log(animals.push('cows'));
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/array/array-push.html
```diff
@@ -1,15 +1,15 @@
 <pre>
-<code id="static-js">var animals = ['pigs', 'goats', 'sheep'];
+<code id="static-js">
+const animals = ['pigs', 'goats', 'sheep'];
```
wbamberg: This has the effect of adding a blank line at the start of the example. It should be:

```
<code id="static-js">const animals = ['pigs', 'goats', 'sheep'];
```

(I know, this is very non-obvious).

PR #1440, opened 2019-09-16
adding examples for new css text properties - text-decoration-thickne…

chrisdavidmills:
…ss and text-underline-offset

As per https://bugzilla.mozilla.org/show_bug.cgi?id=1573631.

These will be enabled in release as of Fx 70. Therefore I've:

* Added an example for `text-decoration-thickness`. I'll not add it to the relevant property page as yet, because it will only be supported in Firefox, when v 70 comes out. Maybe wait till it has wider support?
* Added an example for `text-underline-offset`. I'll not add it to the relevant property page as yet, because it will only be supported in Firefox, when v 70 comes out. Maybe wait till it has wider support?
* Started to display the relevant interactive example on https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink, is it will be supported across Fx and Chromium as of the Fx 70 release

I was also considering updating the interactive example at https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration to show a version that includes the text-decoration-thickness value as part of the shorthand. But this won't work anywhere except Firefox, and not until 70 is released. Do you think I should leave this for now?

### DIFF
--- a/live-examples/css-examples/text-decoration/meta.json
+++ b/live-examples/css-examples/text-decoration/meta.json
@@ -35,6 +35,13 @@
             "title": "CSS Demo: text-decoration-style",
             "type": "css"
         },
+        "textDecorationThickness": {
+            "cssExampleSrc": "./live-examples/css-examples/text-decoration/text-decoration-thickness.css",
+            "exampleCode": "./live-examples/css-examples/text-decoration/text-decoration-thickness.html",
+            "fileName": "text-decoration-thickness.html",
+            "title": "CSS Demo: text-decoration-thickness",
+            "type": "css"
+        },
         "textEmphasis": {
             "cssExampleSrc": "./live-examples/css-examples/text-decoration/text-emphasis.css",
             "exampleCode": "./live-examples/css-examples/text-decoration/text-emphasis.html",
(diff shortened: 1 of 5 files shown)


wbamberg:
> * Added an example for `text-decoration-thickness`. I'll not add it to the relevant property page as yet, because it will only be supported in Firefox, when v 70 comes out. Maybe wait till it has wider support?
> 
> * Added an example for `text-underline-offset`. I'll not add it to the relevant property page as yet, because it will only be supported in Firefox, when v 70 comes out. Maybe wait till it has wider support?

This all sounds sensible.

> * Started to display the relevant interactive example on https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip-ink, is it will be supported across Fx and Chromium as of the Fx 70 release

Yay!

> I was also considering updating the interactive example at https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration to show a version that includes the text-decoration-thickness value as part of the shorthand. But this won't work anywhere except Firefox, and not until 70 is released. Do you think I should leave this for now?

Yes, I think you should leave it.

I wish we had a better story for handling unsupported properties.

wbamberg:
Similarly in the `text-decoration-thickness` example, `3px` and `0.3rem` are quite close. Maybe have `2px` and `0.5rem`?

chrisdavidmills:
Thanks for the comments Will. I totally agree with you `text-underline-offset`, so I've made the suggestions as given.

For `text-decoration-thickness`, I've updated the second choice to `0.4rem`, not `0.5rem`, as I think the different is noticeable, but it isn't quite as over the top.

PR #1444, opened 2019-09-23
Add Int-DateTimeFormat-formatRangeToParts

romulocintra:


live-examples/js-examples/intl/intl-datetimeformat-prototype-formatrangetoparts.html
```diff
@@ -0,0 +1,15 @@
...
+// > 'Wed, 10 Jan 2007 10:00:00 GMT'
+// > 'Wed, 10 Jan 2007 11:00:00 GMT'
+
+var dateTimeFormat = new Intl.DateTimeFormat("en", {
+    hour: 'numeric',
+    minute: 'numeric'
+});
+
+console.log(dateTimeFormat.formatRangeToParts(startDate, endDate));
+// expected output: Array  [ { type: "hour", value: "11", source: "startRange" }, { type: "literal", value: ":", source: "startRange" }, { type: "minute", value: "00", source: "startRange" }, { type: "literal", value: " ", source: "startRange" }, { type: "dayPeriod", value: "AM", source: "startRange" }, { type: "literal", value: " – ", source: "shared" }, { type: "hour", value: "12", source: "endRange" }, { type: "literal", value: ":", source: "endRange" }, { type: "minute", value: "00", source: "endRange" }, { type: "literal", value: " ", source: "endRange" }, { type: "dayPeriod", value: "PM", source: "endRange" }]
```
wbamberg: This output is too long, it's not readable. The simplest thing would be just something like:

```js
var parts = dateTimeFormat.formatRangeToParts(startDate, endDate);
console.log(parts[0]);
```

romulocintra: I agree that is too long , but there is some relevant info coming from that output , that make API relevant :(

wbamberg: Maybe something like:

```js
var startDate = new Date(Date.UTC(2007, 0, 10, 10, 0, 0)); // > 'Wed, 10 Jan 2007 10:00:00 GMT'
var endDate = new Date(Date.UTC(2007, 0, 10, 11, 0, 0));   // > 'Wed, 10 Jan 2007 11:00:00 GMT'

var dateTimeFormat = new Intl.DateTimeFormat("en", {
    hour: 'numeric',
    minute: 'numeric'
});

var parts = dateTimeFormat.formatRangeToParts(startDate, endDate);
for (let part of parts) {
  console.log(part);
}
// expected output:
// Object { type: "hour", value: "2", source: "startRange" }
// Object { type: "literal", value: ":", source: "startRange" }
// Object { type: "minute", value: "00", source: "startRange" }
// Object { type: "literal", value: " – ", source: "shared" }
// Object { type: "hour", value: "3", source: "endRange" }
// Object { type: "literal", value: ":", source: "endRange" }
// Object { type: "minute", value: "00", source: "endRange" }
// Object { type: "literal", value: " ", source: "shared" }
// Object { type: "dayPeriod", value: "AM", source: "shared" }
```

? If it's important to show all the "expected output" then make the example taller by setting `data-height="taller"` on the example's `code` element:

    <code id="static-js" data-height="taller">

romulocintra: You nailed it!! Will do changes as soon as possible

PR #1447, opened 2019-09-25
Tweaked Number.isFinite example to be more correct

### LINKED ISSUE #1446: !Number.isFinite() does not imply infinity
The [demo code](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) does:
```
function div(x) {
  if (Number.isFinite(1000 / x)) {
    return 'Number is NOT Infinity.';
  }
  return 'Number is Infinity!';
}
```

But `isFinite(n)` is false also for NaN, so it is incorrect to say that "Number is infinity" just because `Number.isFinite(n)` is false, as the demo does.

bjorn:
As pointed out by @PerMildner, `!Number.isFinite` does not imply infinity.

Consider `console.log(div("foo"))`.

Closes #1446

### DIFF
--- a/live-examples/js-examples/number/number-isfinite.html
+++ b/live-examples/js-examples/number/number-isfinite.html
@@ -1,15 +1,15 @@
 <pre>
 <code id="static-js">function div(x) {
   if (Number.isFinite(1000 / x)) {
-    return 'Number is NOT Infinity.';
+    return 'Number is Finite.';
   }
-  return 'Number is Infinity!';
+  return 'Number is NOT Finite!';
 }
 
 console.log(div(0));
-// expected output: "Number is Infinity!"
+// expected output: "Number is NOT Finite!"
 
 console.log(div(1));
-// expected output: "Number is NOT Infinity."
+// expected output: "Number is Finite."
 </code>
 </pre>
--- a/live-examples/js-examples/number/number-isfinite.html
+++ b/live-examples/js-examples/number/number-isfinite.html
@@ -1,18 +1,11 @@
 <pre>
-<code id="static-js">function div(x, y) {
-  if (Number.isFinite(x / y)) {
-    return 'Result is Finite.';
-  }
-  return 'Result is NOT Finite!';
-}
+<code id="static-js">console.log(Number.isFinite(1/0));
+// expected output: false
 
-console.log(div(1, 0));
-// expected output: "Result is NOT Finite!"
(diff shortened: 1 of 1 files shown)


wbamberg:
@bjorn , thanks for the PR!

In the case of `div('foo')`, then the result "Number is NOT Finite!" is still wrong, because `1000/"foo"` is not a number. Or do I misunderstand?

It seems like the `div` function is just assuming its argument is a number, so it might be better for it to test that assumption explicitly and treat non-number arguments as exceptions, if we want to care about non-numbers being passed in.

But rather than that, perhaps it would be more useful for people to have an example like:

```js
console.log(Number.isFinite(1000 / 0));
// expected output: false

console.log(Number.isFinite(1000 / 1));
// expected output: true

console.log(Number.isFinite(Infinity));
// expected output: false

console.log(Number.isFinite('foo'));
// expected output: false
```

What do you think?

PerMildner:
A good example would include the non-finite number NaN, e.g. the result of `0/0`. This is a number, but it is neither infinite nor finite.

PerMildner:
A good example would include the non-finite number NaN, e.g. the result of `0/0`. This is a number, but it is neither infinite nor finite. This seems more relevant than the fact that `isFinite()` is false for non-numbers, like strings.

bjorn:
@PerMildner Totally agree. I've adjusted the example a bit, but feel free to come up with your own as well! Just look at [the file in the repo](https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/number/number-isfinite.html) and press the small pen icon to edit it. :-)

wbamberg:
Thanks for the update. The only problem I can see is that the example is too long to fit without the user needing to scroll. I think 99% of people are never going to see the third example:

Screen Shot 2019-10-03 at 3 16 13 PM

That's partly why I suggested removing the `div()` function and just logging the result directly:

```js
console.log(Number.isFinite(1/0));
// expected output: false

console.log(Number.isFinite(10/5));
// expected output: true

console.log(Number.isFinite(0/0));
// expected output: false
```

Do you think that the `div()` function is adding any value here?

bjorn:
@wbamberg Right, I wasn't aware of the limited space these examples get. I don't think the `div` action adds any value, so feel free to go with your version instead.

wbamberg:
OK, I've made that change.

Thanks for the PR @bjorn !

PR #1448, opened 2019-09-28
Added code to show how async function returns a promise, which is the…

markbolster12:
… whole point of async


wbamberg:
Thanks for the PR @markbolster12 ! In this case I don't think it is necessary for the interactive example to include this. We want these interactive examples to be kept as simple as possible to show the basic usage of a feature, and the rest of the page can go into the details.

PR #1454, opened 2019-10-05
Improve readability and consistency of array-reduce examples

IAmFletcher:
While the changes below improve readability, it also makes the examples consistent with `Array.prototype.map` and `Array.prototype.filter`.

## Reduce
I understand the arrow function was assigned to a constant because it was used multiple times throughout the example. However, this just adds confusion, especially at a quick glance.

For the same reason, I also seperated the assignment from the `console.log` call.

In terms of the second example, I understand how seperating the arrow function may improve readability as there is second argument. However, the `reducer` declaration should be made closer to the second `reduce` call.

## Reduce Right
I split the `reduceRight` call from the declaration, and moved it onto a single line to make it consistent with Reduce, and the [JS Example Style Guide](https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#anonymous-function-definition).


schalkneethling:
Hey @IAmFletcher - would you still be keen to address the point made here?
https://github.com/mdn/interactive-examples/pull/1454#pullrequestreview-310810571

If so, this branch would also need a merge with the latest in `master`

schalkneethling:
thanks, @IAmFletcher 🎉

PR #1456, opened 2019-10-10
Added code to show how async function returns a promise, which is the…

markbolster12:
… whole point of async.

Previously, a maintainer rejected this request because they wanted to keep the example simple. That is fine, but the example should at least show an example of the functionality being displayed. Currently, the example shows a demo of a manual promise, and that await will wait for that promise. It does not show in any way what async does, other than that the keyword goes before a function declaration. As it is now, this example is not simple, it is confusing. Please take a closer look.

### DIFF
--- a/live-examples/js-examples/statement/statement-async.html
+++ b/live-examples/js-examples/statement/statement-async.html
@@ -12,8 +12,11 @@ async function asyncCall() {
   var result = await resolveAfter2Seconds();
   console.log(result);
   // expected output: 'resolved'
+  return result;
 }
 
-asyncCall();
+//The async function implicitly returns a promise
+var implicit_promise = asyncCall();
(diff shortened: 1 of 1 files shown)


wbamberg:
I've talked this over with some of the rest of the team and the answer is the same as before: the interactive example isn't the right place to introduce this.

I do think the async page is hard to follow, and is not very explicit about async making function return promises, and think it could be worth adding an example to the Examples section that clarifies this.

markbolster12:
@wbamberg  I'm having a hard time understanding why you wouldn't want to introduce the functionality being explained on the page in the interactive example at the very top of the page. Is this example displayed on some other page? The way it is now is _actively confusing_. In other words, the interactive example does not even contain an example of using an async correctly - why would we not want to fix it or remove it?

PR #1469, opened 2019-11-04
corrects usage of <keyframes-name> in `animation` shorthand

bradhenslee:
The animation's \<keyframes-name\> should be the last value provided, not the first[1] (though it looks like it might still work either way in many cases).

The position is only incorrect in this interactive code example; the wiki page itself is correct. 

1. https://www.w3.org/TR/css-animations-1/#animation

### DIFF
--- a/live-examples/css-examples/animation/animation.html
+++ b/live-examples/css-examples/animation/animation.html
@@ -1,27 +1,27 @@
 <section id="example-choice-list" class="example-choice-list" data-property="animation">
 <div class="example-choice">
-<pre><code class="language-css">animation: slidein 3s ease-in 1s infinite reverse both running;</code></pre>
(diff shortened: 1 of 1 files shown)


wbamberg:
As I read it, `||` means that the items may occur in any order (https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#Double_bar). Although there are some restrictions on the order here, as given in the spec and the MDN page, they don't rule out having `slidein` first, so the example seems to be correct.

wbamberg:
Having said that though it is perhaps worth putting the name last, since it maps more obviously onto the spec that way.

bradhenslee:
Yep, I guess it's not "incorrect" so much as unnecessarily confusing, IMO, to see it presented two different ways in two different interactives while the wiki page text attempts to explain that it could sometimes be relevant "for distinguishing animation-name values from other keywords".

PR #1470, opened 2019-11-05
Use ES6 variable declaration in globalprops live examples

ikarasz:
This PR aims to replace `var` keyword with `let` and `const` in live examples for the globalprops

### DIFF
--- a/live-examples/js-examples/globalprops/globalprops-decodeuri.html
+++ b/live-examples/js-examples/globalprops/globalprops-decodeuri.html
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">const uri = 'https://mozilla.org/?x=шеллы';
-let encoded = encodeURI(uri);
+const encoded = encodeURI(uri);
 console.log(encoded);
 // expected output: "https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B"
 
--- a/live-examples/js-examples/globalprops/globalprops-decodeuri.html
+++ b/live-examples/js-examples/globalprops/globalprops-decodeuri.html
@@ -1,6 +1,6 @@
 <pre>
-<code id="static-js">var uri = 'https://mozilla.org/?x=шеллы';
-var encoded = encodeURI(uri);
+<code id="static-js">const uri = 'https://mozilla.org/?x=шеллы';
+let encoded = encodeURI(uri);
 console.log(encoded);
 // expected output: "https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B"
 
--- a/live-examples/js-examples/globalprops/globalprops-encodeuri.html
+++ b/live-examples/js-examples/globalprops/globalprops-encodeuri.html
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">const uri = 'https://mozilla.org/?x=шеллы';
-let encoded = encodeURI(uri);
+const encoded = encodeURI(uri);
 console.log(encoded);
 // expected output: "https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B"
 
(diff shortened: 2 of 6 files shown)


live-examples/js-examples/globalprops/globalprops-parseint.html
```diff
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">function roughScale(x, base) {
-  var parsed = parseInt(x, base);
+  let parsed = parseInt(x, base);
```
wbamberg: This could be `const`.

live-examples/js-examples/globalprops/globalprops-null.html
```diff
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">function getVowels(str) {
-  var m = str.match(/[aeiou]/gi);
+  let m = str.match(/[aeiou]/gi);
```
wbamberg: This could be `const`.

live-examples/js-examples/globalprops/globalprops-infinity.html
```diff
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">var maxNumber = Math.pow(10, 1000); // max positive number
+<code id="static-js">let maxNumber = Math.pow(10, 1000); // max positive number
```
wbamberg: This could be `const`.

live-examples/js-examples/globalprops/globalprops-encodeuri.html
```diff
@@ -1,6 +1,6 @@
 <pre>
-<code id="static-js">var uri = 'https://mozilla.org/?x=шеллы';
-var encoded = encodeURI(uri);
+<code id="static-js">const uri = 'https://mozilla.org/?x=шеллы';
+let encoded = encodeURI(uri);
```
wbamberg: This could be `const`.

live-examples/js-examples/globalprops/globalprops-decodeuri.html
```diff
@@ -1,6 +1,6 @@
 <pre>
-<code id="static-js">var uri = 'https://mozilla.org/?x=шеллы';
-var encoded = encodeURI(uri);
+<code id="static-js">const uri = 'https://mozilla.org/?x=шеллы';
+let encoded = encodeURI(uri);
```
wbamberg: This could be `const`.

ikarasz:
```suggestion
const encoded = encodeURI(uri);
```

live-examples/js-examples/globalprops/globalprops-encodeuri.html
ikarasz:
```suggestion
const encoded = encodeURI(uri);
```

live-examples/js-examples/globalprops/globalprops-infinity.html
ikarasz:
```suggestion
<code id="static-js">const maxNumber = Math.pow(10, 1000); // max positive number
```

live-examples/js-examples/globalprops/globalprops-null.html
ikarasz:
```suggestion
  const m = str.match(/[aeiou]/gi);
```

live-examples/js-examples/globalprops/globalprops-parseint.html
ikarasz:
```suggestion
  const parsed = parseInt(x, base);
```

ikarasz:
@wbamberg Thanks for your review, I prefer const declaration as well. All of them are replaced now. What is the suggestion in MDN examples? Use const when I don't change the value anymore, even if there is a chance to change later?

chrisdavidmills:
> I think in many of these cases we could use const rather than let, and that seems like a better practice, although our style guide doesn't talk about it.

The only place we mention using `let`/`const` is here:

https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Code_guidelines/JavaScript#Use_ES6_features

But we don't go into much detail. I'd be happy to expand it (or have others expand it) to include some moer specific guidelines.

PR #1482, opened 2019-11-14
Adding interactive examples for nullish coalescing operator

SphinxKnight:
As [nullish coalescing operator](https://github.com/tc39/proposal-nullish-coalescing) is now a TC39 proposal of Stage 3 and has been implemented in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1566141) and in [Safari](https://webkit.org/blog/9497/release-notes-for-safari-technology-preview-89/), I wrote the [corresponding page](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and provide the corresponding interactive example.

The example worked as intended on my machine using the last Firefox Nightly.

### DIFF
--- a/live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
+++ b/live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
@@ -1,0 +1,11 @@
+<pre>
+<code id="static-js">
+const foo = null ?? "default string";
+console.log(foo);
+// expected output: "default string"
+
+const bar = "" ?? "default string";
+console.log(bar);
+// expected output: ""
+</code>
+</pre>
--- a/live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
+++ b/live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">const foo = null ?? "default string";
+<code id="static-js">const foo = null ?? 'default string';
 console.log(foo);
 // expected output: "default string"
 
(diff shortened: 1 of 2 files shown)


SphinxKnight:
for the record Chromium intends to ship and the bug is https://bugs.chromium.org/p/v8/issues/detail?id=9547

live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
```diff
@@ -0,0 +1,11 @@
...
+<code id="static-js">
+const foo = null ?? "default string";
+console.log(foo);
+// expected output: "default string"
+
+const bar = "" ?? "default string";
+console.log(bar);
+// expected output: ""
+</code>
+</pre>
```
velusgautam: Can we add couple more examples to make sure users understand it correclty

false ?? true;   // => false
0 ?? 1;          // => 0
'' ?? 'default'; // => ''

null ?? [];      // => []
undefined ?? []; // => []

live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
```diff
@@ -0,0 +1,11 @@
+<pre>
+<code id="static-js">
```
wbamberg: The newline here means the example starts with a blank line. So this should be:

```
 <code id="static-js">const foo = null ?? "default string";
```

(I know, this is really confusing and we ought to fix it.)

live-examples/js-examples/expressions/expressions-nullishcoalescingoperator.html
wbamberg: It might be worth having more, but each one takes 4 lines, so to fit in a standard editor we can only have one more.

I'd suggest the one with numbers, that seems like a common case:

```js
const baz = 0 ?? 42;
console.log(baz);
// expected output: 0
```

SphinxKnight:
Changed the second example for the suggestion with `0 ?? 42` and addressed the newline issue.

PR #1484, opened 2019-11-14
Atomics ES6 fix

velusgautam:
ES6 fix to change var to const in Atomics


velusgautam:
@wbamberg  Thanks for the feedback. I have made changes in the wiki as well.

PR #1493, opened 2019-11-25
Update intl-collator-prototype-compare.html

bkardell:
This example was incorrect. The name of the function is `firstAlphabetical` however, the conditional check and expectations were the inverse.

I added an example in english which shows one many reviewers will recognize as correct ('a' is first, not 'z').  My knowledge of other alphabets and collation rules are less good, but some searching and wikipedia seem to confirm that these are correct (and as I would have expected based on the incorrect comparator, simply reversed from the original examples)

### DIFF
--- a/live-examples/js-examples/intl/intl-collator-prototype-compare.html
+++ b/live-examples/js-examples/intl/intl-collator-prototype-compare.html
@@ -1,15 +1,19 @@
 <pre>
 <code id="static-js">function firstAlphabetical(locale, letter1, letter2) {
-  if (new Intl.Collator(locale).compare(letter1, letter2) > 0) {
+  if (new Intl.Collator(locale).compare(letter1, letter2) < 0) {
     return letter1;
   }
   return letter2;
 }
 
+console.log(firstAlphabetical('en', 'z', 'a'));
+// expected output: 'a'
+
 console.log(firstAlphabetical('de', 'z', 'ä'));
-// expected output: "z"
+// expected output: "ä"
 
 console.log(firstAlphabetical('sv', 'z', 'ä'));
-// expected output: "ä"
+// expected output: "z"
+
 </code>
 </pre>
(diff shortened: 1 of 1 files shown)


wbamberg:
It looks like you are right!

I also like the idea of including an 'en' example.

However, this makes the example a bit too long to fit in the editor without scrolling - this might seem pedantic but means 99% of people will never even know there is an example for 'sv', so the point of including both those locales is lost.

Also I think the way the actual `compare` function is buried after the `Collator` constructor call, inside a separate function, tends to obscure the way `compare` is called - when I was just looking at this it took me a while to understand that `locale` is not actually an argument to `compare`, because I was reading line 1 instead of line 2.

So what do you think about a simpler approach, like:

```js
const enCollator = new Intl.Collator('en');
const deCollator = new Intl.Collator('de');
const svCollator = new Intl.Collator('sv');

console.log(enCollator.compare('z', 'a') > 0);
// expected output: true

console.log(deCollator.compare('z', 'ä') > 0);
// expected output: true

console.log(svCollator.compare('z', 'ä') > 0);
// expected output: false
```

This seems to emphasise more clearly that compare is locale-sensitive.

bkardell:
> This code is little complex to understand for new people. Appreciate if you can simplify a bit.

Sorry, slightly confused: It seems my pull is exactly the same complexity as the existing example? It simply corrects a greater than to a less than.

Are you simply saying that you prefer @wbamberg's suggestion @velusgautam ?  I'm fine with that.  Shall I send a new pull or would you like to @wbamberg ?

wbamberg:
@bkardell , if you're OK doing so, and you agree that my suggestion is good, then please do update this PR, yes.

Otherwise I'll probably merge this one anyway, since the existing one is just wrong, and yours in the PR isn't :).

bkardell:
@wbamberg updated.

PR #1504, opened 2019-12-18
Adding regexp assertions

SphinxKnight:
As part of https://github.com/mdn/sprints/issues/1721, I'll be adding both inline / interactive examples.

A few notes about this one:

* Although lookbehind assertions are stage 4, they don't currently work in Fx and make the example slightly longer than 12 lines. I think it is better to include at least one of the "lookbehind" pattern but feel free to voice your objection

* This target a feature of the language but not a property or a method.

### DIFF
--- a/live-examples/js-examples/regexp/regexp-assertions.html
+++ b/live-examples/js-examples/regexp/regexp-assertions.html
@@ -1,15 +1,16 @@
 <pre>
-<code id="static-js">var text = "A quick fox";
-var regexpAhead = /[a-z]+ (?=fox)/;
-console.log(text.match(regexpAhead));
-// expected output [ 'quick ']
+<code id="static-js">const text = "A quick fox";
 
-var regexpNegativeAhead = /[A-Z]+ (?!slow)/
-console.log(text.match(regexpNegativeAhead));
-// expected output [ 'A ']
+const regexpLastWord = /\w+$/;
+console.log(text.match(regexpLastWord));
+// expected output: Array ["fox"]
 
-var regexpBehind = /(?&lt;=quick) \w+/;
-console.log(text.match(regexpBehind));
-// expected output (only in Chrome) for now [ 'fox']
+const regexpWords = /\b\w+\b/g;
+console.log(text.match(regexpWords));
+// expected output: Array ["A", "quick", "fox"]
+
+const regexpFoxQuality = /\w+(?= fox)/;
+console.log(text.match(regexpFoxQuality));
+// expected output: Array ["quick"]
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/regexp/regexp-assertions.html
```diff
@@ -0,0 +1,15 @@
+<pre>
+<code id="static-js">var text = "A quick fox";
```
wbamberg: We should use `const` and `let` in examples from now on.

wbamberg: Also, super-picky, but a blank line after this would help readability, IMO.

live-examples/js-examples/regexp/regexp-assertions.html
```diff
@@ -0,0 +1,15 @@
...
+var regexpAhead = /[a-z]+ (?=fox)/;
+console.log(text.match(regexpAhead));
+// expected output [ 'quick ']
```
wbamberg: We have a `:` here, and the console output is a little different:

`// expected output: Array ["quick "]`

I know this sounds super-picky, but if we ever use "expected output" to test the examples (which I would love, one day) this kind of consistency will be important.

live-examples/js-examples/regexp/regexp-assertions.html
```diff
@@ -0,0 +1,15 @@
...
+
+var regexpNegativeAhead = /[A-Z]+ (?!slow)/
+console.log(text.match(regexpNegativeAhead));
+// expected output [ 'A ']
+
+var regexpBehind = /(?&lt;=quick) \w+/;
+console.log(text.match(regexpBehind));
+// expected output (only in Chrome) for now [ 'fox']
```
wbamberg: I don't think we should include compat statements in the examples, they will get out of date. Just say "expected output" and hope people look at the compat data.

SphinxKnight:
> Then I'm a bit surprised not to see examples for the more familiar `^` or `$` assertions.

Well… I submitted this PR before the (now gone) 'Boundaries' article was merged with the article on assertions ^^: I concur that examples should illustrate them and I'll rewrite those

live-examples/js-examples/regexp/regexp-assertions.html
SphinxKnight: I agree but also need to state that when testing the code in Fx & Chrome consoles, things are not always consistent. I'm not sure what would be the best baseline and what are the details of the "console" implementation of the interactive example "engine".

PR #1505, opened 2019-12-19
Adding regexp character classes

SphinxKnight:
As part of https://github.com/mdn/sprints/issues/1721

I was tempted to add an example with emojis as I was unable to do so, please let me know if this is not a good idea, even in interactive-examples :s

### DIFF
--- a/live-examples/js-examples/regexp/regexp-character-classes.html
+++ b/live-examples/js-examples/regexp/regexp-character-classes.html
@@ -1,12 +1,12 @@
 <pre>
-<code id="static-js">var chessStory = "He played the King in a8 and she moved her Queen in c2.";
-var regexpCoordinates = /\w\d/g;
+<code id="static-js">const chessStory = "He played the King in a8 and she moved her Queen in c2.";
+const regexpCoordinates = /\w\d/g;
 console.log(chessStory.match(regexpCoordinates));
-// expected output [ 'a8', 'c2']
+// expected output: Array [ 'a8', 'c2']
 
-var moods = "happy 🙂, confused 😕, sad 😢";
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/regexp/regexp-character-classes.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">var chessStory = "He played the King in a8 and she moved her Queen in c2.";
```
wbamberg: We're using `const` and `let` in new examples now, and slowly updating existing ones.

live-examples/js-examples/regexp/regexp-character-classes.html
```diff
@@ -0,0 +1,12 @@
...
+var regexpCoordinates = /\w\d/g;
+console.log(chessStory.match(regexpCoordinates));
+// expected output [ 'a8', 'c2']
```
wbamberg:
```suggestion
// expected output: Array ["a8", "c2"]
```

live-examples/js-examples/regexp/regexp-character-classes.html
```diff
@@ -0,0 +1,12 @@
...
+
+var moods = "happy 🙂, confused 😕, sad 😢";
+var regexpEmoticons = /[\u{1F600}-\u{1F64F}]/gu;
+console.log(moods.match(regexpEmoticons));
+// expected output ['🙂', '😕', '😢']
```
wbamberg:
```suggestion
// expected output: Array ["🙂", "😕", "😢"]
```

SphinxKnight:
Thanks for the review Will :) (I'll be tackling the other PRs with var as well :))

PR #1506, opened 2019-12-19
Adding groups and ranges examples

SphinxKnight:
Adding another regexp example for https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges as part of https://github.com/mdn/sprints/issues/1721

### DIFF
--- a/live-examples/js-examples/regexp/meta.json
+++ b/live-examples/js-examples/regexp/meta.json
@@ -12,6 +12,12 @@
             "title": "JavaScript Demo: RegExp.getRegExp[@@species]",
             "type": "js"
         },
+        "regexpGroupsRanges": {
+            "exampleCode": "./live-examples/js-examples/regexp/regexp-groups-ranges.html",
+            "fileName": "regexp-groups-ranges.html",
+            "title": "JavaScript Demo: RegExp Groups and ranges",
+            "type": "js"
+        },
         "regexpLastIndex": {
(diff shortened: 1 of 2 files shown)


wbamberg:
Thanks @SphinxKnight ! I made some really minor wording changes and changes to the expected output.

I'm a little uncomfortable with the date one, because I worry someone will take it as an example of how to parse dates. But I can't think of a better example, so unless you can, or I think of one before tomorrow, I'm happy to merge this.

I think it is odd that https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Ranges treats groups and ranges together. I think of them as quite different things (ranges seem more like character classes than they do like groups).

SphinxKnight:
Agreed with date parsing, I provided another example which I hope is less risky.

Regarding groups vs. ranges, I don't have a strong feeling on this one but I think https://github.com/mdn/sprints/issues/1721 would be the place to voice this (on the other hand, this issue is getting longer so I'm not sure what's the most effective way to solve this).

SphinxKnight:
Agreed with date parsing, I provided another example which I hope is less risky.

Regarding groups vs. ranges, I don't have a strong feeling on this one but I think https://github.com/mdn/sprints/issues/1721 would be the place to voice this (on the other hand, this issue is getting longer so I'm not sure what's the most effective way to solve this).

PR #1507, opened 2019-12-20
Adding interactive examples for regexp quantifiers

SphinxKnight:
Another set of examples for regexps and https://github.com/mdn/sprints/issues/1721

### DIFF
--- a/live-examples/js-examples/regexp/regexp-quantifiers.html
+++ b/live-examples/js-examples/regexp/regexp-quantifiers.html
@@ -4,9 +4,13 @@ const regexpSpooky = /bo{3,}h/;
 console.log(ghostSpeak.match(regexpSpooky));
 // expected output: Array ["boooooooh"]
 
-let htmlSnippet = "&lt;span foo='bar' baz&gt;Hello&lt;span/&gt;"
-const regexpSpan = /&lt;span.*?&gt;|&lt;span\/&gt;/g;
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/regexp/regexp-quantifiers.html
```diff
@@ -0,0 +1,12 @@
+<pre>
+<code id="static-js">const ghostSpeak = "booh boooooooh";
+const regexpSpooky = /bo{3,}h/;
+console.log(ghostSpeak.match(regexpSpooky));
+// expected output: Array ["boooooooh"]
+
+let htmlSnippet = "&lt;span foo='bar' baz&gt;Hello&lt;span/&gt;"
```
wbamberg: let htmlSnippet = "&lt;span foo='bar' baz&gt;Hello&lt;span/&gt;"
```suggestion
const htmlSnippet = "&lt;span foo='bar' baz&gt;Hello&lt;span/&gt;";
```

SphinxKnight: Agreed with this. I was too quick to apply my "mdn-linting-fixing with regexp" own procedure to create an example. I'll figure out another one :)

SphinxKnight:
@wbamberg tried another example (I had a hard time finding ^ ^)  and resolved the conflict on meta.json

PR #1509, opened 2019-12-23
addded Date.prototype[@@toPrimitive] example

KRONOS-11:
Added the date toPrimitive example to the interactive examples page, this is my first contribution on github , so feel free to criticize.

### DIFF
--- a/live-examples/js-examples/date/date-toprimitive.html
+++ b/live-examples/js-examples/date/date-toprimitive.html
@@ -1,13 +1,11 @@
 <pre>
-<code id="static-js" class="language-js">
-  const date = new Date('20 December 2019 14:48'); //initialize a new Date object
+<code id="static-js">// Depending on timezone, your results will vary
+const date = new Date('20 December 2019 14:48');
 
-  console.log(date[Symbol.toPrimitive]('string')); //log primitive of date with hint as string
-  //expected output: "Fri Dec 20 2019 20:18:00 GMT+0530 (India Standard Time)"
-  //note: the timezone specific data(GMT+0530 (India Standard Time)) may vary according to your location
+console.log(date[Symbol.toPrimitive]('string'));
+// expected output: "Fri Dec 20 2019 20:18:00 GMT+0530 (India Standard Time)"
 
-  console.log(date[Symbol.toPrimitive]('number')); //log primitive of date with hint as number
-  //expected output: 1576853280000 ,this may also vary
-
+console.log(date[Symbol.toPrimitive]('number'));
+// expected output: 1576853280000
 </code>
 </pre>
--- a/live-examples/js-examples/date/date-toprimitive.html
+++ b/live-examples/js-examples/date/date-toprimitive.html
@@ -1,0 +1,13 @@
+<pre>
+<code id="static-js" class="language-js">
+  const date = new Date('20 December 2019 14:48'); //initialize a new Date object
+
+  console.log(date[Symbol.toPrimitive]('string')); //log primitive of date with hint as string
+  //expected output: "Fri Dec 20 2019 20:18:00 GMT+0530 (India Standard Time)"
+  //note: the timezone specific data(GMT+0530 (India Standard Time)) may vary according to your location
+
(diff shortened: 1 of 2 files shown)


KRONOS-11:
@wbamberg

live-examples/js-examples/date/date-toprimitive.html
```diff
@@ -0,0 +1,13 @@
+<pre>
+<code id="static-js" class="language-js">
```
wbamberg: We don't need the `class="language-js"` attribute.

Also, having a new line after this opening `code` tag will make the example start with a blank line, and we'd prefer not to do that.

So this line should be like:

`<code id="static-js">const date = new Date('20 December 2019 14:48');`

live-examples/js-examples/date/date-toprimitive.html
```diff
@@ -0,0 +1,13 @@
...
+  const date = new Date('20 December 2019 14:48'); //initialize a new Date object
```
wbamberg: All the lines have an indentation, which they should not have.

Also we don't generally use generic comments saying what each line it doing. We aim to write simple examples that can speak for themselves, and only use comments sparingly to point out particularly nonobvious things.

live-examples/js-examples/date/date-toprimitive.html
```diff
@@ -0,0 +1,13 @@
...
+
+  console.log(date[Symbol.toPrimitive]('string')); //log primitive of date with hint as string
+  //expected output: "Fri Dec 20 2019 20:18:00 GMT+0530 (India Standard Time)"
```
wbamberg: There should be a space between `//` and the comment:

`// expected output: "Fri Dec 20 2019 20:18:00 GMT+0530 (India Standard Time)"`

live-examples/js-examples/date/date-toprimitive.html
```diff
@@ -0,0 +1,13 @@
...
+  //note: the timezone specific data(GMT+0530 (India Standard Time)) may vary according to your location
```
wbamberg: Actually the whole thing may vary, so it would be better to say:

`// timezone specific data may vary according to your local settings`

live-examples/js-examples/date/date-toprimitive.html
```diff
@@ -0,0 +1,13 @@
...
+
+  console.log(date[Symbol.toPrimitive]('number')); //log primitive of date with hint as number
+  //expected output: 1576853280000 ,this may also vary
```
wbamberg: rather than ` ,this may also vary` it would be better to have this on a separate line and restate the issue:

`// timezone specific data may vary according to your local settings`

live-examples/js-examples/date/date-toprimitive.html
KRONOS-11: I don't think the hours and minutes will vary because, i haven't specified timezone in the date declaration `'20 December 2019 14:48 U̶T̶C̶'` so it will be inferred by the runtime according to the locale specifications, and what do you mean by local settings ?  Thanks.

KRONOS-11:
Hey @wbamberg , i've committed the changes you requested. please review it again. Thanks for the review.

live-examples/js-examples/date/date-toprimitive.html
wbamberg: I'm in Pacific time, and I do see hours and minutes that differ from `expected output`:

Screen Shot 2020-01-07 at 8 25 35 AM

So perhaps the expected output is incorrect?

KRONOS-11: yep the expected output is incorrect I'm getting the following result.
blob:null/472b3468-0753-fe44-bcad-544c9ba8f324

KRONOS-11: fixed it.

KRONOS-11:
Thanks @wbamberg for your help and review, love you.

wbamberg:
You're welcome @KRONOS-11 , thanks for your contribution!

PR #1510, opened 2019-12-23
Const to var

BenoitAdam94:
I'm sorry but the Const is totally out of context for an Array. It must be a var.

On Chrome and Firefox I got 4 7 7 instead of what is expected : 4 4 7

### DIFF
--- a/live-examples/js-examples/array/array-push.html
+++ b/live-examples/js-examples/array/array-push.html
@@ -1,7 +1,7 @@
 <pre>
-<code id="static-js">const animals = ['pigs', 'goats', 'sheep'];
+<code id="static-js">var animals = ['pigs', 'goats', 'sheep'];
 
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for your PR!

> I'm sorry but the Const is totally out of context for an Array. It must be a var.

Can you explain what "out of context" means here? And why must it be `var`?

> On Chrome and Firefox I got 4 7 7 instead of what is expected : 4 4 7

On Chrome and Firefox I get this output:

```
> 4
> Array ["pigs", "goats", "sheep", "cows"]
> Array ["pigs", "goats", "sheep", "cows", "chickens", "cats", "dogs"]
```

...which is expected.

wbamberg:
Closing, because I don't think this PR is valid.

Issue #464, opened 2018-01-25
Add "taller" to examples that need it


wbamberg:
Now https://github.com/mdn/interactive-examples/issues/408 is landed, we should add the "taller" class to examples that need it.

wbamberg:
What's involved here:

* get a list of JS examples that are more than 13 lines long
* do some manual analysis to validate that we actually want to add "taller" to all these examples (this means, for example, that if 90% of them are at 14 lines, we might want to reconsider)
* once we have the final set of examples that need to be "taller", add the class both in the example source and in the macro call

wbamberg:
If my analysis is right, here's the distribution of the sizes of JS examples:

```
Example length  :  Number of examples
-------------------------------------
        4       :  13
        5       :  20
        6       :  24
        7       :  12
        8       :  56
        9       :  30
       10       :  38
       11       :  91
       12       :  96
       13       :  14
       14       :   8
       15       :   2
       16       :   1
       17       :   3
       18       :   1
       19       :   1
       20       :   1
       21       :   3
       22       :   2
       23       :   1
       29       :   1
```

wbamberg:
And these are the 24 examples with a length > 13:

```
weakset-prototype-add
number-isnan
regexp-prototype-multiline
reflect-isextensible
regexp-prototype-unicode
reflect-deleteproperty
regexp-prototype-tostring
function-bind
classes-extends
regexp-prototype-test
statement-async
proxyhandler-getownpropertydescriptor
proxyhandler-construct
proxyhandler-apply
proxyhandler-defineproperty
proxyhandler-ownkeys
proxyhandler-setprototypeof
proxyhandler-set
proxyhandler-preventextensions
proxyhandler-getprototypeof
proxyhandler-get
proxyhandler-deleteproperty
proxyhandler-has
proxyhandler-isextensible
```

wbamberg:
I've also updated all the MDN pages to use the "taller" parameter, so this should be finished now.

PR #1511, opened 2019-12-24
Replace example for string.normalize()

wbamberg:
As part of https://github.com/mdn/sprints/issues/2594 I decided that the interactive example for [`normalize()`](https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize ) is not very readable: these long lines assembling complex strings are hard to follow.

So this PR rewrites it. It only covers "NFD" and "NFC", but is already as long as I'd like an example to be.

### DIFF
--- a/live-examples/js-examples/string/string-normalize.html
+++ b/live-examples/js-examples/string/string-normalize.html
@@ -1,17 +1,25 @@
 <pre>
-<code id="static-js">var first = '\u212B';         // "Å"
-var second = '\u0041\u030A';  // "Å"
+<code id="static-js" data-height="taller">const name1 = '\u0041\u006d\u00e9\u006c\u0069\u0065';
+const name2 = '\u0041\u006d\u0065\u0301\u006c\u0069\u0065';
 
-console.log(first + ' and ' + second + ' are' +
-   ((first === second)? '': ' not') + ' the same.');
-   // expected output: "Å and Å are not the same."
+console.log(`${name1}(${name1.length})`, `${name2}(${name2.length})`);
+// expected output: "Amélie(6)" "Amélie(7)"
+console.log(name1 === name2);
+// expected output: false
 
-console.log(first + ' and ' + second + ' can' +
-   ((first.normalize('NFC') === second.normalize('NFC'))? '': ' not') + ' be normalized');
-   // expected output: "Å and Å can be normalized"
+const name1NFC = name1.normalize('NFC');
+const name2NFC = name2.normalize('NFC');
 
-var oldWord = 'mañana';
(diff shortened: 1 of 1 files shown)


wbamberg:
@chrisdavidmills , please could you review this when you get the time?

wbamberg:
> So, I'm approving this, but I do wonder whether this is a bit too long to be an ie at the top of the page? Couldn't you just remove the last bit that uses NFD, and just stick with the NFC stuff? I feel that this on its own shows a good usage of normalize(), and your in-page examples that I reviewed previously expand on it nicely and show the difference between NFC and NFD.

I think showing just NFC is a good idea, just to make the example look simpler and less intimidating. But unfortunately even just showing NFC is 15 lines, which is just over the limit for a standard-size editor. And it's hard to think of how to compress it further without losing useful information. So removing NFD won't help with the sizing issue.

> If that isn't something you want to do, perhaps you could show a smaller code window and just accept that people will have to scroll to see all the code?

This is going to be tricky since https://github.com/mdn/bob/pull/368 landed, because that automatically selects the taller window based on the actual example size.

So options include:

1. do nothing

2. update https://github.com/mdn/bob/pull/368 so we can manually set code window size, as we used to, perhaps to override the automatically selected one

3. ditch the whole idea of "taller" and just use a single size, and let it overflow

Since (2) is extra work (and more to maintain, and more for contributors to understand) then we should only do it if *some* longer examples should be allowed to overflow, while others should get a taller window. I'm not sure how we should make this determination really. For example, why should this example be allowed to overflow, but [`Proxy.handler.preventExtensions()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/preventExtensions) should not?

Which would leave (3), but my belief, based the way I saw people interacting with these things in user testing, is that users you need to scroll to see some of the output, then almost noone will do it.

As a possible consolation, there aren't very many examples that overflow. This analysis: https://github.com/mdn/interactive-examples/issues/464#issuecomment-360572522 is old but still probably roughly accurate (I'm going to repeat it after https://github.com/mdn/bob/pull/368 permeates) and that indicated that only 24 examples had a length > 13, which gets you the taller editor.

wbamberg:
I've removed the "NFD" variant, for simplicity, and also simplified the output strings, as without NFD there didn't seem much value in listing the exact string lengths.

chrisdavidmills:
That looks better to me Will. About the concerns I raised earlier, I think the best thing for now is the "do nothing" option, as this isn't broken by any means, and we have higher priority things for you to be working on right now.

wbamberg:
Thanks Chris!

PR #1513, opened 2019-12-26
es6 changes for promise and reflect properties

velusgautam:
ES6 update for reflect and promise files in js-examples

### DIFF
--- a/live-examples/js-examples/promise/promise-constructor.html
+++ b/live-examples/js-examples/promise/promise-constructor.html
@@ -11,7 +11,7 @@ promise1.then(function(value) {
 });
 
 console.log(promise1);
-// expected output: Promise { <state>: "pending" }
+// expected output: [object Promise]
 
 </code>
 </pre>
--- a/live-examples/js-examples/promise/promise-constructor.html
+++ b/live-examples/js-examples/promise/promise-constructor.html
@@ -1,5 +1,5 @@
 <pre>
-<code id="static-js">var promise1 = new Promise(function(resolve, reject) {
+<code id="static-js">const promise1 = new Promise(function(resolve, reject) {
   setTimeout(function() {
     resolve('foo');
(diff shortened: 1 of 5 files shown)


velusgautam:
@wbamberg Can you please verify and merge

live-examples/js-examples/reflect/reflect-deleteproperty.html
```diff
@@ -12,6 +12,6 @@
 Reflect.deleteProperty(array1, '3');
 
 console.log(array1);
-// expected output: Array [1, 2, 3, , 5]
+// expected output: Array(5) [ 1, 2, 3, undefined, 5 ]
```
wbamberg: Again, this comment change is not quite accurate - the result shown in the interactive example's console here is: `Array [1, 2, 3, undefined, 5]`.

live-examples/js-examples/promise/promise-constructor.html
```diff
@@ -11,7 +11,7 @@
 });
 
 console.log(promise1);
-// expected output: [object Promise]
+// expected output: Promise { <state>: "pending" }
```
wbamberg: The value for `expected output` should be the result shown in the interactive example console, not in the browser's JS console. In this case the example displays the array as: `[object Promise]`, so this comment change is not quite accurate.

(Ideally we would have tests for this, but that's quite hard.)

velusgautam: I have made the changes as requested. Thank you for the feedback

live-examples/js-examples/reflect/reflect-deleteproperty.html
velusgautam: I have made the changes as requested.

velusgautam:
I have made the changes you have requested. Please have a check @wbamberg

PR #1517, opened 2020-01-06
Un-`var`-ify (most) JS examples

Zearin:
Didn’t touch examples for the `var` keyword itself, or in examples demonstrating special scope rules.

### DIFF
--- a/live-examples/js-examples/map/map-prototype-entries.html
+++ b/live-examples/js-examples/map/map-prototype-entries.html
@@ -1,10 +1,11 @@
 <pre>
-<code id="static-js">var map1 = new Map();
+<code id="static-js">
+const map1 = new Map();
 
 map1.set('0', 'foo');
 map1.set(1, 'bar');
 
-var iterator1 = map1.entries();
+const iterator1 = map1.entries();
 
 console.log(iterator1.next().value);
 // expected output: ["0", "foo"]
--- a/live-examples/js-examples/map/map-prototype-entries.html
+++ b/live-examples/js-examples/map/map-prototype-entries.html
@@ -1,6 +1,5 @@
 <pre>
-<code id="static-js">
-const map1 = new Map();
+<code id="static-js">const map1 = new Map();
 
 map1.set('0', 'foo');
 map1.set(1, 'bar');
--- a/live-examples/js-examples/number/number-parseint.html
+++ b/live-examples/js-examples/number/number-parseint.html
@@ -1,6 +1,7 @@
 <pre>
-<code id="static-js">function roughScale(x, base) {
-  var parsed = Number.parseInt(x, base);
+<code id="static-js">
+function roughScale(x, base) {
+  let parsed = Number.parseInt(x, base);
   if (Number.isNaN(parsed)) {
     return 0;
   }
--- a/live-examples/js-examples/number/number-parseint.html
+++ b/live-examples/js-examples/number/number-parseint.html
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">function roughScale(x, base) {
-  let parsed = Number.parseInt(x, base);
+  const parsed = Number.parseInt(x, base);
   if (Number.isNaN(parsed)) {
     return 0;
   }
--- a/live-examples/js-examples/number/number-parseint.html
+++ b/live-examples/js-examples/number/number-parseint.html
@@ -1,6 +1,5 @@
 <pre>
-<code id="static-js">
-function roughScale(x, base) {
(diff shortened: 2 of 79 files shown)


live-examples/js-examples/map/map-prototype-entries.html
```diff
@@ -1,10 +1,11 @@
 <pre>
-<code id="static-js">var map1 = new Map();
+<code id="static-js">
```
wbamberg: Adding the return after the opening `<code>` tag will add an extra blank line at the start of the example, which we don't want. I know this is quite non-obvious and when https://github.com/mdn/bob/pull/368 lands the problem will go away, but for now we should avoid it.

Zearin: I was afraid of that. :)  

Thanks.  Fixed.

live-examples/js-examples/number/number-parseint.html
```diff
@@ -1,6 +1,6 @@
 <pre>
 <code id="static-js">function roughScale(x, base) {
-  var parsed = Number.parseInt(x, base);
+  let parsed = Number.parseInt(x, base);
```
wbamberg: This could be `const`.

live-examples/js-examples/statement/statement-empty.html
```diff
@@ -1,8 +1,8 @@
 <pre>
-<code id="static-js">var array1 = [1, 2, 3];
+<code id="static-js">let array1 = [1, 2, 3];
```
wbamberg: This could be `const`.

live-examples/js-examples/statement/statement-functionasterisk.html
```diff
@@ -4,7 +4,7 @@
   yield i + 10;
 }
 
-var gen = generator(10);
+let gen = generator(10);
```
wbamberg: This could be `const`.

live-examples/js-examples/symbol/symbol-split.html
```diff
@@ -4,7 +4,7 @@
     this.value = value;
   }
   [Symbol.split](string) {
-    var index = string.indexOf(this.value);
+    let index = string.indexOf(this.value);
```
wbamberg: This could be `const`.

package-lock.json
```diff
@@ -5291,7 +5291,7 @@
             "integrity": "sha1-VAVy0096wxGfj3bDDLwbHgN6/74=",
             "dev": true,
             "requires": {
-                "builtin-modules": "1.1.1"
+                "builtin-modules": "^1.0.0"
```
wbamberg: Can you please revert this package-lock change?

live-examples/js-examples/reflect/reflect-get.html
```diff
@@ -7,7 +7,7 @@
 console.log(Reflect.get(object1, 'x'));
 // expected output: 1
 
-var array1 = ['zero', 'one'];
+let array1 = ['zero', 'one'];
```
velusgautam: Here also const will work. @wbamberg what do you think?

live-examples/js-examples/reflect/reflect-deleteproperty.html
```diff
@@ -8,7 +8,7 @@
 console.log(object1.property1);
 // expected output: undefined
 
-var array1 = [1, 2, 3, 4, 5];
+let array1 = [1, 2, 3, 4, 5];
```
velusgautam: Here also const will work. @wbamberg what do you think?

live-examples/js-examples/set/set-prototype-values.html
```diff
@@ -3,7 +3,7 @@
 set1.add(42);
 set1.add('forty two');
 
-var iterator1 = set1.values();
+let iterator1 = set1.values();
```
velusgautam: Here also const will work. @wbamberg what do you think?

live-examples/js-examples/reflect/reflect-get.html
wbamberg: Agreed.

live-examples/js-examples/reflect/reflect-deleteproperty.html
wbamberg: Agreed.

live-examples/js-examples/set/set-prototype-values.html
wbamberg: Agreed.

Zearin:
Thanks, @wbamberg !  Feels great to have a Mozilla merge under my belt.  :)

PR #1519, opened 2020-01-08
Replace h1 with h2 since its used more than once

participator:
- Promote using h1 once to avoid encouraging bad html practices
- > You should consider avoiding using `<h1>` more than once on a page. See Defining sections for more information.
> [Heading Elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)

### DIFF
--- a/live-examples/html-examples/content-sectioning/section.html
+++ b/live-examples/html-examples/content-sectioning/section.html
@@ -1,9 +1,10 @@
+<h1>How to Fish</h1>
 <section>
-    <h1>Introduction</h1>
+    <h2>Introduction</h2>
     <p>People have been catching fish for food since before recorded history…</p>
 </section>
 
 <section>
-    <h1>Equipment</h1>
+    <h2>Equipment</h2>
     <p>The first thing you’ll need is a fishing rod or pole that you find comfortable and is strong enough for the kind of fish you’re expecting to land…</p>
 </section>


wbamberg:
The example here is for the `<section>` page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section.

It's my understanding that this example is fine: it shows multiple `<section>` elements, each with an `<h1>`. There is a similar example in the [spec](https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements):

```html
<body>
 <h1>Let's call it a draw(ing surface)</h1>
 <section>
  <h1>Diving in</h1>
 </section>
 <section>
  <h1>Simple shapes</h1>
 </section>
 <section>
  <h1>Canvas coordinates</h1>
  <section>
   <h1>Canvas coordinates diagram</h1>
  </section>
 </section>
 <section>
  <h1>Paths</h1>
 </section>
</body>
```

It's actually quite unclear to me whether it's OK to have multiple `<h1>` elements where they are not in separate sectional elements. And our docs are unhelpful here. Saying *"You should consider avoiding using `<h1>` more than once on a page"* is totally unhelpful without explaining what the considerations are and when they apply. Saying *"See Defining sections for more information"* is totally unhelpful when the linked section gives no relevant guidance on this point.

So I think this should be a content bug to clarify this situation. If it turns out after that that this example is wrong, we can come back to it. @chrisdavidmills , what do you think?

participator:
The examples showing multiple `<h1>` elements on a single page go against good accessible and semantic practices.

For examples, please refer to:
[Accessible Headings](https://webaccess.berkeley.edu/resources/tips/web-accessibility#accessible-headings)
[HTML Sections & Outlines](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines) *Doesn't directly speak against using multiple `<h1>` elements, but has good examples

participator:
@wbamberg and @chrisdavidmills #2658 is a really clear update and thanks for merging my PR!

PR #1520, opened 2020-01-11
Added example for toLocaleDateString(undefined, options)

Irfan434:
This is a common use case, but doesn't have an example currently.

### DIFF
--- a/live-examples/js-examples/date/date-tolocaledatestring.html
+++ b/live-examples/js-examples/date/date-tolocaledatestring.html
@@ -9,12 +9,7 @@ console.log(event.toLocaleDateString('de-DE', options));
 console.log(event.toLocaleDateString('ar-EG', options));
 // expected output: الخميس، ٢٠ ديسمبر، ٢٠١٢
 
-console.log(event.toLocaleDateString('ko-KR', options));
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/date/date-tolocaledatestring.html
```diff
@@ -11,5 +11,10 @@
 
 console.log(event.toLocaleDateString('ko-KR', options));
 // expected output: 2012년 12월 20일 목요일
+
+// Pass undefined as the first argument to use the browser's default locale
```
wbamberg: Please remove this line: we don't generally use descriptive comments in the examples unless there's something very unusual to point out.

live-examples/js-examples/date/date-tolocaledatestring.html
```diff
@@ -11,5 +11,10 @@
...
+console.log(event.toLocaleDateString(undefined, options));
+// expected output (when default locale is en-US): Thursday, December 20, 2012
```
wbamberg:
```suggestion
// expected output: Thursday, December 20, 2012 (varies according to default locale)
```

PR #1522, opened 2020-01-13
Added explicit "true" to contenteditable attribute

PG-8:
From https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable 

"This attribute is an enumerated one and not a Boolean one. This means that the explicit usage of one of the values true, false or the empty string is mandatory and that a **shorthand like `<label contenteditable>Example Label</label>` is not allowed**. The correct usage is `<label contenteditable="true">Example Label</label>`."

Addition of **`="true"`** to the `contenteditable` attribute improves the example in question as it previously used the "not allowed" shorthand.

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-spellcheck.html
+++ b/live-examples/html-examples/global-attributes/attribute-spellcheck.html
@@ -1,3 +1,3 @@
-<p contenteditable spellcheck="true">This exampull will be checkd fur spellung when you try to edit it.</p>
+<p contenteditable="true" spellcheck="true">This exampull will be checkd fur spellung when you try to edit it.</p>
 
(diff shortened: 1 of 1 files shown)


wbamberg:
This has come up before: https://github.com/mdn/interactive-examples/pull/1333. I think our docs for `contenteditable` are wrong, and I filed https://github.com/mdn/sprints/issues/1596 to fix them last time. We should really fix the docs.

wbamberg:
I've clarified the docs over in https://github.com/mdn/sprints/issues/1596, so am closing this. Thanks for your PR anyway (and for giving us the impetus to fix the issue!).

Issue #250, opened 2017-08-31
Add a linter for JS examples


wbamberg:
As Mark suggested, having a linter for the JS examples especially would really help guarantee a consistent style.

schalkneethling:
@wbamberg I am guessing here you are referring to linting the JS that is contributed i.e. everything inside the `code` block ~ https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/array-concat.html#L2

wbamberg:
@schalkneethling , yes, exactly.

schalkneethling:
![](https://github.trello.services/images/mini-trello-icon.png) [Add a linter for JS examples](https://trello.com/c/vL6HpReS/1-add-a-linter-for-js-examples)

7ayushgupta:
I'd like to try working on this issue. Can you please guide me on how to proceed?

wbamberg:
That would be great! @schalkneethling can give you better guidance than me on the best way to do it.

7ayushgupta:
@schalkneethling, can you please help me on how to begin and what all to do?

schalkneethling:
@7ayushgupta I will get back to you as soon as possible with some ideas. Basically I was first thinking to make it part of the main page builder Nodejs script here[1] but, after thinking about it a bit more, I am thinking a standalone script utility would be a better way forward.

These two can then be run independently or together, which offers us more flexibility. Do you have experience with writing Nodejs utilities that does file parsing etc?

[1] https://github.com/mdn/interactive-examples/blob/master/index.js

schalkneethling:
@7ayushgupta This one is pretty high on our list. I am wondering, perhaps one of the items on this list might be of interest to you? https://github.com/mdn/interactive-examples/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Agood-first-bug

7ayushgupta:
Would you be using scripts like JSLint or write the linting code yourself?

schalkneethling:
I will be using the `eslint` Nodejs API

wbamberg:
See also https://github.com/mdn/interactive-examples/pull/1470.

ikarasz:
[...]

PR #1528, opened 2020-01-15
Replace some .html example sources with .js

ikarasz:
With this PR I would like to replace some `.html` source files with `.js` files as the next step in #250.

### DIFF
--- a/live-examples/js-examples/classes/classes-extends.js
+++ b/live-examples/js-examples/classes/classes-extends.js
@@ -1,5 +1,4 @@
-<pre>
-<code id="static-js" data-height="taller">class formatDate extends Date {
+class formatDate extends Date {
   constructor(dateStr) {
     super(dateStr);
   }
@@ -14,5 +13,3 @@
 
 console.log(new formatDate('August 19, 1975 23:15:30').getFormattedDate());
 // expected output: "19-Aug-1975"
-</code>
-</pre>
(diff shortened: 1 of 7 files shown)


ikarasz:
@wbamberg, I've checked the result and I'm thinking about to reduce the height of the shorter examples. Since we reduced the limit to 7 lines it looks a bit strange to me with very short examples like some of the date related ones. What do you think?

wbamberg:
> @wbamberg, I've checked the result and I'm thinking about to reduce the height of the shorter examples. Since we reduced the limit to 7 lines it looks a bit strange to me with very short examples like some of the date related ones. What do you think?

Sorry, I'm not sure I understand. Do you mean, we should make the actual iframe shorter when the example length is <= 7 lines? Do you have an example of one that looks strange?

For example, the [`Date.getUTCSeconds`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCSeconds) example live on the site currently looks like:

Screen Shot 2020-01-15 at 5 19 28 PM

As far as I can tell this looks the same as the version in this PR. So maybe we should reduce the height in the "shorter" case, but that seems to be a separate discussion. Even if we did... well, the "shorter" size has to work for examples of 7 lines. It's nice to have a blank line at the end of the example. So we would be looking at a length of 8 lines. We currently have a length of 9 lines, so it's only a small amount we'd be able to remove. Anyway, I'm totally happy to have that conversation, but it seems separate to this effort.

Or do I misunderstand you?

ikarasz:
Thank you @wbamberg, also I made an issue about my height-related question #1530

PR #1537, opened 2020-01-21
longer cache for all static assets

### LINKED ISSUE #1536: Interactive example static assets aren't very cached
For example, a page [like this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) uses an [iframe like this](https://interactive-examples.mdn.mozilla.net/pages/js/array-concat.html) which has a `cache-control: max-age=1800` which I think is fine (1800==30min). 

However, there are a bunch of other assets could be cached longer:

These are `max-age=1800`:
`https://interactive-examples.mdn.mozilla.net/css/editor-js.css?v=f6bdadc` 
`https://interactive-examples.mdn.mozilla.net/js/editor-js.js?v=f6bdadc` 
`https://interactive-examples.mdn.mozilla.net/js/editor-tabbed.js?v=f6bdadc`

Curiously, these are already `max-age=2629800`:
`https://interactive-examples.mdn.mozilla.net/css/codemirror-5-50-2.css?v=f6bdadc`
`https://interactive-examples.mdn.mozilla.net/js/codemirror-5-50-2.js?v=f6bdadc`
`https://interactive-examples.mdn.mozilla.net/media/dino.svg`

(Note, these listed example URLs are the only ones. )

peterbe:
Fixes #1536

### DIFF
--- a/bin/s3-sync.sh
+++ b/bin/s3-sync.sh
@@ -15,8 +15,7 @@ for path in pages live-examples; do
 done
 
 for path in css js; do
-    aws s3 sync $path s3://$1/$path --exclude '*' --include 'editor-*' ${SHORT_CACHE} ${ARGS}
-    aws s3 sync $path s3://$1/$path --exclude 'editor-*' ${LONG_CACHE} ${ARGS}
+    aws s3 sync $path s3://$1/$path ${LONG_CACHE} ${ARGS}
 done
 
 aws s3 sync media s3://$1/media ${LONG_CACHE} ${ARGS}


peterbe:
This is actually tricky now. If you `curl` something like `https://interactive-examples.mdn.mozilla.net/js/editor-tabbed.js?v=f6bdadc` you'll get:
```
x-cache: Hit from cloudfront
```
which makes sense. But suppose you change the `?v=....` thing. E.g `https://interactive-examples.mdn.mozilla.net/js/editor-tabbed.js?v=f6bdadc0123456789` and `curl` that one you get:
```
x-cache: Hit from cloudfront
```
again :(

Meaning, our CloudFront config drops all query string before doing the lookup to S3. 

The more robust solution is for `interactive-examples` to NOT use `?query=strings` as a solution to make URLs always unique. But `bob` is NOT ready for that because of how it works. 

Perhaps we just forget this and hope that we can do `interactive-examples` properly when we find the time to refactor the platform for these things.

peterbe:
The reason the CDN gives the same `x-cache` independent of the query string is because of the way our CloudFront is configured. 
If we just allow query strings to matter as part of the URL, it would be possible to proceed with this PR. 

CC @escattone @limed

schalkneethling:
Thank you for the PR @peterbe. I would like to get @wbamberg's input here as well. I believe one of the reasons we had a short cache time was because it allowed us to ship things to production quickly in the early days. Also, we could fix breaking bugs quickly :)

Not sure if this is still the case now so, perhaps it is time to increase the overall cache time. Thoughts Will?

peterbe:
Ideally, the URLs should have hashes in the *name* but we don't. We will when we replace `bob` with Yari or whatever our next-gen solution is. 
The saving grace is that these files, that I want to cache longer, are very very rarely updated. 
And suppose they are, at least the *browser* will respect the `?v=hash` solution.

wbamberg:
As long as we use the short cache for all the content (that is, everything that is built from https://github.com/mdn/interactive-examples/tree/master/live-examples, including CSS and JS of course), then I'm fine with this change.

peterbe:
So @schalkneethling approve and merge?

PR #1542, opened 2020-02-19
Use JS source files for js examples

ikarasz:
Replace `html` source files in 3 more folders


ikarasz:
@wbamberg do you mind if I change all the remaining files in one PR? It will be pretty big so I am okay with smaller chunks as well

wbamberg:
@ikarasz , I agree, all in one big PR would be best at this point. It seems like this approach is properly proven now :).

Issue #573, opened 2018-02-11
CSS examples: need example for font-stretch


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch

Syntactically this looks quite simple. The only tricky bit is finding a font that supports the stretched and condensed variants.

For consistency it would be good to base the example on one of the existing font examples:
https://developer.mozilla.org/en-US/docs/Web/CSS/font-family
https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight
etc.

akoliyot:
@wbamberg Can I help with this?

wbamberg:
You certainly can @akoliyot ! Please take a look at the [contribution docs](https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING.md), and ask if you have any questions. Thanks!

akoliyot:
@wbamberg Sorry, I haven't been able to find any fonts that would work with the font-stretch property. 

I found this: 
http://koe.berlin/variablefont/
But for now it's only working when setting the width like this: 
font-variation-settings: "wdth" 10;

I'll keep looking and update as soon as I find something.

AreaOfAKite:
I will research applicable fonts.

wbamberg:
@AreaOfAKite , I went through this a little while ago when updating the [`font-stretch`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch) page, and found [League Mono Variable](http://tylerfinck.com/leaguemonovariable/), which works very nicely :).

ashvaibhav:
Looks like this issue can be closed now.
Examples are added in https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch#Font_face_selection

wbamberg:
> Looks like this issue can be closed now.

I don't think so: this issue is to add interactive examples, and there are no interactive examples for `font-stretch` in that page or in this repo.

pillig3:
Gave this a try, it's working in Chrome (79.0) but not Firefox (73.0) nor Safari (12.1.2).
Those should support font-stretch; can anyone tell what's wrong with the code?

- https://github.com/mdn/interactive-examples/compare/master...pillig3:master

wbamberg:
@pillig3 , thanks for your contribution! Your example works fine for me in both Firefox (73 and Nightly) and Safari (13). In what way is it not working for you? Do you get any console errors?

[...]

PR #1545, opened 2020-02-23
CSS example for font-stretch

pillig3:
For issue [573](https://github.com/mdn/interactive-examples/issues/573)


wbamberg:
Sorry to be slow reviewing this @pillig3 . I will get to it soon I promise.

pillig3:
@wbamberg  no worries!

live-examples/css-examples/fonts/font-stretch.css
```diff
@@ -0,0 +1,13 @@
+@font-face {
+    font-family: 'LeagueMonoVariable';
+    src: local('LeagueMonoVariable'),
+         url('../../../media/fonts/LeagueMonoVariable.ttf') format('truetype');
```
wbamberg: I think the url can just be '/media/fonts/LeagueMonoVariable.ttf', can't it? That's how we are referencing fonts in the other `@font-face` rules, like https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/fonts/font-variation-settings.css.

live-examples/css-examples/fonts/font-stretch.html
```diff
@@ -0,0 +1,48 @@
...
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+</section>
+
+<div id="output" class="output hidden">
+    <section id="default-example" class="default-example">
+      <div id="example-element">
```
wbamberg: The other font examples use `<p>` here, which seems more semantic than a `<div>`. Is there a reason you chose a `<div>`?

live-examples/css-examples/fonts/font-stretch.html
```diff
@@ -0,0 +1,48 @@
...
+          an elephantine lizard
```
wbamberg: I think it would be nice for the sample text to be the same as for the [font-variation-settings](https://github.com/mdn/interactive-examples/blob/master/live-examples/css-examples/fonts/font-variation-settings.html#L35) example.

I wondered about using the same font as well, then we couldn't have to add a new one here. But the League Mono seems to be stretchier, and it's the same one as in the body of the page, so perhaps it's a better choice after all.

wbamberg:
Closing this, since it seems to be inactive.

PR #1546, opened 2020-03-03
Replace all the remaining js examples which are in html

ikarasz:
@wbamberg some `meta.json` had `crlf` line endings, I have replaced them now based on the `.editorconfig`. If this change is already too big, I can create another pr for the line endings.


wbamberg:
\o/

PR #1548, opened 2020-03-10
Update contribution docs now we supply JS examples in .js files

wbamberg:
This PR updates our contribution docs now we:

* provide JS examples in .js files, rather than .html files
* calculate the JS editor height automatically

@ikarasz , please let me know if this looks good to you.

### DIFF
--- a/CONTRIBUTING-JavaScript.md
+++ b/CONTRIBUTING-JavaScript.md
@@ -2,53 +2,26 @@
 
 ## Writing the example
 
-With a JavaScript example you start by creating a new `.html` file in a subfolder of `live-examples/js-examples`. In this example we are going to contribute an example demonstrating the use of `Array.from` so, we'll create a new file called `array-from.html`. Since it is part of the `Array` object, we're going to put it inside the "array" subfolder.
+With a JavaScript example you start by creating a new `.js` file in a subfolder of `live-examples/js-examples`. In this example we are going to contribute an example demonstrating the use of `Array.from`, so we'll create a new file called `array-from.js`. Since it is part of the `Array` object, we're going to put it inside the "array" subfolder.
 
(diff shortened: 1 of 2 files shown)


wbamberg:
> Looks good to me, I am just wondering if we really need to set the height for `EmbedInteractiveExample` or could it be automated somehow. (It is another issue, I know)

I would love for us not to have to do this. But as far as I know it's necessary, or at least would be hard to avoid, given the architecture.

In our current architecture, as you know, we build interactive examples as their own separate web pages, served from interactive-examples.mdn.mozilla.net, and we show them in pages by embedding them in iframes. This architecture has some advantages but several quite severe problems, and as I understand it this is one of them.

A while ago we looked into trying out a different architecture, based on using embedded CodePens (https://gist.github.com/wbamberg/ca85831ec166a0bfed4bbcc031d3bf73) and I think that would have been a really interesting approach. But unfortunately we never got the resources to pursue it.

@schalkneethling probably knows more about this side of things than me though.

schalkneethling:
> Looks good to me, I am just wondering if we really need to set the height for `EmbedInteractiveExample` or could it be automated somehow. (It is another issue, I know)

It's been a while :) but I believe that has to a lot to do with the fact that we live inside an `iframe` in production. Added to that we have this concept of taller and shorter examples and, we need to do some tweaking to make this all responsive.

The way we communicate all of this on the page is then via `postMessage` as the `iframe` source and the page is on different domains. Cooky I know :)

From what I understand the idea is to break out of all/most of this as part of the major rewrite of our overall platform(including the frontend) so for now, I would suggest we stick to this way of doing things.

Having moved the JS examples from being HTML to JS is already a massive improvement to some of the internals. Thank you again @ikarasz

PR #1552, opened 2020-03-17
New example for CSS min()

jmswisher:
Based on live sample in https://developer.mozilla.org/en-US/docs/Web/css/min


wbamberg:
Thanks @jmswisher ! This works great and is correct, but I think it's a little weird to have an example with only one choice. I think with the CSS examples it's much more helpful to have several choices illustrating different ways an item can be specified and the different effects it can have. So for instance, if you had:

```css
width: min(20vw, 100px); 
```

and

```css
width: min(20vw, 1000px); 
```

...then it illustrates how it chooses `20vw`, or not, depending on the other option.

You can have I think up to 6 one-line choices, but how many to use is really a judgement call depending on the item in question. But I think it's worth having more than one. The example for `calc()`, which is a bit similar, provides 4 choices.

live-examples/css-examples/values-and-units/function-min.html
```diff
@@ -0,0 +1,30 @@
+<section id="example-choice-list" class="example-choice-list large">
+
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">width: min(50vw, 400px); </code></pre>
```
wbamberg:
```suggestion
        <pre><code class="language-css">width: min(50vw, 200px); </code></pre>
```

wbamberg: This change makes the difference between the first two options more obvious.

PR #1553, opened 2020-03-17
New example for CSS max()

jmswisher:
Based on the live sample in https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/max


wbamberg:
Is there a reason you chose `50vw` for the first two? With that I see no difference between them, I guess because at most sensible window sizes `50vw` is more than both `100px` and `400px`. If I change them to `20vw` then I get:

* first example: chooses `400px`
* second example: chooses `20vw`
* third example: chooses `100px`

...which seems more illustrative. But maybe I'm missing something here?

PR #1571, opened 2020-04-09
Updated the interactive example for Map.prototype.get().

asiby:


wbamberg:
@asiby , thanks for your PR! Please note that I won't review PRs with no description, since I don't know what the problem is that the PR is trying to solve.

asiby:
@wbamberg , that's fair. This PR is adding something to the documentation that draw people's attention to the fact that when we retrieve something from a Map (ex. `let x = myMap.get("myKey")` , the variable x will be a copy of the stored value if that value is a primitive. But x will be a direct reference to the stored value if it is a object, and it that case, if we modify that object, then the change will occur inside the Map. It will not be necessary do to `myMap.set("myKey", x)`

wbamberg:
Thanks for the clarification @asiby !

I don't think we should include this in the top-level interactive example. It's really intended to provide a basic overview of the item, and kept as simple as possible. Any more detailed documentation should be explained further down the page.

But I *do* think this is an important and useful thing to include. I'd like to suggest that you add this to the main page as a new subsection of "Examples", after the example headed "Using get()".

You could call it something like "Using get() with objects" and include something like your explanation above along with some version of the example code.

This I think would be a valuable addition to this page. What do you think?

asiby:
Thanks for you suggestions @wbamberg.

You are right. I was not thinking "simplicity" when I first added this. It makes total sense to keep it simple for the first interaction with the page and leave the rest for someone who needs to know more. I have discovered this behavior when it unexpectedly caused an error in my code because the content of the Map object got modified without calling the put() method.

I went just now and read the rest of the MDN page for Map.get() and realized that a way to go could be to modify the "Return value" section and add a sentence that reads something like "Careful considerations need to be made when the return value is an object". The term "Careful considerations" will be a link to an anchor for the "Using get() with objects" section that you have suggested.

Thoughts?

wbamberg:
I think this is good, but I think it would be more helpful to say, in "Return value", something more explicit, like:

> If the value identified by `key` is an object, a reference to the stored value is returned. See [Using get with objects]()".

Also cc @Elchi3 , who curates the JS docs and might have opinions here :).

PR #1579, opened 2020-04-29
Configure eslint and standardize the stylequide all over the js examples 

ikarasz:
Based on issue #250 I have standardized the js examples and added an eslint config.
The styleguide is not enforced yet since I cannot add a pre-commit hook.
@wbamberg what do you think? Shall I add already the script to the CI?


wbamberg:
This is fantastic work @ikarasz ! I love all the fixes this PR includes, let alone the errors it will protect us from in future. This is as far as I know the first time we are linting example code on MDN, and it's really exciting to see.

I've:

* reviewed and tested all the code changes you have made to the examples, and they all look good.
* checked that `npm run lint:js" is working.
* reviewed all the overrides in eslintrc, and they look sensible to me

I'd like @schalkneethling to have a look at this, especially wrt:

* the eslintrc changes: I think probably that the old eslintrc dated from when this repo included the editor code, before that was split out into mdn-bob. The config you have here looks sensible to me, but also I think the exact details of the config are less important than having this set up.

* adding it as a pre-commit hook. I saw in https://github.com/mdn/interactive-examples/issues/250#issuecomment-618873955 that you asked for help with that, I have no idea but Schalk might :). I don't think that has to block merging this though.

>  Shall I add already the script to the CI?

I think that would be great!

wbamberg:
Also, we should update our contribution docs. The relevant bit I think is: https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#javascript-coding-style. What we should do in my opinion is start that section by talking about ESLint. Something like:

> ## JavaScript coding style
>
> We use ESLint to help guarantee a consistent code style. We define our configuration in /.eslintrc.js. Sometimes individual examples may need to disable certain rules: to do that use the `overrides` key in the configuration file.
>
> In the rest of this section we'll describe additional conventions, not enforced using ESLint but instead using code review.


After that we should document **only conventions that we don't enforce using ESLint**. So to go through the rest of this section as it currently stands:

> ### Language choice (ES6)
>
> According to the general MDN JS guideline, aim to use ES6 to illustrate examples.
>
> ES6 examples should use:
>
> * let and const instead of var
> * Arrow functions (=>) for Anonymous Functions (see below)
> * Template literals (string text ${expression} string text)
> * Spread syntax (myFunction(...iterableObj);)

This suggests we should add 'no-var' and 'prefer-arrow-callback'. I think we already require template literals, and I don't think the bit abut spread syntax is very helpful. So I suggest we can delete this section.

> ### Semi-colons

Already enforced in ESLint, so delete this section.

> ### Line spacing

Keep this.

> ### Indentation

Already enforced in ESLint, so delete this section.

> ### Instantiating an object

Keep this.

> ### Instantiating an array

I'm not sure whether this would be covered by 'no-array-constructor'.

> ### Passing parameters

I couldn't see a rule for this, so I guess we need to keep this.

> ### Single quotes

Already enforced in ESLint, so delete this section.

> ### Property definition

This looks like 'object-property-newline'?

> ### Spaces between operators

This looks like 'space-infix-ops'?

> ### Function definition

It's not very clear to me what this is asking for or which ESLint rules might help.

> ### Anonymous function definition

Covered above by 'prefer-arrow-callback' I think. Delete this section.

> ### Class definition

The first para here is confusing, the rest I think we need to keep.

> ### Return early from if statements

Keep this.

> ### Formatting switch statements

Keep this, probably.

> ### Testing for equality

Already enforced in ESLint, so delete this section.

*****

Anyway, I'm happy to make the changes above in a follow-up PR or in this one. Or, to add the extra ESLint rules in this PR and update the documentation in a follow-up. Whichever you prefer.

ikarasz:
@wbamberg I've added all the missing rules and also updated the contribution guideline. Only the line spacing before `console` statements cannot be enforced.
2 sections left:
- Classes: I don't really understand that section. I think we can remove it since the PascalCase is already enforced with `new-cap` just like all the other es6 features.
- Property definition: We can set a rule for this, but there are 41 examples with single line definitions and we also state that we should be as short as possible. I have no preference in this question, so what do you think?

wbamberg:
Great work @ikarasz !

> * Classes: I don't really understand that section. I think we can remove it since the PascalCase is already enforced with new-cap just like all the other es6 features.
> * Property definition: We can set a rule for this, but there are 41 examples with single line definitions and we also state that we should be as short as possible. I have no preference in this question, so what do you think?

Let's just remove these sections. I agree the first is very unclear and on the second I'm happy to relax the constraint.

ikarasz:
@wbamberg, removed. I think we can merge it if you don't find any issues. Later on, a pre-commit hook can be added, but since it is documented and Travis will check it, we won't face any big problems.

wbamberg:
I've talked to @schalkneethling , and let's merge this :). Really awesome work @ikarasz .

PR #1584, opened 2020-05-12
Update globalprops-encodeuricomponent.js

LariSinisalo:
Do not escape the url parts required for parameters to parse correctly if appended to an url.

This makes the provided example be closer to what users typically would want to do.

### DIFF
--- a/live-examples/js-examples/globalprops/globalprops-encodeuricomponent.js
+++ b/live-examples/js-examples/globalprops/globalprops-encodeuricomponent.js
@@ -1,6 +1,6 @@
 // encodes characters such as ?,=,/,&,:
-console.log(encodeURIComponent('?x=шеллы'));
-// expected output: "%3Fx%3D%D1%88%D0%B5%D0%BB%D0%BB%D1%8B"
+console.log('?x='+encodeURIComponent('test?'));
+// expected output: "?x=test%3F"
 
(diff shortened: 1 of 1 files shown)


wbamberg:
@LariSinisalo , thanks for this PR. Please note that it is currently failing our linter:

```
/home/travis/build/mdn/interactive-examples/live-examples/js-examples/globalprops/globalprops-encodeuricomponent.js
  2:13  error  Unexpected string concatenation  prefer-template
  2:18  error  Operator '+' must be spaced      space-infix-ops
  5:13  error  Unexpected string concatenation  prefer-template
  5:18  error  Operator '+' must be spaced      space-infix-ops
```

If you replace your string concatenation with [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) that will fix both these problems I think.

Please see our [JS style guide](https://github.com/mdn/interactive-examples/blob/master/JS-Example-Style-Guide.md#javascript-coding-style) to see how to run the linter locally, to check for problems before pushing.

(also I just filed https://github.com/mdn/interactive-examples/pull/1585 which should make it easier to fix ESLint problems)

LariSinisalo:
Thank you for your feedback. I fixed the jslint failure by using the --fix option you suggested, which correctly replaced it with an appropriate template string.

PR #1585, opened 2020-05-12
Add lint-fix command

wbamberg:
I thought it would be helpful to expose ESLint's `--fix` option, so contributors can easily fix up code with lint errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
         "start-server": "http-server -p 9090 ./docs",
         "start-watch": "chokidar '**/{*.js,*.css,*.html,*.json}' -i 'package.json' -i 'docs/**' -i 'node_modules/**' -i 'js/editor-*.js' -c 'npm run build' --initial --silent",
         "start": "npm-run-all --parallel start-watch start-server",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,8 @@
         "start-server": "http-server -p 9090 ./docs",
         "start-watch": "chokidar '**/{*.js,*.css,*.html,*.json}' -i 'package.json' -i 'docs/**' -i 'node_modules/**' -i 'js/editor-*.js' -c 'npm run build' --initial --silent",
         "start": "npm-run-all --parallel start-watch start-server",
-        "lint:js": "eslint ./live-examples/js-examples"
+        "lint:js": "eslint ./live-examples/js-examples",
+        "lint-fix:js": "eslint --fix ./live-examples/js-examples"
```
schalkneethling, line 23: This is a complete nit so, feel free to disregard :) It seems the accepted "standard" for these things would be something like:

```suggestion
        "lint:js:fix": "eslint --fix ./live-examples/js-examples"
```

PR #1589, opened 2020-05-22
Add 'new' for constructor and remove repeated line

peterbe:
I'm still baffled that it worked with `const regex = Regex(...)` without the `new`.


wbamberg:
The extra call is intentional, I think - it is intended to show the difference the global flag makes. So:

```js
console.log(regex.test(str));
// expected output: true

console.log(regex.test(str));
// expected output: true

console.log(globalRegex.test(str));
// expected output: true

console.log(globalRegex.test(str));
// expected output: false 
```

But if this isn't clear, perhaps it's not worth keeping?

schalkneethling (approved):

PR #1596, opened 2020-06-01
Interactive examples for Intl.DisplayNames

emwsc:
Interactive examples for Intl.DisplayNames (based on issue https://github.com/tc39/ecma402-mdn/issues/1)


emwsc:
Thanks for your comments! As you asked, i removed extra examples.

PR #1615, opened 2020-07-05
Update expected output style and browser-specific output

pnevares:
Hey there! 👋 

I ran some examples recently that looked slightly off and updated them here. I also did a review of browser-specific results and updated them based on Chrome 83, Firefox 78, and Safari 13.

Would appreciate any feedback if something was missed. Thank you!

### DIFF
--- a/live-examples/js-examples/array/array-foreach.js
+++ b/live-examples/js-examples/array/array-foreach.js
@@ -2,7 +2,6 @@ const array1 = ['a', 'b', 'c'];
 
 array1.forEach(element => console.log(element));
 
-// expected output:
-// "a"
-// "b"
-// "c"
+// expected output: "a"
+// expected output: "b"
+// expected output: "c"
--- a/live-examples/js-examples/array/array-foreach.js
+++ b/live-examples/js-examples/array/array-foreach.js
@@ -2,6 +2,7 @@ const array1 = ['a', 'b', 'c'];
 
 array1.forEach(element => console.log(element));
 
-// expected output: "a"
-// expected output: "b"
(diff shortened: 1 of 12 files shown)


live-examples/js-examples/array/array-foreach.js
```diff
@@ -2,6 +2,7 @@ const array1 = ['a', 'b', 'c'];
 
 array1.forEach(element => console.log(element));
 
-// expected output: "a"
-// expected output: "b"
-// expected output: "c"
+// expected output:
+// "a"
```
ikarasz, line 6: Hi @pnevares!
Currently, I am working on unit tests. They will test the real output against the expected ones. Unfortunately, I couldn't find a better solution than adding the "expected output" prefix to each line. So I think we should leave them now as they are.

pnevares: Thanks for the info! I can restore the prefixes in this PR shortly.

wbamberg:
@pnevares , thanks for this, and sorry to be slow reviewing! All these changes look really good to me with the exception of the ones highlighted by @ikarasz . Since the exact formatting of "expected output" is important for unit tests, I'm happy to defer to @ikarasz to give his OK on this pull request.

Thanks for your contribution!

live-examples/js-examples/array/array-foreach.js
pnevares: This part of the PR should be reverted!

pnevares:
Thanks for reviewing! Apologies for adding the "expected output" change to the PR without checking first.

PR #1620, opened 2020-07-09
Update expressions-conditionaloperators.js

Krinkle:
Add an example for the common case of null and 0 being used as falsy.


wbamberg:
Thanks for your contribution @Krinkle !

I really want to keep these interactive examples to 12 lines or less, unless it's really not possible to fit a sensible example in that space.

In this case I think the existing example is OK and doesn't need the extra space, so I'm inclined to close this PR. Note that interactive examples don't need to show all the main use cases of the API, only the most common/basic uses. We have a complete "Examples" section for anything extra.

It might be worth replacing the `console.log(getFee(1));` case with `console.log(getFee(null));`, since that might be a more common usage?

PR #1623, opened 2020-07-10
Add prefer-const to ESLint config

wbamberg:
In reviewing https://github.com/mdn/interactive-examples/pull/1621 I complained about some `let` declarations that could (and therefore should) be `const`: since ESLint can enforce this for us I thought it would be worth doing.

This PR adds [`prefer-const`](https://eslint.org/docs/rules/prefer-const) to our default config, and various automatic fixes that look reasonable.

I've also relaxed `prefer-const` for three examples:
* https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/statement/statement-block.js, where using `let` seems more plausible
* https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/statement/statement-let.js, where using `let` is obviously needed
* https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/expressions/expressions-destructuringassignment.js, which is just weird: if I don't relax `prefer-const` then it complains that `rest` could be `const`, but if I make that change then I get "Error: invalid assignment to const 'rest'". I'd be happy if someone had a better solution here :(.

@ikarasz , I'd be happy to get your feedback in addition to Florian's review :).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -53,6 +53,7 @@ module.exports = {
         'object-curly-spacing': ['error', 'always'],
         'one-var-declaration-per-line': ['error', 'always'],
         'prefer-arrow-callback': 'error',
+        'prefer-const': 'error',
         'prefer-template': 'error',
         'quotes': ['error', 'single'],
         'semi': ['error', 'always'],
@@ -83,6 +84,7 @@ module.exports = {
             files: ['**/expressions/expressions-destructuringassignment.js'],
             rules: {
                 'one-var-declaration-per-line': 'off',
+                'prefer-const': 'off',
             },
         },
         {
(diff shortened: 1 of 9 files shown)


.eslintrc.js
```diff
@@ -83,6 +84,7 @@ module.exports = {
             files: ['**/expressions/expressions-destructuringassignment.js'],
             rules: {
                 'one-var-declaration-per-line': 'off',
+                'prefer-const': 'off',
```
ikarasz, line 87: I think the `destructuring: all` [option](https://eslint.org/docs/rules/prefer-const#destructuring) would solve the issue but simply turning off the rule is fine as well in my opinion.

PR #1627, opened 2020-07-22
Update regexp-prototype-test.js

owenshaupt:
Duplicate lines removed

### DIFF
--- a/live-examples/js-examples/regexp/regexp-prototype-test.js
+++ b/live-examples/js-examples/regexp/regexp-prototype-test.js
@@ -6,9 +6,6 @@ const globalRegex = RegExp('foo*', 'g');
 console.log(regex.test(str));
 // expected output: true
 
-console.log(regex.test(str));
-// expected output: true
-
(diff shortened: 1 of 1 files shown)


wbamberg:
You're the second person to file a PR about this recently :). As mentioned here: https://github.com/mdn/interactive-examples/pull/1589#issuecomment-632915091, the extra call is intentional, I think, to show the difference when the `"g"` flag is used.

But since two people have now been confused by it, maybe we should just remove it. Still, if I had to choose one of these PRs to merge I think I would prefer https://github.com/mdn/interactive-examples/pull/1589 as that includes another fix.

@schalkneethling , what do you think?

schalkneethling:
> @schalkneethling , what do you think?

Thank you for your contribution @owenshaupt 🎉  I agree with you @wbamberg, it makes sense to merge the PR with the additional fix.

schalkneethling:
As agreed: https://github.com/mdn/interactive-examples/pull/1627#issuecomment-666285020

PR #1655, opened 2020-08-09
Add an example for Array.flat()

tigleym:
PR for #1636

### DIFF
--- a/live-examples/js-examples/array/array-flat.js
+++ b/live-examples/js-examples/array/array-flat.js
@@ -1,0 +1,11 @@
+const arr1 = [0, 1, 2, [3, 4]];
+
+// Flatten to single level array.
+console.log(arr1.flat());
+// expected output: [0, 1, 2, 3, 4]
+
+const arr2 = [0, 1, 2, [[[3, 4]]]];
+
+// Flatten to two-level array.
+console.log(arr2.flat(2));
+// expected output: [0, 1, 2, [3, 4]]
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/array/array-flat.js
```diff
@@ -0,0 +1,11 @@
+const arr1 = [0, 1, 2, [3, 4]];
+
+// Flatten to single level array.
```
wbamberg, line 3: Generally we don't use comments in interactive examples except for "expected output" or exceptional non-obvious things. I suppose it's a kind of "show don't tell" kind of thing - we have a whole page of docs to say what the feature does, the example is just supposed to show it. So could you remove this line?

live-examples/js-examples/array/array-flat.js
```diff
@@ -0,0 +1,11 @@
...
+console.log(arr1.flat());
+// expected output: [0, 1, 2, 3, 4]
+
+const arr2 = [0, 1, 2, [[[3, 4]]]];
+
+// Flatten to two-level array.
```
wbamberg, line 9: Same comment for this line.

tigleym:
@wbamberg Thanks for the quick feedback! Please disregard my latest commit since I just submitted it before seeing your feedback.

live-examples/js-examples/array/array-flat.js
tigleym: Done!

PR #1664, opened 2020-08-18
Update example for track element

Pablo-No:
I've updated the example of the track element because the formatting didn't look well.

### DIFF
--- a/live-examples/html-examples/image-and-multimedia/css/track.css
+++ b/live-examples/html-examples/image-and-multimedia/css/track.css
@@ -1,5 +1,5 @@
 video {
-    width="250"
+    width: 250px;
 }
 video::cue { 
     font-size: 1rem;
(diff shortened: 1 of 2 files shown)


live-examples/html-examples/image-and-multimedia/css/track.css
```diff
@@ -1,3 +1,6 @@
-video::cue {
+video {
+    width="250"
```
schalkneethling, line 2: This is not valid CSS:

```suggestion
    width: 250px;
```

live-examples/html-examples/image-and-multimedia/css/track.css
```diff
@@ -1,3 +1,6 @@
...
+}
```
schalkneethling, line 3: Please add an empty line between style rule blocks
```suggestion
}

```

schalkneethling (changes requested):
Thank you for your contribution @Pablo-No 🎉  - A couple of items to address and then we can merge.

schalkneethling (approved):
r+ Thank you @Pablo-No 🥳

PR #1677, opened 2020-08-22
Update number-min-value.js

lesterpotter:
The former example that was fixed demonstrated Number.NEGATIVE_INFINITY and not Number.MIN_VALUE.  Number.MIN_VALUE is the smallest POSITIVE number representable.

### DIFF
--- a/live-examples/js-examples/number/number-min-value.js
+++ b/live-examples/js-examples/number/number-min-value.js
@@ -1,12 +1,8 @@
-function multiply(x, y) {
-  if (x * y < Number.MIN_VALUE) {
-    return 'Process as -Infinity';
-  }
-  return (x * y);
-}
-
-console.log(multiply(5e-324, 1));
+console.log(Number.MIN_VALUE * .5);
 // expected output: 5e-324
 
-console.log(multiply(-1.7976931348623157e+308, 2));
-// expected output: Process as -Infinity
+console.log(Number.MIN_VALUE * (.5 * Number.EPSILON));
+// expected output: 0
+
+console.log(Number.MIN_VALUE * (.5 + Number.EPSILON));
+// expected output: 5e-324


live-examples/js-examples/number/number-min-value.js
```diff
@@ -1,12 +1,8 @@
...
-  return (x * y);
-}
-
-console.log(multiply(5e-324, 1));
+console.log(Number.MIN_VALUE * .5);
 // expected output: 5e-324
 
-console.log(multiply(-1.7976931348623157e+308, 2));
-// expected output: Process as -Infinity
+console.log(Number.MIN_VALUE * (.5 * Number.EPSILON));
```
velusgautam, line 4: This example you suggested is more  complex to understand than the current example. 
Also it uses another value (Number.EPSILON) to explain Number.MIN_VALUE, that I feel is not the best way to explain.

velusgautam (changes requested):
Need more simple ways to explain so everyone can understand it easy.

live-examples/js-examples/number/number-min-value.js
lesterpotter: The current example may be easy to understand, but it is wrong because it does not explain MIN_VALUE, it explains NEGATIVE_INFINITY which is not the subject of the example page.

But you are right. because (.5 * Number.EPSILON) is strange given the definition of Number.EPSILON.

How about...
console.log(Number.MIN_VALUE);
// expected output: 5e-324
console.log(Number.MIN_VALUE * 0.6);
// expected output: 5e-324
console.log(Number.MIN_VALUE * 0.5);
// expected output 0

lesterpotter: Sorry, my comment has a mistake.  I said (.5 * Number.EPSILON).  I missed the "+" in the original and saw a "*" instead.  So, though it correctly shows the smallest achievable granularity, maybe just using "0.6" would be sufficient without confusing things with EPSILON.

velusgautam: Yes this look more clear. And I also agree its not always -Infinity

schalkneethling:
@wbamberg would appreciate your thoughts on this one. Thanks.

wbamberg:
Thanks for your PR! Please note that this example has been rewritten since the PR was filed: https://github.com/mdn/content/issues/7975. So I'm closing it, but let us know if you still feel it is incorrect.

PR #1694, opened 2020-09-28
Live example const update

velusgautam:
Update var to const in css-examples js files
Removed unused variables


velusgautam:
@wbamberg

PR #1698, opened 2020-10-05
Fix Grammar Error in a comment

krankj:
The correction is regarding the usage of word here.

Since Math.random() always returns a value 0 <= x < 1, using the word 'between' in  "expected output: a number between 0 and 1" falsely indicates that the function Math.random() does not ever return 0.


chrisdavidmills (approved):
LGTM; thanks @krankj !

velusgautam (approved):

PR #1703, opened 2020-10-14
Update flex-basis.html

makaria:
[flex-basis](https://drafts.csswg.org/css-flexbox/#valdef-flex-flex-basis) accepts the same values as the width and height properties, which does not include number zero.

### DIFF
--- a/live-examples/css-examples/flexbox/flex-basis.html
+++ b/live-examples/css-examples/flexbox/flex-basis.html
@@ -7,7 +7,7 @@
     </div>
 
     <div class="example-choice">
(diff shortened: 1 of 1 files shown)


velusgautam:
I can see in [css-flexbox](https://drafts.csswg.org/css-flexbox/#valdef-flex-basis-content) we can use 0. Can you please check again and see your change request is correct? 

Screenshot 2020-10-14 at 5 42 50 PM

velusgautam (changes requested):
Please review

makaria:
Screen Shot 2020-10-15 at 13 19 16

PR #1704, opened 2020-10-16
Use a fieldset within the input/radio example

mfairchild365:
The fieldset and legend are required for accessibility. My concern is that some developers will copy the example without the needed fieldset/legend.


scottaohara:
would request this be merged

schalkneethling (approved):
Thank so much for your contribution @mfairchild365 🎉

PR #1705, opened 2020-10-16
fix: Clarify string-endswith examples

JasonMFry:
These examples were quite confusing. As a user I expect the easy,
obvious example first, not an example with a "gotcha". And unfortunately
both of these examples were "gotchas", or best case scenario they
illustrated the more complicated things one can do with
string.endsWith(), but such examples should come _after_ the simple
ones.


schalkneethling (approved):
Thank you for your contribution, @JasonMFry 🎉

PR #1706, opened 2020-10-20
Add example for Intl.toString

hamishwillee:
This adds trivial example for `Intl[Symbol.toStringTag]`  following its addtition in https://bugzilla.mozilla.org/show_bug.cgi?id=1670053
This is to populate https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/@@toStringTag

Tested on a codepen. Seems to work OK. But I am not great a JavaScript, so appreciate your review!

### DIFF
--- a/live-examples/js-examples/intl/intl-prototype-@@tostringtag.js
+++ b/live-examples/js-examples/intl/intl-prototype-@@tostringtag.js
@@ -1,0 +1,4 @@
+const intl1 = Object.prototype.toString.call(Intl);
+
+console.log(intl1);
+// expected output: "[object Intl]"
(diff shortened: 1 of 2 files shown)


chrisdavidmills (changes requested):
The actual example itself looks good @hamishwillee, but to get it to appear in the interactive example system, you also need to add an entry to https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/intl/meta.json.

hamishwillee:
Thanks @chrisdavidmills. I've added the item to the meta. I think correct - at least the example works. 

FYI, I also created https://github.com/mdn/interactive-examples/pull/1709 to confirm I understand the process.

chrisdavidmills (approved):
@hamishwillee I've tested this locally, works like a charm! Let's get this merged.

PR #1707, opened 2020-10-21
Show anonymous function usage in array-findindex.js

benjamingr:
After feedback I got regarding the current example - show an anonymous function first.

### DIFF
--- a/live-examples/js-examples/array/array-findindex.js
+++ b/live-examples/js-examples/array/array-findindex.js
@@ -1,6 +1,11 @@
 const array1 = [5, 12, 8, 130, 44];
 
-const isLargeNumber = (element) => element > 13;
+console.log(array1.findIndex((element) => element > 13));
+// expected output: 3
+
+// or with a named function
+
+const isLargeNumber = (element) => element > 13);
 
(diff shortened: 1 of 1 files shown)


benjamingr:
🦗

caugner:
Hi @benjamingr, thank you for taking the time to submit this PR, and sorry it took so long get back to you! 👋 

I have decided to close this PR without merging for the following reasons:

- Anonymous/arrow functions a [concept on their own](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), and their different use cases don't need to be covered by each function example accepting a callback function. (Note also that, strictly speaking, the existing example already uses an anonymous/arrow function, even though it's assigned to a variable).
- Using an anonymous/arrow function is covered by the [Syntax section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#syntax) of the article.
- Most examples of other Array.prototype functions are using the same pattern (assigning an arrow function to a descriptive variable, and using that as a parameter).

Thank you for your understanding, and keep up the good work! 🙏

PR #1708, opened 2020-10-21
autocapitalize example

rachelandrew:
Adding an example for the global attribute autocapitalize, working on https://github.com/mdn/sprints/issues/3802

This is shipping in Firefox 83. MDN page: https://wiki.developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-autocapitalize.html
+++ b/live-examples/html-examples/global-attributes/attribute-autocapitalize.html
@@ -1,0 +1,11 @@
+<h3>Sentence Case</h3>
+<div class="fields">
+  <label for="sentence">Write sentences in lowercase.</label>
+  <input type="text" autocapitalize="sentences" id="sentence">
+</div>
+
+<h3>Word Case</h3>
+<div class="fields">
+  <label for="word">Write sentences in lowercase.</label>
+  <input type="text" autocapitalize="words" id="word">
+</div>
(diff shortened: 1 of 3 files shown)


schalkneethling:
According to the BCD table this should be supported in Chrome, Edge and behind a flag in Firefox. I tried it out using the on-screen keyboard on MacOS, but it does not seem to work in Chrome nor Firefox. @rachelandrew Thoughts?

rachelandrew:
it's supported on iOS, there is discussion in the PR for the BCD: https://github.com/mdn/browser-compat-data/pull/7003

schalkneethling:
> it's supported on iOS, there is discussion in the PR for the BCD: [mdn/browser-compat-data#7003](https://github.com/mdn/browser-compat-data/pull/7003)

Thank you, Rachel. I reckon we should then still wait until there is wider support before merging this one.

schalkneethling:
> it's supported on iOS, there is discussion in the PR for the BCD: [mdn/browser-compat-data#7003](https://github.com/mdn/browser-compat-data/pull/7003)

Also, just so I know/understand. This will be supported on desktop and mobile correct? It is not a mobile-only feature?

mattiapontonio:
If you need it and you have a URL I can test it on my smartphone.

wbamberg:
@Rumyra , as the page says:

> The autocapitalize attribute doesn't affect behavior when typing on a physical keyboard. Instead, it affects the behavior of other input mechanisms, such as virtual keyboards on mobile devices and voice input. 

For this reason I don't think this is really suitable for an interactive example - it's more likely to confuse people.

Also I could not get this to work on Chrome desktop using the voice assistant or virtual keyboard on macos.

wdyt? Shall we close it?

PR #1711, opened 2020-10-23
scrollable-table live example

brookjordan:
Adding an example for the scrollable table, working on mdn/sprints#3792

Planning to add to the MDN page here: https://wiki.developer.mozilla.org/en-US/docs/Web/HTML/Element/table

Prototype for testing available here: https://codepen.io/brookjordan/pen/YzWNpmx


brookjordan:
This will be better placed in an inline example

PR #1712, opened 2020-10-23
Update expressions-logical-nullish-assignment.js

davidgilbertson:
This more clearly demonstrates what `??=` does, to differentiate it from `||=`.


Rumyra:
Hi @davidgilbertson thanks so much for your pr, however we believe the current example is a good illustration of `??=` bearing in mind that `a.speed` is undefined, so we'll be closing this - but thank you!

Issue #1649, opened 2020-08-09
JS examples: need example for Intl.NumberFormat.formatToParts()


wbamberg:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts

The interactive example could follow a similar pattern to the example for [Intl.RelativeTimeFormat.formatToParts()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/formatToParts).

NiedziolkaMichal:
Closed via #1717

PR #1717, opened 2020-11-08
Added Intl.NumberFormat.formatToParts() example

MendyBerger:
#1649 

I took the `Intl.NumberFormat` parts from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat), and the `formatToParts` parts from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/formatToParts)


mattiapontonio:
In general should be challanging to not use `const` at all here. 🤔 This is why reading You only have to think at the "standard keyword" in the options and so on.

Rumyra (approved):

Rumyra:
Thanks so much @MendyBerger 👍

PR #1718, opened 2020-11-08
Removes redundant aria role attribute

btrem:
Per "Document conformance requirements for use of
ARIA attributes in HTML"
https://www.w3.org/TR/html-aria/#docconformance 
"explicitly assigning [the element’s implicit
role] in markup is NOT RECOMMENDED”.

### DIFF
--- a/live-examples/html-examples/text-content/main.html
+++ b/live-examples/html-examples/text-content/main.html
@@ -1,6 +1,6 @@
 <header>Gecko facts</header>
 
-<main role="main">
+<main>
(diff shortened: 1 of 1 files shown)


schalkneethling:
@btrem I believe we mainly kept this for older browsers like IE11 but, I reckon we have come far enough that it is OK to remove. Thoughts @Rumyra?

Rumyra:
Oh yeh absolutely 👍

schalkneethling (approved):
Thank you, @btrem

PR #1721, opened 2020-11-12
moving the gap examples

rachelandrew:
The `grid` folder still had the `row-gap` and `column-gap` examples grid prefixed. Moved these to Box Alignment and renamed. The gap property was already there.


chrisdavidmills (approved):
LGTM, thanks @rachelandrew !

I'm assuming the next stage is to rewrite all the interactive example references?

PR #1726, opened 2020-11-20
Remove reduce function from restparameters example

jespertheend:
reduce() can be confusing for new users, it's better to use a simple for loop here.

### DIFF
--- a/live-examples/js-examples/functions/functions-restparameters.js
+++ b/live-examples/js-examples/functions/functions-restparameters.js
@@ -1,6 +1,6 @@
 function sum(...theArgs) {
   let total = 0;
-  for(const arg of theArgs){
+  for (const arg of theArgs) {
     total += arg;
   }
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/functions/functions-restparameters.js
```diff
@@ -1,7 +1,9 @@
 function sum(...theArgs) {
-  return theArgs.reduce((previous, current) => {
-    return previous + current;
-  });
+  let total = 0;
+  for(const arg of theArgs){
```
Rumyra, line 3:
```suggestion
  for (const arg of theArgs) {
```

Rumyra (changes requested):

Rumyra (approved):

Rumyra:
Thanks @jespertheend - we agree this does make it easier to understand!

PR #1727, opened 2020-11-22
Update Intl.DateTimeFormat to show options in constructor

hamishwillee:
Previous example just shows default formatting based on locale, and locale with language fallback. This extends the example just a little to show how the date can be formatted using style and custom options.

### DIFF
--- a/live-examples/js-examples/intl/intl-datetimeformat.js
+++ b/live-examples/js-examples/intl/intl-datetimeformat.js
@@ -1,12 +1,22 @@
-const date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));
+const date = new Date(Date.UTC(2020, 11, 20, 3, 23, 16, 738));
 // Results below assume UTC timezone - your results may vary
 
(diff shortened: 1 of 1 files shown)


chrisdavidmills (changes requested):
Hi @hamishwillee !

This is looking pretty good. I just had a coupleof issues here:

1. Generally we try to keep interactive examples to about 13 lines long at a maximum, otherwise they don't fit very well in the context of the rest of the page. Can you try to cut this down a bit?

2. You've got some CI failures to pick through and fix, mainly just stylistic stuff.

hamishwillee:
@chrisdavidmills This was already thirteen lines before I started, and showed only a fraction of what this can do :-(. I deleted a few interesting options and got it down to 14. If that is too many we might as well just just close this an accept we can't do a good job.

chrisdavidmills:
This looks much better! You still have a CI error, but nearly there.

hamishwillee:
Fixed, I hope! Travis running a bit slow.

chrisdavidmills (approved):
Much better, thanks!

PR #1730, opened 2020-11-26
fixing perspective-origin

rachelandrew:
Reported in https://github.com/mdn/sprints/issues/3922 the `perspective-origin` demo wasn't working.

It was referencing the CSS file for `perspective` which meant it didn't have the `perspective` property set and therefore `perspective-origin` quite reasonably wasn't doing anything. I have set it to ref a separate CSS file which adds back the property.


chrisdavidmills (approved):
@rachelandrew tested it locally, and it works fine; thanks for the fix!

PR #1748, opened 2020-12-31
Add options example to intl-collator

hamishwillee:
This modifies the Intl.Collator() example to show how you can use options, since these can radically affect behaviour.

Note that I could have extended the function to pass through the options as well, but decided not to - IMO better to show the actual call as a user will try it.

Appreciate this is arguable.

### DIFF
--- a/live-examples/js-examples/intl/intl-collator.js
+++ b/live-examples/js-examples/intl/intl-collator.js
@@ -4,5 +4,5 @@ console.log(['Z', 'a', 'z', 'ä'].sort(new Intl.Collator('de').compare));
 console.log(['Z', 'a', 'z', 'ä'].sort(new Intl.Collator('sv').compare));
 // expected output: ["a", "z", "Z", "ä"]
 
-console.log(['Z', 'a', 'z', 'ä'].sort(new Intl.Collator('de', {caseFirst: 'upper'} ).compare));
+console.log(['Z', 'a', 'z', 'ä'].sort(new Intl.Collator('de', { caseFirst: 'upper' } ).compare));
 // expected output: ["a", "ä", "Z", "z"]
--- a/live-examples/js-examples/intl/intl-collator.js
+++ b/live-examples/js-examples/intl/intl-collator.js
@@ -1,10 +1,8 @@
-function letterSort(lang, letters) {
-  letters.sort(new Intl.Collator(lang).compare);
-  return letters;
-}
+console.log(['Z', 'a', 'z', 'ä'].sort(new Intl.Collator('de').compare));
+// expected output: ["a", "ä", "z", "Z"]
 
-console.log(letterSort('de', ['a', 'z', 'ä']));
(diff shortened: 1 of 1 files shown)


hamishwillee:
Any advice on the error here. It might be a line ending thingy, but if so, then line endings should perhaps be forced by the repo itself.

hamishwillee:
@chrisdavidmills as per my comment above, can you please advise on the error?

chrisdavidmills:
@schalkneethling Hi Schalk — another example of a failing build on interactive examples, and I have literally no idea how to find what the cause is. Can you advise me on how you find the error messages, so I don't have to keep bugging you each time? ;-)

schalkneethling:
@chrisdavidmills Next to the Travis CI logo there is a Details link. If you click it, it takes you to a page where there is another link to the failed build. Once you click on it, it will take you to the build log. Once you scroll down you can see why it failed.

The alternative is to pull down the branch and run `yarn lint:js` and see what the output there is. In this case it is:

```
/home/travis/build/mdn/interactive-examples/live-examples/js-examples/intl/intl-collator.js
  7:63  error  A space is required after '{'   object-curly-spacing
  7:82  error  A space is required before '}'  object-curly-spacing
✖ 2 problems (2 errors, 0 warnings)
```

The above is kinda odd as I would have thought Prettier would have taken care of that. I need to have a look at the tooling and CI for this project. For one thing, we should move to Github Actions.

hamishwillee:
@schalkneethling Thanks very much for that.

FYI, I ran the linter locally I couldn't see it among all the LF errors that appear on windows. Perhaps we should have a **.gitattributes** of `text eol=lf` if the format matters to the tooling?
```
   2:1   error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style
```
I ended up fixing this by running `yarn test --fix` and just pushing the changed file.

schalkneethling:
> Perhaps we should have a **.gitattributes** of `text eol=lf` if the format matters to the tooling?

Ah yes, that is one option. The other might be to add a pre-commit hook to run `--fix`. The thing that is a little annoying about that is if it does fix something, you then have to `git add .` and `git commit -m 'automated eslint and/or prettier fixes'`

Perhaps one could automate that away as well. Perhaps even do that via a Github action on pull requests or something like that. Then again, it may just be simpler to add a `.gitattributes` file as you suggested 😃

chrisdavidmills (approved):
OK, tested, and this works fine; nice work @hamishwillee !

hamishwillee:
@schalkneethling Re https://github.com/mdn/interactive-examples/pull/1748#issuecomment-761989839 - can I leave this problem in your capable hands, or do you need me to do something to follow this up?

schalkneethling:
> @schalkneethling Re [#1748 (comment)](https://github.com/mdn/interactive-examples/pull/1748#issuecomment-761989839) - can I leave this problem in your capable hands, or do you need me to do something to follow this up?

💯

PR #1762, opened 2021-02-10
Added all Tab-Sizes.

sahilempire:


sahilempire:
CSS styleguide.

sahilempire:
Hi @charliecalvert,

**Please help me** to go ahead with my pull request.
This shows i have **no write access** to this repository.

Please guide me.

Thank You.

chrisdavidmills:
Hi @sahilempire ! 

I'm not concerned about the "no write access" — I think this just means that you can't directy write to the repository. You have to get your changes in my submitting pull requests, which we will then review. This is normal and fine, and it looks like you'v submitted your PR OK.

What is a little more concerning is that the Travis CI process has returned a build error — see the "All checks have failed" message below. Looking at the [build report](https://travis-ci.com/github/mdn/interactive-examples/builds/216696044), it says:

```
ENOENT: no such file or directory, open './live-examples/css-examples/cssom-view/style.css'

MDN-BOB: (mdn-bob.js/@init) Error during build: Error: MDN-BOB: (pageBuilder.js/@buildPages) Error while building pages: ./live-examples/css-examples/tab-size/meta.json: Error: ENOENT: no such file or directory, open './live-examples/css-examples/cssom-view/style.css'
```  

Let's look at your files and see what the problem could be...

OK, the build is failing because your meta.json file is pointing to the wrong locations for the CSS and HTML files (`cssExampleSrc` and `exampleCode`). You need to update those URLs to the correct location. Your meta `fileName` and `title` also spell `tab` incorrectly as `tabe` — this needs to be updated too.

On to the example itself:

* First of all, you should probably put the tab-size example inside the text folder, as it is part of the text spec.
* In fact, there is already a tab-size example in the text folder! You should remove the tab-size folder.
* I've tried testing the tab-size example from the text folder locally, and it works in Chrome, but not Firefox, as it's not supported there yet without a prefix. Your example in the tab-size folder doesn't work — make sure you test examples locally before submitting pull requests.

So it looks your new example is not needed, because there was already an example present. However, one thing does need updating on the existing example. As Will said on his original description — "Note that it's prefixed in Firefox, so the interactive example choices need to set both `tab-size` and `-moz-tab-size`." 

So this means you need to update the choices to include both, e.g.

```
<pre><code class="language-css">-moz-tab-size: 4;
tab-size: 8;</code></pre>
```

PR #1771, opened 2021-02-25
Remove extra "new" instantiation from TypedArray.from() live demo

echuber2:
The extra `new` step isn't necessary before using `Int16Array.from`, is it?

### DIFF
--- a/live-examples/js-examples/typedarray/typedarray-from.js
+++ b/live-examples/js-examples/typedarray/typedarray-from.js
@@ -1,4 +1,4 @@
-let uint16 = Int16Array.from('12345');
+const uint16 = Int16Array.from('12345');
 
 console.log(uint16);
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/typedarray/typedarray-from.js
```diff
@@ -1,5 +1,4 @@
-let uint16 = new Int16Array;
-uint16 = Int16Array.from('12345');
+let uint16 = Int16Array.from('12345');
```
velusgautam, line 1:
```suggestion
const uint16 = Int16Array.from('12345');
```

velusgautam (changes requested):

wbamberg (approved):
Looks good to me. Thanks @echuber2 !

PR #1773, opened 2021-02-26
Update -  array-slice.js

NegiAkash890:
This adds two examples to the current snippet . The two changes are as follows :
1. Slice method when only one index(negative) is provided .
2. Slice method when a range of index is provided ( one index being positive and other negative ) .

 Fixes mdn/content#1441


wbamberg (approved):
Looks good to me, thanks @NegiAkash890 !

PR #1776, opened 2021-03-04
Fixes markup for <cite> example

danburzo:
I don't think `<footer>` was meant to be used with non-sectioning content like this. The [page for `blockquote`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) contains what I believe to be the correct markup,  [corroborated by the HTML living spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element).

(Update: the `<footer>` usage was previously recommended in the W3C HTML spec)


wbamberg (approved):
Looks good to me, thanks @danburzo !

PR #1780, opened 2021-03-15
Remove redundant Math.floor()

Napuu:
Outer Math.floor() is already enough.
As edge scenario, let's assume Math.random returns 0.9999
and we call this function with 5.9999
=> Math.floor(0.9999 * 5.9999) = 5


velusgautam (approved):
LGTM 🙂

velusgautam:
@wbamberg

wbamberg (approved):
Sure, that sounds convincing :). Thanks for the update @Napuu !

PR #1781, opened 2021-03-18
Clarify example. Closes #1622

### LINKED ISSUE #1622: Object.prototype.isPrototypeOf
Replace existing example of the [isPrototypeOf] ([MDN]) by the following:

```js
function Foo() {}
function Bar() {}

Bar.prototype = Object.create(Foo.prototype);

const bar = new Bar();

console.log(Foo.prototype.isPrototypeOf(bar));
// expected output: true

console.log(Bar.prototype.isPrototypeOf(bar));
// expected output: true
```
Current example is vague.

[isPrototypeOf]: https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/object/object-prototype-isprototypeof.js
[MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf

hinell:
More details at #1622

### DIFF
--- a/live-examples/js-examples/object/object-prototype-isprototypeof.js
+++ b/live-examples/js-examples/object/object-prototype-isprototypeof.js
@@ -1,12 +1,10 @@
-function object1() {}
-function object2() {}
+function Foo() {}
+function Bar() {}
 
-object1.prototype = Object.create(object2.prototype);
+Bar.prototype = Object.create(Foo.prototype);
 
-const object3 = new object1();
+const bar = new Bar();
 
-console.log(object1.prototype.isPrototypeOf(object3));
-// expected output: true
+console.log(Foo.prototype.isPrototypeOf(bar));
 
-console.log(object2.prototype.isPrototypeOf(object3));
-// expected output: true
+console.log(Bar.prototype.isPrototypeOf(bar));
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/object/object-prototype-isprototypeof.js
```diff
@@ -1,12 +1,10 @@
...
+function Bar() {}
 
-object1.prototype = Object.create(object2.prototype);
+Bar.prototype = Object.create(Foo.prototype);
 
-const object3 = new object1();
+const bar = new Bar();
 
-console.log(object1.prototype.isPrototypeOf(object3));
-// expected output: true
```
velusgautam, line 9: Why you deleted expected outputs?

live-examples/js-examples/object/object-prototype-isprototypeof.js
```diff
@@ -1,12 +1,11 @@
...
+console.log(Foo.prototype.isPrototypeOf(bar));
```
wbamberg, line 8: This should have:

```
 // expected output: true
```

after it

wbamberg (changes requested):
Thanks @hinell . I like this change but it should have an `// expected output` line for both `console.log()` statements.

wbamberg (approved):
Thanks @hinell !

PR #1782, opened 2021-03-18
Added example for RegExp.prototype.hasIndices

SebastianZ:
This adds an interactive example for the [newly introduced `RegExp.prototype.hasIndices`](https://bugzilla.mozilla.org/show_bug.cgi?id=1519483) property.

This is in preparation of adding the related content.

Sebastian

### DIFF
--- a/live-examples/js-examples/regexp/meta.json
+++ b/live-examples/js-examples/regexp/meta.json
@@ -66,7 +66,7 @@
             "title": "JavaScript Demo: RegExp.prototype[@@split]",
             "type": "js"
         },
-        "regexPrototypeDotAll": {
+        "regexpPrototypeDotAll": {
             "exampleCode": "./live-examples/js-examples/regexp/regexp-prototype-dotall.js",
             "fileName": "regexp-prototype-dotall.html",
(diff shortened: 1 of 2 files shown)


chrisdavidmills (approved):
Thanks @SebastianZ ! This looks fine. The `DotAll` example works fine, and the `HasIndices` examples works fine in newest Firefox 88 build and newest Chrome Canary.

The only problem we've got here is that the [ESLint CI build is failing](https://travis-ci.com/github/mdn/interactive-examples/builds/220549120), I'm guessing because the version of ESLint we are using doesn't know about `hasIndices` yet. @schalkneethling can this be updated easily?

escattone:
I updated `eslint` to the latest version `7.23.0`, but it still doesn't recognize the `d` flag as it should. I'm going to merge this for now, and we can keep looking into resolving the `eslint` issue.

SebastianZ:
Muchas gracias, Ryan, for merging my PR! 😃

I assume, the `d` flag is just too new and therefore not implemented in Node.js yet.

Sebastian

PR #1784, opened 2021-03-23
Async generator examples

Rumyra:
Interactive examples for async generator docs; see https://github.com/mdn/content/pull/2861

Both statement and operator added.

NB there is an existing `statement-async-function-asterisk.js` which is a great example, but a great _in-page_ example. I have added `statement-asyncfunctionasterisk.js` as the interactive example. I have not removed the former in case it's being used somewhere I am unaware of 👍


wbamberg (changes requested):
eek, sorry to be slow reviewing this @Rumyra . 

These examples look great, but you need to update the metadata to tell the site builder about these examples, or they won't get built: https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING-JavaScript.md#updating-the-metadata.

> NB there is an existing statement-async-function-asterisk.js which is a great example, but a great in-page example. I have added statement-asyncfunctionasterisk.js as the interactive example. I have not removed the former in case it's being used somewhere I am unaware of 👍

I think it would be better to replace that example with the new one. I can't imagine an example in this repo could be being used as anything other than an interactive example.

Rumyra:
Okey dokey - I have tidied this up a bit - made sure the metadata is up to date and copied over the existing example `statement-async-function-asterisk.js` and renamed the other because 👍

wbamberg (approved):
Thanks @Rumyra , these are super clean examples!

PR #1798, opened 2021-04-14
change no-invalid-regexp to warn for hasindices

escattone:
Changing the `no-invalid-regexp` rule to a warning (rather than a error), but only for `**/regexp/regexp-prototype-hasindices.js`. We can remove this change once `eslint` recognizes the `"d"` flag to the `RegExp` constructor.

This will get the builds passing again after they were broken by #1782.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -226,6 +226,13 @@ module.exports = {
             },
         },
         {
+            // Only until the "d" flag is supported by eslint.
+            files: ['**/regexp/regexp-prototype-hasindices.js'],
+            rules: {
+                'no-invalid-regexp': 'warn',
+            },
+        },
+        {
             files: ['**/set/set-prototype-foreach.js'],
             rules: {
                 'no-unused-vars': 'off',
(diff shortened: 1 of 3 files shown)


peterbe:
What about upgrading to [`eslint` 7.24.0](https://www.npmjs.com/package/eslint/v/7.24.0) first?

escattone:
@peterbe 

> What about upgrading to [`eslint` 7.24.0](https://www.npmjs.com/package/eslint/v/7.24.0) first?

I tried that first, but it didn't help.

escattone:
@peterbe In fact, I'll add that upgrade to this PR, just to show that approach doesn't resolve the issue (the warning is still issued).

peterbe:
So even the latest version of eslint doesn't like this `d` flag. 
https://travis-ci.com/github/mdn/interactive-examples/builds/222796771 from https://github.com/mdn/interactive-examples/pull/1796
I'm guessing this new `d` flag is that something that's bleeding edge super-new that not even latest eslint can recognize it?

schalkneethling (approved):

PR #1801, opened 2021-04-15
Fix missing content in grid-auto-rows CSS demo

### LINKED ISSUE #1800: Final item missing content in grid-auto-rows demo
## What page(s) did you find the problem on?
[grid-auto-rows](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-rows)

## Specific page section or heading?
At the beginning of the page, right underneath the `grid-auto-rows` heading text. It's the interactive CSS demo.

## What is the problem?
The final block is missing its content. It should have the text "Five" displayed but its content is empty. 

Screen Shot 2021-04-14 at 11 50 40 PM

## What did you expect to see?
The text "Five" displayed in the fifth box of demo.

Screen Shot 2021-04-15 at 12 10 13 AM

## Did you test this? If so, how?
Synced with `mdn/interactive-examples` and ran locally with `yarn start`, the error still persists as the last child `<div>` of `<div class="example-element">` is empty.

```html
  <div>One</div>
  <div>Two</div>
  <div>Three</div>
  <div>Four</div>
  <div></div>
</div>
```

tannerdolby:
Fixes #1800


escattone:
@tannerdolby Just FYI, if you merge the `master` branch into your PR, the build should work again.

tannerdolby:
@escattone Merged my feature branch with `master` and the build is now working.

tannerdolby:
@wbamberg ping for review

wbamberg (approved):
Looks good to me @tannerdolby . Thanks for the fix!

tannerdolby:
Your welcome @wbamberg !

wbamberg:
Actually now I look at it, I wonder if it was intentional to omit "Five", because it shows how the last row gets a different height when `min-content` and `auto` are specified. But perhaps rather than just omitting the content entirely, we could make it smaller?

Screen Shot 2021-04-30 at 9 16 08 PM

tannerdolby:
@wbamberg Now that I look back at it, there could be a chance that was the case. I guess without the text for the fifth box, it's a bit easier to understand and see the size of the grid row track when using `min-content` and `auto` values. But in my opinion, having the text "Five" in the last box is equally as clear that `min-content` will fill the largest minimal content distribution and `auto` the maximal content distribution. I think it's clear with the "Five" text having the same font-size as it's siblings. Reducing the font size a bit does also highlight that `min-content` and `auto` will always fill a minimal or maximal content distribution but I'm not sure it's necessary.

Without text content for the fifth box.

https://user-images.githubusercontent.com/48612525/116770890-cad87180-a9fb-11eb-8455-2fda4c7c7eaf.mov

Video with text in last box and normal size font

https://user-images.githubusercontent.com/48612525/116771004-d4aea480-a9fc-11eb-89ac-48207c5e9f44.mov

rachelandrew:
Yes the "missing" text was intentional as otherwise you can't see the difference (it's a bit tricky in such a small space). I think the solution of adding a smaller font size would work.

tannerdolby:
> I think the solution of adding a smaller font size would work.

@rachelandrew Ok thank you for the clarification. I will submit a PR to make the font size a bit smaller and we can go from there.

PR #1802, opened 2021-04-19
Adds border-block examples

### LINKED ISSUE #1637: CSS examples: need examples for border-block property and its longhands
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block

and its longhands:
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block-width
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block-style
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block-color

This property is part of the "logical properties". This group of properties set things like `padding`, `margin`, and `border`, but instead of using physical attributes to describe the part of the box (like top, left, bottom, right) they use logical attributes, that are relative to the direction in which the text runs, so as to make sense for vertical and RTL writing systems. In this case we set borders in the "block" sides as opposed to "inline" (see https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties/Basic_concepts#Block_and_inline_dimensions).

This particular property is like setting https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block-end and https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/border-block-start both together, and it would be fine to repurpose those interactive examples here.

rachelandrew:
Fixes #1637 

Adds `border-block` and longhands.

### DIFF
--- a/live-examples/css-examples/logical-properties/border-block-color.html
+++ b/live-examples/css-examples/logical-properties/border-block-color.html
@@ -1,0 +1,32 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="border-block-color">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">border-block-color: red;
+writing-mode: horizontal-tb;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+         </button>
+    </div>
+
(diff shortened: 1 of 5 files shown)


chrisdavidmills (approved):
I've tested these locally, and they look fine. Thanks @rachelandrew !

PR #1803, opened 2021-04-19
text-emphasis examples

### LINKED ISSUE #1644: CSS examples: need examples for text-emphasis longhand properties and text-emphasis-position
We have an interactive example for the text-emphasis shorthand property (https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis) but none for its longhands:

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-color

We should also have an interactive example for https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position.

For the sake of consistency it would be best to adapt the interactive example from the text-emphasis (in the same way that we do for [`margin`](https://developer.mozilla.org/en-US/docs/Web/CSS/margin) and its longhands).

rachelandrew:
Fixes #1644 

There were already examples for style and color, but they haven't been added to pages and needed the -webkit prefixed versions. So I have added that and also the text-emphasis-position example.

### DIFF
--- a/live-examples/css-examples/text-decoration/meta.json
+++ b/live-examples/css-examples/text-decoration/meta.json
@@ -56,6 +56,13 @@
             "title": "CSS Demo: text-emphasis-color",
             "type": "css"
         },
+        "textEmphasisPosition": {
+            "cssExampleSrc": "./live-examples/css-examples/text-decoration/text-emphasis-position.css",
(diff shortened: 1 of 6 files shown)


chrisdavidmills (approved):
LGTM; thanks @rachelandrew

PR #1807, opened 2021-05-01
Show that setter does set the property (line 10)

karlhorky:
Hi there! 👋  First of all, let me say that I really appreciate the work that's being put into all of the MDN web documentation, it's a really great resource for beginners and more experienced developers alike!

Just wanted to propose a quick PR to be a bit more instructive in this example - to show that the setter actually does change the value if it doesn't hit this error case with the odd number of monster eyes.

### DIFF
--- a/live-examples/js-examples/proxyhandler/proxyhandler-set.js
+++ b/live-examples/js-examples/proxyhandler/proxyhandler-set.js
@@ -19,3 +19,7 @@ proxy1.eyeCount = 1;
 
 console.log(proxy1.eyeCount);
 // expected output: 4
+
+proxy1.eyeCount = 2;
+console.log(proxy1.eyeCount);
+// expected output: 2


wbamberg (approved):
Thanks @karlhorky !

It's unfortunate that the Proxy examples are so long, that with this change, even using the big interactive example window size, this bit of the example will overflow. But I don't think it's really feasible to compress these examples any more without sacrificing readability, and I agree it is nice to demonstrate this aspect of the example.

karlhorky:
> this bit of the example will overflow

Ah, that is unfortunate, yeah...

What about this as an alternative for this example? It's at least a few less lines...

```
const monster = { eyeCount: 4 };

const proxy1 = new Proxy(monster1, {
  set(obj, prop, value) {
    if ((prop === 'eyeCount') && ((value % 2) !== 0)) {
      console.log('Monsters must have an even number of eyes');
    } else {
      return Reflect.set(...arguments);
    }
  }
});
proxy1.eyeCount = 1;
// expected output: "Monsters must have an even number of eyes"

console.log(proxy1.eyeCount);
// expected output: 4

proxy1.eyeCount = 2;
console.log(proxy1.eyeCount);
// expected output: 2
```

wbamberg:
This is shorter but to my eyes less readable. I like the way that in the current version the target, the handler, and the proxy are all separated out and named, so you can easily see how the proxy glues the handler to the target.

Maybe:

```
const monster1 = { eyeCount: 4 };

const handler1 = {
  set(obj, prop, value) {
    if ((prop === 'eyeCount') && ((value % 2) !== 0)) {
      console.log('Monsters must have an even number of eyes');
    } else {
      return Reflect.set(...arguments);
    }
  }
};

const proxy1 = new Proxy(monster1, handler1);

proxy1.eyeCount = 1;
// expected output: "Monsters must have an even number of eyes"

console.log(proxy1.eyeCount);
// expected output: 4

proxy1.eyeCount = 2;
console.log(proxy1.eyeCount);
// expected output: 2
```

? 23 lines, which does fit, and I get to waste a blank link after the `Proxy` constructor call, which I like for readability. Do you think there's any significant drawback to using an object literal form for `monster1`?

karlhorky:
Understood about the naming. Nice, I think this is good too! Should I do a pull request with that?

wbamberg:
> Understood about the naming. Nice, I think this is good too! I think the object literal is more clear than the class.
> 
> If you're also happy with it, should I do a pull request with that?

If you do, I will press merge :).

karlhorky:
Open: https://github.com/mdn/interactive-examples/pull/1810

PR #1808, opened 2021-05-02
Add smaller font-size for last element in grid-auto-rows demo

tannerdolby:
Resolves #1801 @wbamberg 

> Yes the "missing" text was intentional as otherwise you can't see the difference (it's a bit tricky in such a small space). I think the solution of adding a smaller font size would work.

In this [comment](https://github.com/mdn/interactive-examples/pull/1801#issuecomment-830572374), @rachelandrew mentioned that adding a smaller font size for the last element would be a way to resolve this. I'm happy to push another commit and remove the text content if its more preferred without text to really highlight the different heights the last box will occupy.

### DIFF
--- a/live-examples/css-examples/grid/grid-auto-rows.css
+++ b/live-examples/css-examples/grid/grid-auto-rows.css
@@ -12,8 +12,9 @@
 #example-element > div {
     background-color: rgba(0, 0, 255, 0.2);
     border: 3px solid blue;
+    font-size: 22px;
 }
 
 #example-element div:last-child {
-    font-size: 10px;
+    font-size: 13px;
 }
--- a/live-examples/css-examples/grid/grid-auto-rows.css
+++ b/live-examples/css-examples/grid/grid-auto-rows.css
@@ -13,3 +13,7 @@
     background-color: rgba(0, 0, 255, 0.2);
     border: 3px solid blue;
 }
+
+#example-element div:last-child {
+    font-size: 10px;
+}
--- a/live-examples/css-examples/grid/grid-auto-rows.html
+++ b/live-examples/css-examples/grid/grid-auto-rows.html
@@ -21,7 +21,7 @@
     </div> 
 
     <div class="example-choice">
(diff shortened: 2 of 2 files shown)


wbamberg:
Thank you for this change @tannerdolby !

From the original example, I *think* the desired effect is going to be:

A. `auto`: boxes 1,2,3,5 are the same, 4 is smaller, because height is being taken from content size
B. `50px`: all boxes are the same size
C. `min-content`: same as `auto`
D. `minmax(10px, auto)`:  boxes 1,2,3,5 are the same, 4 is smaller. BUT box 4 is bigger than in cases A and C, because the minimum height for box 4 is < 10 pixels.

In the example we get in this PR, cases A,B,D are all the same, I think because the minimum height for box 4 is > 10px.

So what we could do is something like:
* make the font in the other boxes bigger, say 24px
* make the code in example D something like `minmax(30px, auto)`: that way box 4 in example D is bigger than in A and C, because the minimum height for box 4 is < 30px

Does that make sense?

tannerdolby:
Your welcome @wbamberg !

I agree with the *desired* effects you've listed above.

Yeah that makes sense. I will commit some changes and we can go from there. Thanks for the feedback!

wbamberg:
Sounds great, thank you @tannerdolby 👍

tannerdolby:
Ping for review @wbamberg 

Increasing the font size for all but the last box and changing minmax to `minmax(30px, auto)` works really well to more clearly differentiate the last option from `auto` and `min-content`. I think this should be resolved now, let me know your thoughts.

wbamberg:
This looks great to me, thank you @tannerdolby . I'd also love to hear what @rachelandrew thinks.

tannerdolby:
Ok great! Thanks for the review @wbamberg !

wbamberg (approved):
Let's just merge this. Thanks for your work on it, @tannerdolby !

tannerdolby:
Your welcome @wbamberg !

Issue #1645, opened 2020-08-07
CSS examples: need example for tab-size


wbamberg:
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/tab-size

This looks quite straightforward. Note that it's prefixed in Firefox, so the interactive example choices need to set both `tab-size` and `-moz-tab-size`.

sahilempire:
I wanna work on this issue, Please assign this to me. @wbamberg

sahilempire:
Please tell me what exactly i have to do. As I am working first time for your organization so I don't have enough knowledge, so please guide me @chrisdavidmills

sahilempire:
Please tell me **what exactly i have to do**. As I am **working first time for your organization** so I don't have enough knowledge, so please **guide me** @chrisdavidmills

sahilempire:
Hello @chrisdavidmills ,

I have **successfully submitted my Pull Request**** for this issue, **Please check it**.

Thanks.

sahilempire:
Hi @chrisdavidmills ,

**Please help** me to go ahead with my pull request.
This shows **i have no write access** to this repository.

Please guide me.

Thank You.

Blakelist7:
@wbamberg I think issue is resolved(PR #1764) so you might close it.

wbamberg:
@Blakelist7 thanks for pointing this out. It's almost done but not quite. If you try it: https://interactive-examples.mdn.mozilla.net/pages/css/tab-size.html , it works in Chrome but in Firefox you get a weird static version.

This is because the live sample system tests that a property is supported in the browser before loading the editor, and falls back to the static version if the property is not supported.

Firefox needs the `-moz` prefix, and although the example code has been updated to use that, you also need to update the example metadata to tell the live samples system to test with the prefixed property as well.

See https://github.com/mdn/interactive-examples/blob/master/CONTRIBUTING-CSS.md#supporting-prefixed-properties for more on this.

In practice that would mean editing:

https://github.com/mdn/interactive-examples/blob/34006c974e69e26d54f57ee269cf95bb06c54101/live-examples/css-examples/text/tab-size.html#L1

...to make `data-property="tabSize -moz-tab-size"`, and then checking how it looks in Firefox.

Blakelist7:
ohk, I see I'll request a PR for the same
btw @wbamberg thanks for the hack! :smile:

Blakelist7:
[...]

PR #1811, opened 2021-05-08
Fix: tabSize example support 

Blakelist7:
Firefox supports tabSize example
#1645


wbamberg (approved):
This works great, thank you!

PR #1824, opened 2021-05-24
fix: change title attribute example

### LINKED ISSUE #1822: The title attribute example should be updated to not use abbr elements
The `title` attribute used on an `abbr` element has [numerous accessibility issues](https://www.tpgi.com/short-note-the-abbreviation-appreciation-society/). Can we update it to demonstrate a more appropriate usage?

I was thinking along the lines of:

```html
<iframe title="Wikipedia page for Tangerines" src="https://en.m.wikipedia.org/wiki/Tangerine"></iframe>
```

schalkneethling:
Change the `title` attribute interactive example as the current example use case is problematic from an accessibility perspective and is not the best use case to use for the example.

fix #1822

/cc @ericwbailey

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-title.html
+++ b/live-examples/html-examples/global-attributes/attribute-title.html
@@ -1,6 +1,6 @@
 <p>Use the <code>title</code> attribute on an <code>iframe</code> to clearly identify the content of the <code>iframe</code> to screen readers.</p>
 
-<div class="frame-container">
-  <iframe title="Wikipedia page for the HTML language" src="https://en.m.wikipedia.org/wiki/HTML"></iframe>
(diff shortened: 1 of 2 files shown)


ericwbailey:
Heck yeah! Thank you!

chrisdavidmills (changes requested):
Thanks @schalkneethling . I think the markup looks good; the only issue I have is that the layout doesn't look very good when you display the example at the size it will be seen embedded into MDN. Roughly something like this:

Screen Shot 2021-05-26 at 12 35 49 PM

I think it would look better overall to just show the iframes at 100% width:

Screen Shot 2021-05-26 at 12 36 10 PM

schalkneethling:
> I think it would look better overall to just show the iframes at 100% width:

Updated.

chrisdavidmills (approved):
Much better, cheers @schalkneethling !

PR #1827, opened 2021-05-30
Fix responsiveness of .example-container output from `flex` interactive demo

### LINKED ISSUE #1825: `flex` interactive example fails for a few different viewport width ranges
The `flex` shorthand property was something that took me awhile to fully understand and use. The examples for viewports > 900 - 905px and < 1024px are very clear and help to illustrate how the shrink, grow, and basis flex properties work. For viewports below 900px, each of the example options produce the same output as the example container has decreased in width and there isn't any available space left for the flex items to shrink or grow. Below 800px the content overflows the parent container and < 680px viewport width the example overflows more where only two flex items remain in the example output container.

There is a media query that kicks in and collapses the left sidebar at `(min-width: 63.9385em)`. This helps to make the `.example-container` have more space to occupy in its parent element, but from 63.9385em to 1110px the third box overflows and the example options all produce the same unchanged output.

From 1198px to 1024px, the examples produce the same output, and like stated above from 1110px to 1024px the third item overflows the `.example-container`.

https://user-images.githubusercontent.com/48612525/120115466-9eb62a80-c138-11eb-9f9d-05f4d2e05d3d.mov

Below 800px the example output overflows the parent container, this is before the media query kicks in and collapses the left sidebar which resizes the interactive example output container by giving it more space to occupy.

Screen Shot 2021-05-30 at 11 24 11 AM

Proposed fix(s): 
- Give `.example-container` element a `width: auto` and this allows the example to be much more responsive without writing any media conditions.
- Create media query to bump the `width` on `<div class="example-container">` from 80% to be a bit bigger maybe 5-10% bigger around 90% so the flex items have more space to shrink or fill too.

tannerdolby:
Fixes #1825

### DIFF
--- a/live-examples/css-examples/flexbox/flex.css
+++ b/live-examples/css-examples/flexbox/flex.css
@@ -2,7 +2,7 @@
     background-color: #eee;
     border: .75em solid;
     padding: .75em;
-    width: 80%;
+    width: auto;
     max-height: 300px;
     display: flex;
 }
@@ -19,4 +19,29 @@
 #example-element {
     background-color: rgba(255, 0, 200, .2);
     border: 3px solid rebeccapurple;
+}
+
+@media screen and (max-width: 760px) {
+
+    .editor-wrapper {
+        display: flex;
+        flex-direction: column;
+    }
+
+    .example-choice-list.large.live,
+    .example-choice-list.live {
+        width: auto;
+    }
+
+    /* Face the triangle down towards the example output */
+    .live .example-choice:before {
+        top: 35%;
+        transform: rotate(90deg);
+    }
+
+    #output {
+        height: auto;
+        width: auto;
+        margin-top: 10px;
+    }
 }
--- a/live-examples/css-examples/flexbox/flex.css
+++ b/live-examples/css-examples/flexbox/flex.css
@@ -1,4 +1,4 @@
-.example-container {
+.default-example {
     background-color: #eee;
     border: .75em solid;
     padding: .75em;
@@ -7,7 +7,7 @@
     display: flex;
 }
 
-.example-container > div {
+.default-example > div {
     background-color: rgba(0, 0, 255, .2);
     border: 3px solid blue;
     margin: 10px;
--- a/live-examples/css-examples/flexbox/flex.css
+++ b/live-examples/css-examples/flexbox/flex.css
@@ -20,28 +20,3 @@
     background-color: rgba(255, 0, 200, .2);
     border: 3px solid rebeccapurple;
 }
-
-@media screen and (max-width: 760px) {
-
-    .editor-wrapper {
-        display: flex;
-        flex-direction: column;
-    }
-
-    .example-choice-list.large.live,
-    .example-choice-list.live {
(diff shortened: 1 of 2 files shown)


tannerdolby:
I added a media query in my first commit so that the interactive examples shift to a column layout at smaller viewport widths like stated in [#1825](https://github.com/mdn/interactive-examples/issues/1825#issuecomment-851047341). The link for where its stated that this layout is implemented https://github.com/mdn/interactive-examples/blob/master/CSS-Example-Style-Guide.md#output-width-considerations

Do you want me to add the same media query to my other ~3 PRs where I remove the inner container? Or should we handle adding a media condition for the layout change to column for smaller screen sizes in one big batch? Since we also need to remove the inner container for essentially every interactive example page. Let me know your thoughts @schalkneethling !

schalkneethling:
> I added a media query in my first commit so that the interactive examples shift to a column layout at smaller viewport widths like stated in [#1825](https://github.com/mdn/interactive-examples/issues/1825#issuecomment-851047341). The link for where its stated that this layout is implemented https://github.com/mdn/interactive-examples/blob/master/CSS-Example-Style-Guide.md#output-width-considerations
> 
> Do you want me to add the same media query to my other ~3 PRs where I remove the inner container? Or should we handle adding a media condition for the layout change to column for smaller screen sizes in one big batch? Since we also need to remove the inner container for essentially every interactive example page. Let me know your thoughts @schalkneethling !

I think this is a great idea @tannerdolby but, we would need to get some input from others, make a call and then orchestrate with changes in Yari possibly. So, let’s remove that from this and any other related pull requests and create a dedicated issue to discuss this further. Thanks so much.

schalkneethling (commented):
Other than removing the media query piece, for now, this is good to go.

tannerdolby:
> I think this is a great idea @tannerdolby but, we would need to get some input from others, make a call and then orchestrate with changes in Yari possibly

Sounds good to me @schalkneethling ! If we don't end up making the changes, lets make sure to update the CSS Example style guide where it states that the column layout is shifted too at smaller viewports. 

>  So, let’s remove that from this and any other related pull requests and create a dedicated issue to discuss this further. Thanks so much.

Your welcome! I will add a commit to this PR and remove those media query additions.

schalkneethling:
> Sounds good to me @schalkneethling ! If we don't end up making the changes, lets make sure to update the CSS Example style guide where it states that the column layout is shifted too at smaller viewports.

Agreed. Also, while working on something else I believe I found why this may not be happening anymore. We used to have the communication between the `iframe` and the parent document via `postMessage` that would cause the code in the `iframe` to toggle a class on the container.

This is no longer happening 😱 - Perhaps we will just resurrect that, or decide on a completely new approach.

schalkneethling (approved):

schalkneethling:
@tannerdolby Do we have an existing issue we can use for the discussion around how to approach this column layout switching, or should we create one? Thanks!

tannerdolby:
@schalkneethling Yeah we can move this discussion to the existing issue #1834. No problem!

PR #1829, opened 2021-05-30
Add auto width to .example-container for flex-basis interactive demo

### LINKED ISSUE #1828: Add auto width to .example-container of flex-basis interactive demo
This issue stems from #1825 where the `width: 80%` for `.example-container` in the output of interactive demo has width ranges where the options don't produce any change in the output, which is because of the available space they have to shrink/fill to. 

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis

I'm not going to include a video as this problem is the same as whats shown in the linked issue. Can easily be verified by navigated to the page and resizing the window then clicking options and seeing the output not change at all.

tannerdolby:
Fixes #1828 

The "side by side" layout doesn't change, but we could perform the same column layout shift like I proposed in #1825 and that could be one more commit added here before this PR is ready to merge.

### DIFF
--- a/live-examples/css-examples/flexbox/flex-basis.css
+++ b/live-examples/css-examples/flexbox/flex-basis.css
@@ -1,4 +1,4 @@
-.example-container {
+.default-example {
     background-color: #eee;
     border: .75em solid;
     padding: .75em;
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
🎉

PR #1831, opened 2021-05-30
Add auto width to .example-container for flex-grow interactive demo

### LINKED ISSUE #1830: Add auto width to .example-container for flex-grow interactive demo
Related to #1825

Having the `.example-container` width be 80% causes there to be many areas where the selected options don't change the interactive code demo output, making it quite difficult to understand how each option behaves. Same issue as the other `flex-*` interactive demos, it doesn't shift to the column layout for smaller viewports beneath tablet size. But the implementation doesn't change from #1825 so it can easily be added with another commit on this issues related PR.

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow

tannerdolby:
Fixes #1830

### DIFF
--- a/live-examples/css-examples/flexbox/flex-grow.css
+++ b/live-examples/css-examples/flexbox/flex-grow.css
@@ -1,4 +1,4 @@
-.example-container {
+.default-example {
     background-color: #eee;
     border: .75em solid;
     padding: .75em;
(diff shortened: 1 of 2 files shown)


schalkneethling (approved):
🎉 🥦

PR #1833, opened 2021-05-30
Add auto width to .example-container for flex-shrink demo and media query for column layout

### LINKED ISSUE #1832: Fix width issue for .example-container demo output on flex-shrink page
Related to #1825 

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink

tannerdolby:
Fixes #1832

### DIFF
--- a/live-examples/css-examples/flexbox/flex-shrink.css
+++ b/live-examples/css-examples/flexbox/flex-shrink.css
@@ -15,28 +15,3 @@
     flex-shrink: 1;
     flex-basis: 300px;
 }
-
-@media screen and (max-width: 760px) {
-
-    .editor-wrapper {
-        display: flex;
-        flex-direction: column;
-    }
-
-    .example-choice-list.large.live,
-    .example-choice-list.live {
-        width: auto;
-    }
-
-    /* Face the triangle down towards the example output */
-    .live .example-choice:before {
-        top: 35%;
-        transform: rotate(90deg);
-    }
-
-    #output {
-        height: auto;
-        width: auto;
-        margin-top: 10px;
-    }
-}
--- a/live-examples/css-examples/flexbox/flex-shrink.css
+++ b/live-examples/css-examples/flexbox/flex-shrink.css
@@ -1,4 +1,4 @@
-.example-container {
+.default-example {
     background-color: #eee;
     border: .75em solid;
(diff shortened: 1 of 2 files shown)


schalkneethling:
@tannerdolby Thanks so much for digging into this. I am wondering, is there really any benefit to that inner container? What if we made it like the example below:

https://user-images.githubusercontent.com/10350960/120669802-12856980-c490-11eb-845c-146764d40b3b.mov

tannerdolby:
@schalkneethling Your welcome! I just came across it recently as felt we should provide a fix for it.

> I am wondering, is there really any benefit to that inner container? What if we made it like the example below:

To be honest, I don't think there is any benefit to that inner container. If we removed it, there would be much more available space for the demo output to occupy. Keeping the inner container and using `width: auto` like I suggested would work, but feels like a bandaid. Removing the inner container completely seems like the more sturdy fix.

schalkneethling:
> Removing the inner container completely seems like the more sturdy fix.

Let’s do it. Thanks, @tannerdolby

tannerdolby:
Ok deal. Your welcome @schalkneethling ! I will update my ~3 other PRs and start chugging away at removing inner containers for the `/flex/` interactive examples. 

Since almost all pages in `mdn/interactive-examples` need to have the inner container removed, would you prefer multiple PRs or one PR with many commits. Let me know! 👍

schalkneethling:
> Since almost all pages in `mdn/interactive-examples` need to have the inner container removed, would you prefer multiple PRs or one PR with many commits?

Let’s do one with multiple commits.

live-examples/css-examples/flexbox/flex-shrink.css
```diff
@@ -1,17 +1,42 @@
...
+        top: 35%;
+        transform: rotate(90deg);
+    }
+
+    #output {
+        height: auto;
+        width: auto;
+        margin-top: 10px;
+    }
+}
```
schalkneethling, line 42: As mentioned in another PR, I think this is a great idea but, let’s split it out of these changes and come to a consensus of how we can deal with this from a larger perspective and then make the changes based on our conclusion. Thanks!

schalkneethling (approved):
🚀

PR #1835, opened 2021-05-31
Adds interactive example for Promise.finally()

### LINKED ISSUE #1653: JS examples: need example for Promise.finally()
https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

Inspiration could possibly be got by looking at some of the other `Promise` examples, especially:

https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then
https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

tannerdolby:
Fixes #1653 

https://user-images.githubusercontent.com/48612525/120126420-0d16df00-c171-11eb-9746-ce7715c4c194.mov

### DIFF
--- a/live-examples/js-examples/promise/promise-finally.js
+++ b/live-examples/js-examples/promise/promise-finally.js
@@ -1,0 +1,17 @@
+function hasMailArrived() {
+  return new Promise((resolve, reject) => {
+    (Math.random() > 0.5 && resolve('Mail has arrived')) ||
+        reject(new Error('Failed to arrive'));
+  });
+}
+
+hasMailArrived()
+  .then((mail) => {
+    console.log(mail);
+  })
+  .catch((err) => {
+    console.error(err);
+  })
+  .finally(() => {
+    console.log('Experiment completed');
+  });
--- a/live-examples/js-examples/promise/promise-finally.js
+++ b/live-examples/js-examples/promise/promise-finally.js
@@ -1,11 +1,14 @@
-function hasMailArrived() {
+function checkMail() {
   return new Promise((resolve, reject) => {
-    (Math.random() > 0.5 && resolve('Mail has arrived')) ||
-        reject(new Error('Failed to arrive'));
+    if (Math.random() > 0.5) {
+      resolve('Mail has arrived');
+    } else {
+      reject(new Error('Failed to arrive'));
+    }
   });
 }
 
-hasMailArrived()
+checkMail()
   .then((mail) => {
     console.log(mail);
   })
--- a/live-examples/js-examples/promise/meta.json
+++ b/live-examples/js-examples/promise/meta.json
@@ -53,6 +53,12 @@
             "fileName": "promise-then.html",
             "title": "JavaScript Demo: Promise.then()",
             "type": "js"
+        },
+        "promiseFinally": {
+            "exampleCode": "./live-examples/js-examples/promise/promise-finally.js",
+            "fileName": "promise-finally.html",
+            "title": "JavaScript Demo: Promise.finally()",
+            "type": "js"
         }
     }
 }


tannerdolby:
I think the PR that was original submitted on Nov 3, 2020 has become stale so I submitted a nice new PR for the `Promise.finally()` interactive example. Since I saw your comment last, I'm pinging you for review @peterbe !

live-examples/js-examples/promise/promise-finally.js
```diff
@@ -0,0 +1,17 @@
+function hasMailArrived() {
+  return new Promise((resolve, reject) => {
+    (Math.random() > 0.5 && resolve('Mail has arrived')) ||
+        reject(new Error('Failed to arrive'));
```
peterbe, line 4: I'm not sure it's my place to "correct" but that syntax feels a bit too fancy. What about:
```suggestion
        if (Math.random() > 0.5) {
          resolve('Mail has arrived');
        } else {
          reject(new Error('Failed to arrive'));
        }
```

peterbe: Clearly, I don't know how to make suggestions here in the UI. But I hope it makes sense.

live-examples/js-examples/promise/promise-finally.js
```diff
@@ -0,0 +1,17 @@
+function hasMailArrived() {
```
peterbe, line 1: Does that name match the code as well as I could?
If turned into a question, "Has mail arrived?" you'd expect the answer to be "Yes" or "No". 
But the function returns a Promise and it has its own side-effect (the `console.log('Experiment completed');`).

It's not a great name either, but perhaps `checkMail()`? What do you think?

peterbe (changes requested):
Great work!! I can't believe we didn't have one of these. 
I've never actually reviewed an interactive-example PR. 

But I think the name needs some thought and the conditional would read better if it was laid out more plainly.

tannerdolby:
Your welcome! I couldn't believe there wasn't one of these already too. This is the third or fourth interactive example I've created/edited so I'm starting to get the hang of things.

> But I think the name needs some thought and the conditional would read better if it was laid out more plainly.

Ya, this was just what I spun up from your `catInBox` example in the comments of that issue. I will have a look at requested changes and add some commits.

live-examples/js-examples/promise/promise-finally.js
tannerdolby: I was thinking the same thing, function name could match better for what it is actually returning, which is `Promise<any>` but is expected to be `Promise<String>`. After turning it into a question, it becomes more clear the name could match the code better. 

> It's not a great name either, but perhaps checkMail()? What do you think?

Agreed, name could be better overall. I think `checkMail()` has a nice ring too it.

live-examples/js-examples/promise/promise-finally.js
tannerdolby: > Clearly, I don't know how to make suggestions here in the UI. But I hope it makes sense.

Yeah it made sense no worries! 😸

peterbe:
> your `catInBox` example in the comments of that issue.

haha!! I know I didn't put much thought into that originally but that's the only thing I can remember. 
Hey, at least I had the idea of calling it "inbox" twice :)

peterbe (approved):
I don't think I've ever reviewed an interactive example that isn't just a light typo fix. I think I'll rope in @wbamberg for a second pair of eyes. 
But I like the code. It's short and simple and works.

tannerdolby:
> I don't think I've ever reviewed an interactive example that isn't just a light typo fix. I think I'll rope in @wbamberg for a second pair of eyes. But I like the code. It's short and simple and works.

Ok sounds good to me, and thanks! I thought `checkMail` worked well with the simplified `if/else` logic. Excited to see this make it onto the site 🎉

wbamberg (approved):
This looks great, thank you @tannerdolby !

To see it on the site you'll need to add something like:

    <div>{{EmbedInteractiveExample("pages/js/promise-finally.html", "taller")}}</div>

...to https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/promise/finally/index.html, a bit like:

https://github.com/mdn/content/blob/3214396a53b3a9ac4aebd2e69f1474d2912a5f00/files/en-us/web/javascript/reference/global_objects/promise/race/index.html#L18

tannerdolby:
Your welcome @wbamberg !

I submitted a PR to add the kumascript macro for displaying the interactive example over in `mdn/content` ([5607](https://github.com/mdn/content/pull/5607)), so once thats merged this should be on the live site.

PR #1836, opened 2021-05-31
Drop role=main from <main> element

sideshowbarker:
Per the ARIA in HTML spec at https://w3c.github.io/html-aria/#el-main, the `main` element should have no role value; its implicit role is `main`, and at https://w3c.github.io/html-aria/#rules-wd, the spec says:

> It is NOT RECOMMENDED for authors to set the ARIA role and aria-\* attributes to values that match the implicit ARIA semantics defined in the table. Doing so is unnecessary and can potentially lead to unintended consequences.

Fixes https://github.com/mdn/content/issues/5461

### DIFF
--- a/live-examples/html-examples/text-content/main.html
+++ b/live-examples/html-examples/text-content/main.html
@@ -1,6 +1,6 @@
 <header>Gecko facts</header>
 
-<main role="main">
+<main>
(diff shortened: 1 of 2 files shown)


wbamberg (approved):
👍 thanks @sideshowbarker !

PR #1839, opened 2021-06-02
Use 32 bit int in examples

pinksynth:
The examples refer to `-3` as `1111111111111101`, which is valid for 16-bit signed integers, but not for 32-bit signed integers, which is what the operator uses. Using 32 bit integers, `1111111111111101` is 65,533.

Addresses mdn/content#5599


sideshowbarker (approved):

sideshowbarker:
I’ve reviewed and approved this PR, but I unfortunately don’t have perms to merge it…

sideshowbarker:
Thanks @peterbe

PR #1851, opened 2021-06-15
Significantly improve by adding an ALL CAPS word

Gitoffthelawn:
This example will benefit significantly by adding a word in ALL CAPS so developers can see how an all caps word is affected.

Fortunately, the first word of the quote from Dickens affords this opportunity perfectly.  According to [GoodReads](https://www.goodreads.com/quotes/195229-london-michaelmas-term-lately-over-and-the-lord-chancellor-sitting), the word "London" should be in all caps, so this improvement is a win-win.

### DIFF
--- a/live-examples/css-examples/text/text-transform.html
+++ b/live-examples/css-examples/text/text-transform.html
@@ -47,7 +47,7 @@
 <div id="output" class="output large hidden">
     <section id="default-example">
         <div id="example-element" class="transition-all">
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
Sorry to be slow here. Yes, this looks like a useful change. Thanks, @Gitoffthelawn !

Gitoffthelawn:
@wbamberg You're welcome! No problem at all for the delay.  I'm happy to help!

Issue #1858, opened 2021-06-27
Issue with "Right shift (>>)": (comment in JavaScript Demo: Expressions - Right shift operator has problem)


blateyang:
https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/expressions/expressions-right-shift.js

MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Right_shift

#### What information was incorrect, unhelpful, or incomplete?
Hi, I found the comment of `console.log(c >> b); ` in JavaScript Demo: Expressions - Right shift operator as follows has problem

#### Specific section or headline?

```
const a = 5;          //  00000000000000000000000000000101
const b = 2;          //  00000000000000000000000000000010
const c = -5;         // -00000000000000000000000000000101

console.log(a >> b);  //  00000000000000000000000000000001
// expected output: 1

console.log(c >> b);  // -00000000000000001111111111111110
// expected output: -2
```
#### What did you expect to see?
console.log(c >> b);  // -00000000000000000000000000000010

<details>
<summary>MDN Content page report details</summary>

* Folder: `zh-cn/web/javascript/reference/operators/right_shift`
* MDN URL: https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Reference/Operators/Right_shift
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/zh-cn/web/javascript/reference/operators/right_shift/index.html
* Last commit: https://github.com/mdn/translated-content/commit/52ccdf1f0a1cee3f0659ac78e81c79dae152399b
* Document last modified: 2021-06-23T00:34:06.000Z

</details>

sideshowbarker:
@t7yang Thanks very much for clarifying

@SaurabhCha Assigned to you now. As @t7yang noted, the fix for this one needs to be made not in the mdn/content repo but instead to the interactive example source at https://github.com/mdn/interactive-examples/blob/master/live-examples/js-examples/expressions/expressions-right-shift.js

SaurabhCha:
@sideshowbarker Thank you for assigning the issue to me , I will make the change and open the PR.

@t7yang Thank you for clarifying.

SaurabhCha:
@sideshowbarker Could you please approve the below PR ?
https://github.com/mdn/interactive-examples/pull/1859
It hasn't updated it this ticket.

PR #1859, opened 2021-06-29
Corrected the binary value of -2

SaurabhCha:
The commented binary value of -2 was incorrect. I have updated the correct value.

Fixes https://github.com/mdn/interactive-examples/issues/1858


SaurabhCha:
Updated the incorrect value.

Rumyra (approved):

PR #1864, opened 2021-07-11
Add live example for `accent-color` css property.

lukewarlow:
Added a live example of the `accent-color` property from https://drafts.csswg.org/css-ui-4/#widget-accent. To test you need Chrome 91+ with the `experimental-web-platform-features` flag enabled, or Firefox 90+ with the `layout.css.accent-color.enabled` flag enabled.

Shipping on by default in Chrome 93 and Firefox 92.

### DIFF
--- a/live-examples/css-examples/basic-user-interface/accent-color.css
+++ b/live-examples/css-examples/basic-user-interface/accent-color.css
@@ -1,4 +1,14 @@
+.container > div {
+    display: flex;
+    align-items: center;
+}
+
 #example-element {
     width: 40px;
     height: 40px;
+}
+
+#example-label {
+    margin-left: 10px;
+    font-size: x-large;
 }
--- a/live-examples/css-examples/basic-user-interface/accent-color.css
+++ b/live-examples/css-examples/basic-user-interface/accent-color.css
@@ -1,0 +1,4 @@
+#example-element {
+    width: 40px;
+    height: 40px;
+}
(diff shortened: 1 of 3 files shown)


lukewarlow:
How long after the release of the property in Chrome and Firefox should I wait before making a new MR to mdn/content to include this live example?

rachelandrew:
@lukewarlow once it is in two stable browsers then that would be the time. So as soon as both are the stable version.

wbamberg (commented):
This is a nice example, thank you!

I think it might be easier to understand how `accent-color` applies if the example shows a checkbox in context, like:

Screen Shot 2021-09-07 at 2 27 25 PM

What do you think?

lukewarlow:
Personally not sure how much extra value a label would bring, the main thing to me is seeing how browsers behave with darker and lighter colours. Observing stuff like the glyph changing colour?

wbamberg:
Because it reinforces that the thing being demonstrated is a checkbox, which is usually associated with a label.

wbamberg:
Sorry, that was a bit terse. What I mean is: when I look at this example it's not totally obvious that I'm looking at `<input type="checkbox">`, partly because it's really big but also because in my experience checkboxes are almost always associated with a label telling you what they are a checkbox for. For instance, it could just be an image that the property is colouring. Adding a label just makes that clearer, to me.

lukewarlow:
Actually thinking about it you're probably right. I know it's a checkbox but others looking for the first time might not immediately know. I've added an example label now. Let me know if there's any other tweaks you think would be good.

wbamberg (approved):
👍 thank you @lukewarlow !

PR #1868, opened 2021-07-15
Remove comment that the order is not guaranteed with `Object.entries()`

andreubotella:


Elchi3 (approved):
Thanks! I'm glad this works now

(I've read https://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties to confirm)

andreubotella:
I should probably have linked to mdn/content#6937, where I explained why that was the case.

PR #1876, opened 2021-07-27
use start and end instead of left and right for the text-align property

rachelandrew:
A small fix based on this content issue: https://github.com/mdn/content/issues/7295


teoli2003 (approved):

teoli2003:
I can't merge, but this is an obvious change after the content PR.

PR #1880, opened 2021-08-02
CONTRIBUTING - suggest running yarn test

hamishwillee:
Perfectly valid javascript can still fail to merge due to the linter - that will, for example, cause a fail based on snake rather than camel case for property names. This suggests running `yarn test` before submitting to CI. We might perhaps instead/as well point to linter rules, but this is easier. 

Note, this actually doesn't work well on Windows, which gives you lots of spurious line ending related errors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,11 +83,14 @@ Point your browser to either of those URLs, then click on the `pages` link. In t
 -   JavaScript examples are under `js`
 -   HTML examples are under `tabbed`
 
-Find your example and try it out. Once you're satisfied, the final step is to confirm that your change passes the same automatic tests that will be run by our continuous integration system.
-```
-yarn test
-```
-If the tests pass, [submit your pull request](https://help.github.com/articles/creating-a-pull-request/).
+Find your example and try it out.
+
+> **Note** On Linux, you might also run the automatic tests used by our continuous integration system.
+>  ```
+>  yarn test
+>  ```
+
+Once you're satisfied, the final step is to [submit your pull request](https://help.github.com/articles/creating-a-pull-request/).
 
 ## Publishing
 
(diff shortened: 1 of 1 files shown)


peterbe:
`yarn test` doesn't work on Windows? Then perhaps not make that suggestion :)

hamishwillee:
@peterbe OK, I have reworded to suggest this only for Linux "as an option", where it works well.

"a" problem here is that the linter requires LF rather than CRLF line endings, which are automatically added by default on Windows. So it might be that the simple fix here is to add a .gitattributes file that forces the right line endings?

peterbe:
> "a" problem here is that the linter requires LF rather than CRLF line endings, which are automatically added by default on Windows. So it might be that the simple fix here is to add a .gitattributes file that forces the right line endings?

A lot of people who use an IDE on Windows have figured out a way to force it to behave like it's Linux by using LF. 

It's good that the linter disallows CRLF so there's nothing wrong with the `yarn test` on Windows. Just "badly configured" editors? 

Ideal would be if you could instead point the reader to some tips about how to configure your IDE, on Windows, to use CR everywhere.

hamishwillee:
I work on heaps of projects and only very rarely does this line-ending issue bite me. In those cases, unless I'm paid, I move on - not worth the hassle to contribute. 
I don't think I'm alone, which is why I think the right fix for any project that requires particular line endings is to force those line endings using `.gitattributes`. If you disagree, just close this - I will move on :-)

> there's nothing wrong with the yarn test on Windows. Just "badly configured" editors?

I am not sure - I think it may be more than this. Right now all I see is a message that all the files are invalid, offering me the option to call -fix to correct the line endings. I suspect that it stops on the first error. Could be wrong, but since I know that the line ending thing is enough to block me I'm not inclined to continue.

peterbe:
@hamishwillee We totally got off-topic here but I'd like to announce that, from now on, every PR will run linting thanks to https://github.com/mdn/interactive-examples/pull/1889

schalkneethling:
> I don't think I'm alone, which is why I think the right fix for any project that requires particular line endings is to force those line endings using `.gitattributes`.

I would be happy to accept this. @Rumyra Thoughts?

Rumyra:
Yes!

PR #1890, opened 2021-08-16
Updated code bug

Biki-das:
the eariler code,at line no.3 had a return of 'process as infinity' with a negative value attached to it, this is a false operation for MIN_VALUE, which represents the adjacent positive value to 0, and not the most negative number


AnilSeervi:
Original Issue mdn/content#7975
Perhaps you may coordinate with original creator of the issue and get his opinion.

teoli2003 (approved):
LGTM!

AnilSeervi:
Both the inputs to the changes you've made result in `false` and output the multiplied value.
I did suggest you to coordinate with the OP to get his insight.
Nevertheless I've opened #1892 with the suggestions of the OP.

PR #1896, opened 2021-08-26
Add position fixed example

shapkarin:
Hello. 
At page https://developer.mozilla.org/en-US/docs/Web/CSS/position at interactive example there was no `position: fixed;`.
Hope it fits well in iframe and page. Because I didn't check locally with full setup by `mdn/yari`.
.env with INTERACTIVE_EXAMPLES_BASE_URL didn't help to point iframe to the localhost.

### DIFF
--- a/live-examples/css-examples/positioned-layout/position.html
+++ b/live-examples/css-examples/positioned-layout/position.html
@@ -22,6 +22,15 @@ top: 40px; left: 40px;</code></pre>
         </button>
     </div>
 
+
+    <div class="example-choice">
+        <pre><code class="language-css">position: fixed;
+top: 164px;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
     <div id="choice-sticky" class="example-choice">
         <pre><code class="language-css">position: -webkit-sticky;
 position: sticky;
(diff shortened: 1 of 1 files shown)


shapkarin:
Hello. Is anyone saw current PR? Maybe someone can review it or I break some contribution rules?

wbamberg:
Sorry @shapkarin to be slow responding.

I've tried out your example and it works well. And I agree it is a shame not to show `fixed`. 

The problem is that with these interactive examples we have a strictly limited amount of space for them. Adding this example gives us a scrollbar for the examples, which I really want to avoid. I don't really see how we can get around this without making some other compromises that I'd like tio avoid.

The interactive examples don't aspire to show every value of a property, just a quick idea of the main usages. We have the whole page to get into all the other details.

schalkneethling:
Hey @shapkarin, have you seen @wbamberg’s response? https://github.com/mdn/interactive-examples/pull/1896#issuecomment-931623797

shapkarin:
[schalkneethling](https://github.com/schalkneethling) Hi! Sure. I see my PR was merged, so maybe let's close this issue?

schalkneethling:
> [schalkneethling](https://github.com/schalkneethling) Hi! Sure. So what the solution? Just close this PR?

Yeah, I do believe so. Thank you so much for contributing. Feel free to look at the [other known issues](https://github.com/mdn/interactive-examples/issues) here and let me know if there is anything that looks interesting to you.

PR #1898, opened 2021-09-02
add example for transform: perspective(none);

mix5003:
from this https://github.com/w3c/csswg-drafts/issues/6488. 
i think it should add example for `transform: perspective(none);` to be identity transform
i not sure should i remove `transform: perspective(0);` or keep it to show that now behavior of it already change


Rumyra (approved):

Rumyra:
Thanks @mix5003 - agreed this is a good addition to illustrate the difference between `none` and `0`

PR #1904, opened 2021-09-11
Fix moon landing date

teoli2003:
Fixes mdn/content#8831

(Note this really is the UTC time and not the GMT one, but it will add confusion to translate it (UTC and GMT are off by a few seconds)


Elchi3 (approved):
https://www.youtube.com/watch?v=nOcDftgR5UQ

PR #1911, opened 2021-09-18
Fix vendor prefix order for text-emphasis

AnilSeervi:


caugner:
Hi @AnilSeervi, thank you for taking the time to submit this PR and sorry it took so long to hear back! 👋 

The vendor prefixes have meanwhile been removed from the example, making this PR obsolete, so I will go ahead and close it.

Thanks again for contributing to MDN!

PR #1912, opened 2021-09-21
Add Intl.supportedValuesOf() example

hamishwillee:
This adds a example of `Intl.supportedValuesOf()` which delivers in FF93. This is part of doing the work for: https://github.com/mdn/content/issues/8618


live-examples/js-examples/intl/intl-supportedvaluesof.js
```diff
@@ -0,0 +1,14 @@
...
+console.log(Intl.supportedValuesOf('currency'));
+console.log(Intl.supportedValuesOf('numberingSystem'));
+console.log(Intl.supportedValuesOf('timeZone'));
+console.log(Intl.supportedValuesOf('unit'));
+// expected output: Array ['key'] (for each key)
+
+try {
+  Intl.supportedValuesOf('someInvalidKey');
+} catch (err) {
+  console.log(err.toString());
```
hamishwillee, line 12: REVIEWER: Note that I was forced to do `toString` here to get a string - without this the output is an object. That seems like a bug in the example infrastructure

live-examples/js-examples/intl/intl-supportedvaluesof.js
```diff
@@ -0,0 +1,14 @@
+console.log(Intl.supportedValuesOf('calendar'));
+console.log(Intl.supportedValuesOf('collation'));
+console.log(Intl.supportedValuesOf('currency'));
+console.log(Intl.supportedValuesOf('numberingSystem'));
+console.log(Intl.supportedValuesOf('timeZone'));
+console.log(Intl.supportedValuesOf('unit'));
+// expected output: Array ['key'] (for each key)
```
hamishwillee, line 7: I know this format is non standard - usually you would have the comment line after each call. However ALL of these are the same so the extra line just pads things out. What this does though is mean that I can show the entire use of the API within the allocated space. Which I see as a good thing.

Rumyra (approved):

PR #1918, opened 2021-09-28
Add class static initialization block example

hamishwillee:
This adds examples for class static initialization block, added in FF93 . This will appear in docs here: https://github.com/mdn/content/pull/9281

There is a lot that might be shown. This primarily just shows the syntax for declaring a block, and that it can access/declare a property using the `this`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,5 @@
 js/lib/*
 js/mode/*
+
+# js class static blocks is cuurently ignored as it fails test
+live-examples/js-examples/classes/classes-static-initialization.js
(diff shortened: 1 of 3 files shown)


Rumyra:
So my theories on tests failing are:

* No new line at end of file (see review)
* JSlint actually can't parse a static block(?)

I'm not sure of the implications of merging this if the second point is true 🤔

hamishwillee:
As per the discussion you had with Will:

> Looks like it might be https://github.com/eslint/eslint/issues/15016 ? If so I think we should disable eslint for this example, but remember to enable it again when support for `static` lands in eslint.

I am presuming no action from me. Happy to do this, but would need instruction on how.

Rumyra (approved):

PR #1929, opened 2021-10-23
fix functional notation (rgb to rgba)

keikoro:
for code examples feat. RGB colours for which four values are currently provided

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/border-bottom-color.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-bottom-color.html
@@ -15,7 +15,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-bottom-color: rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-bottom-color: rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-bottom.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-bottom.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-bottom: 4mm ridge rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-bottom: 4mm ridge rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-left-color.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-left-color.html
@@ -15,7 +15,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-left-color: rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-left-color: rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-left.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-left.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-left: 4mm ridge rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-left: 4mm ridge rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-right-color.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-right-color.html
@@ -15,7 +15,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-right-color: rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-right-color: rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-right.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-right.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-right: 4mm ridge rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-right: 4mm ridge rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-top-color.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-top-color.html
@@ -15,7 +15,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-top-color: rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-top-color: rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/backgrounds-and-borders/border-top.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-top.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">border-top: 4mm ridge rgb(170, 50, 220, .6);</code></pre>
+        <pre><code class="language-css">border-top: 4mm ridge rgba(170, 50, 220, .6);</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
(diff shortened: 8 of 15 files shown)


keikoro:
Hi, are you even actively accepting PRs? I can see PRs by non-members from last year which haven't received any comments/feedback yet... @wbamberg @schalkneethling

wbamberg:
> Hi, are you even actively accepting PRs? I can see PRs by non-members from last year which haven't received any comments/feedback yet... @wbamberg @schalkneethling

Sorry to be slow reviewing your PR, @keikoro . Actually I haven't been actively monitoring this repo since August 2020, but I should probably start again, you're right.

I think the current examples are also correct though? At least according to https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb() it seems to be OK to pass an alpha value?

schalkneethling:
> Hi, are you even actively accepting PRs? I can see PRs by non-members from last year which haven't received any comments/feedback yet... @wbamberg @schalkneethling

Same here. Apologies for taking so long to give feedback here. Our focus was on other items. Thankfully we now have more time to focus on platform and related work so, your pull request will definitely get more attention going forward.

keikoro:
@wbamberg Thanks for the update, and ohh, interesting. I didn't know that `rgb()` supported four values.

I originally encountered a mismatch between code given as an example and what actually happens in the source code. The fact my IDE flags up usages of rgb with four values and that Firefox itself also autocorrects `rgb` to `rgba` when a fourth value is added when editing styles via the developer tools Inspector is definitely not helping.

I just had a look at the original example and saw there are even more mismatches between displayed example code and rendered stylings – a hex colour value is substituted with an rgb value, rgb and hsl values show up as rgba values. I haven't been able to track down the code that converts the CSS examples, so don't know what happens "behind the scenes", but would have expected the CSS in the source code to match the examples 1:1. Why doesn't it, and can this be rectified somehow? Consistency between examples + source code is ultimately what I wanted to achieve with the PR, even if I went about it the wrong way (having been misled by the tools I use). (;

schalkneethling:
> The fact my IDE flags up usages of rgb with four values

Which IDE are you using? Is this stylelint that complains?



> Firefox itself also autocorrects `rgb` to `rgba` when a fourth value is added when editing styles via the developer tools Inspector is definitely not helping.

Which version of Firefox are you using? I just tried editing a hex colour to `rgb(0,0,0,.5);` and Firefox devtools did not auto correct to `rgba`.

> a hex colour value is substituted with an rgb value, rgb and hsl values show up as rgba values.

I assume this is in devtools? I wonder, you can change the colour mode shown by shift clicking on a colour value in the dev tools. Not sure if there is a default that is configurable though. I would expect it to, by default, display it in the format that is defined in the CSS.

> I haven't been able to track down the code that converts the CSS examples, so don't know what happens "behind the scenes"

We definitely do not do any transforms on the CSS so, the CSS you see in this repo, is the CSS we ship to the browser.

keikoro:
> Which IDE are you using? Is this stylelint that complains?

I'm using Webstorm version 2021.2.1, without any CSS-related linting or plugins enabled, as far as I'm aware. I'm assuming the built-in checks default to CSS3 whereas rgb with four values seems to be a CSS4 thing. I don't know if there is a way to manually change CSS versions, I only know how/where to switch JS versions.

> Which version of Firefox are you using? I just tried editing a hex colour to `rgb(0,0,0,.5);` and Firefox devtools did not auto correct to `rgba`.

I'm using Firefox version 94.0.1 on macOS 10.14. I don't see anything in the devtools settings that would suggest I'm interfering with what Firefox defaults to.

When I enter `rgb(120,120,120,.3);` for a border color in the `element {}` section in `Devtools > Inspector > Filter Styles`, it gets corrected to `rgba(120, 120, 120, 0.3);` (note FF changes three things here: it switches the notation and adds spaces after the commas and a leading zero to the decimal).

I also tried your example, `rgb(0,0,0,.5);` with the same result: it gets changed to `rgba(0, 0, 0, 0.5);` for me. When I try with `rgb(0,0,0,1);`, however, FF doesn't interfere in any way (leaves both the notation and commas not followed by spaces alone).

> I assume this is in devtools? I wonder, you can change the colour mode shown by shift clicking on a colour value in the dev tools. Not sure if there is a default that is configurable though. I would expect it to, by default, display it in the format that is defined in the CSS.

Everything I've described with regard to Firefox was/is happening in devtools. And yeah, there's a devtools setting for Inspector which says "Default color unit" + several options via dropdown, which might be what you're referring to here. The selected dropdown value on my end is "As Authored".

I also checked with developer tools-equipped Chrome, Opera and Safari on the same machine, and I'm seeing the exact same discrepancies in their Inspectors. The pages on [border-bottom-color](https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-color) and [border-color](https://developer.mozilla.org/en-US/docs/Web/CSS/border-color) are two examples where the `style` attribute of the box displayed on the right contains different notations when looked at in Inspector than what the selected/active example on the left says. The only other machine I can currently test this on is sadly an even older Mac with a pretty outdated (esr) FF version (results are the same).

When I look at the source for the HSL table cells on the page on [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value), on the other hand, they look fine. E.g. the first one, for `hsl(0, 100%, 50%)` shows up as `<td style="background: hsl(0, 100%, 50%)"></td>` in Inspector.

wbamberg:
OK, so I think the issue you're seeing is that the CSS value shown in the editor uses a different form from the value the devtools are telling you? I see this as well.

For example looking at https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-color, if I select the third choice `border-bottom-color: rgb(170, 50, 220, .6);`, then I see `rgba(170, 50, 220, 0.6)` in the Firefox inspector:

Screen Shot 2021-11-08 at 2 31 16 PM

Is this the discrepancy you are talking about? If so: that is weird.

But I don't think changing the source to match is the right thing to do. For one thing, there are other cases where this is also happening. For example in the page for https://developer.mozilla.org/en-US/docs/Web/CSS/color, if I select the second option `color: #00ff00;`, the devtools shows me `rgb(0, 255, 0)`:

Screen Shot 2021-11-08 at 2 38 48 PM

...but I'm pretty sure I don't want to change the hex values to rgb().

So I think the question I'm interested in is: is this discrepancy happening in the interactive-examples machinery or in the devtools. If it's in the devtools, there's no bug here: interactive-examples doesn't have any control over how the devtools presents values to the user.

If you run the interactive examples locally (using npm run start) you can see the generated files that are used in the interactive examples iframe. The one for `color`, for example could be found at interactive-examples/docs/pages/css/color.html. If I look at this in a text editor I see the original values:

Screen Shot 2021-11-08 at 2 55 09 PM

Also, if I open https://interactive-examples.mdn.mozilla.net/pages/css/color.html and select "View source" I also see the original values:

Screen Shot 2021-11-08 at 2 56 48 PM

So all that makes me think this is an artefact of the devtools, and not a thing we could or should try to address in interactive-examples.

wbamberg:
OK, so I think the issue you're seeing is that the CSS value shown in the editor uses a different form from the value the devtools are telling you? I see this as well.

For example looking at https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-color, if I select the third choice `border-bottom-color: rgb(170, 50, 220, .6);`, then I see `rgba(170, 50, 220, 0.6)` in the Firefox inspector:

Screen Shot 2021-11-08 at 2 31 16 PM

Is this the discrepancy you are talking about? If so: that is weird.

But I don't think changing the source to match is the right thing to do. For one thing, there are other cases where this is also happening. For example in the page for https://developer.mozilla.org/en-US/docs/Web/CSS/color, if I select the second option `color: #00ff00;`, the devtools shows me `rgb(0, 255, 0)`:

Screen Shot 2021-11-08 at 2 38 48 PM

...but I'm pretty sure I don't want to change the hex values to rgb().

So I think the question I'm interested in is: is this discrepancy happening in the interactive-examples machinery or in the devtools. If it's in the devtools, there's no bug here: interactive-examples doesn't have any control over how the devtools presents values to the user.

wbamberg:
I'm also not very familiar with the interactive-examples machinery (although @schalkneethling is) but it looks as if the relevant bit might be https://github.com/mdn/bob/blob/b0b44fd5fa362ca717eac2318c2f5674cd6a7d04/editor/js/editor-libs/events.js#L18-L20 :

```js
    exampleChoiceList.addEventListener('keyup', function(event) {
        var exampleChoiceParent = event.target.parentElement;

        cssEditorUtils.applyCode(
            exampleChoiceParent.textContent,
            exampleChoiceParent
        );
    });
```

...on `keyup` we fetch the text content of the example choice (for example, `color: #00ff00;` in the second choice on https://developer.mozilla.org/en-US/docs/Web/CSS/color, and pass it to [`applyCode`](https://github.com/mdn/bob/blob/main/editor/js/editor-libs/css-editor-utils.js#L3-L26):

```js
    applyCode: function(code, choice, targetElement) {
        // http://regexr.com/3fvik
        var cssCommentsMatch = /(\/\*)[\s\S]+(\*\/)/g;
        var element =
            targetElement || document.getElementById('example-element');

        // strip out any CSS comments before applying the code
        code.replace(cssCommentsMatch, '');

        element.style.cssText = code;

        // clear any existing timer
        clearTimeout(this.editTimer);
        /* Start a new timer. This will ensure that the state is
        not marked as invalid, until the user has stopped typing
        for 500ms */
        this.editTimer = setTimeout(function() {
            if (!element.style.cssText) {
                choice.parentNode.classList.add('invalid');
            } else {
                choice.parentNode.classList.remove('invalid');
            }
        }, 500);
    },
```

...where we're setting the CSS string as `element.style.cssText = code;`.

So it seems that interactive-examples is just setting the exact content of the selected choice.

I tried replicating this in the devtools.

1. Visit http://example.org/
2. Open the console
3. Get the text paragraph
4. Set the paragraph's `style.cssText` to `color: 00ff00`
5. Get the paragraph's `style.cssText`

Here's what the console looks like:

Screen Shot 2021-11-08 at 5 03 14 PM

...that suggests to me that the conversion from one color format to another is happening in the browser/devtools, and isn't a thing we can do anything about.

schalkneethling:
Thank you for digging into this @wbamberg! Indeed, we do not CSS post processing at all with interactive examples so, the discrepancy is definitely caused by devtools.

@keikoro Do you see the same discrepancy in Chrome devtools?

keikoro:
Also, thanks for linking to those scripts, I guess it would be helpful to make note of the other repository somewhere in the README (which is a different issue/should get its own issue, just mentioning it).

wbamberg:
I'm closing this PR, as this doesn't seem to be an issue we can do anything about.

PR #1931, opened 2021-10-24
Throw in catch statement to show .finally dynamics

Tomassito:
Simply logging in the catch block doesn't show any difference between .finally  and .then  (both producing the same behaviour regardless what we use in line 18.)

### DIFF
--- a/live-examples/js-examples/promise/promise-finally.js
+++ b/live-examples/js-examples/promise/promise-finally.js
@@ -13,7 +13,7 @@ checkMail()
     console.log(mail);
   })
   .catch((err) => {
-    console.error(err);
+    throw new Error('Error from within catch block');
   })
   .finally(() => {
     console.log('Experiment completed');


schalkneethling:
Hey @Tomassito, and thank you for your contribution. The example demonstrates that `finally` is always run, irrespective of an error.

If there was no error, you see:

```bash
Mail has arrived
Experiment completed
```

if there was, you see:

```bash
Error: Failed to arrive
    at /home/runner/4oaf7xpwkrq/index.js:6:14
    at new Promise (<anonymous>)
    at checkMail (/home/runner/4oaf7xpwkrq/index.js:2:10)
    at Object.<anonymous> (/home/runner/4oaf7xpwkrq/index.js:11:1)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
Experiment completed
```

So in both cases, we see, `Experiment completed` output to the console.

wbamberg:
Thanks for your PR, @Tomassito ! I agree that the current example doesn't differentiate `finally` from `then` here. But I don't think that is necessary for the interactive example. The different between `finally` and `then` is a subtle and the page does explain it further down: I don't think the example needs to explain it. And I'm not sure the change here really does "explain it": I think most people will just be confused that `catch` is throwing again, and won't understand how that is relevant to the example.

So I'm closing this PR, but do appreciate your contribution.

PR #1946, opened 2021-11-12
Remove styling for anchor example

wbamberg:
In https://github.com/mdn/content/issues/10459 someone was confused that after removing all the content from the links the icons still showed up in the example. I agree that this is confusing and suggest we remove the `::before` style so we just show the default behavior.


teoli2003 (approved):
Makes sense.

PR #1949, opened 2021-11-16
Date.toLocaleDateString() - add note that result depends on timezone

hamishwillee:
Fixes https://github.com/mdn/content/issues/10483

As requested in https://github.com/mdn/content/pull/10525

### DIFF
--- a/live-examples/js-examples/date/date-tolocaledatestring.js
+++ b/live-examples/js-examples/date/date-tolocaledatestring.js
@@ -1,13 +1,11 @@
 const event = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));
 const options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
 
-//NOTE: Depending on your local timezone, your output may differ from "expected output"
-
 console.log(event.toLocaleDateString('de-DE', options));
-// expected output: Donnerstag, 20. Dezember 2012
+// expected output (varies according to local timezone): Donnerstag, 20. Dezember 2012
 
 console.log(event.toLocaleDateString('ar-EG', options));
-// expected output: الخميس، ٢٠ ديسمبر، ٢٠١٢
+// expected output (varies according to local timezone): الخميس، ٢٠ ديسمبر، ٢٠١٢
 
 console.log(event.toLocaleDateString(undefined, options));
-// expected output: Thursday, December 20, 2012 (varies according to default locale)
+// expected output (varies according to local timezone and default locale): Thursday, December 20, 2012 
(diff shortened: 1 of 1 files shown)


wbamberg (changes requested):
Sorry to be a pain, but if we're going to add a generic comment here, it looks weird to have a specific comment at line 12 (now 13). So I suggest we should remove the " (varies according to default locale)" bit from line 12 (now 13)

live-examples/js-examples/date/date-tolocaledatestring.js
```diff
@@ -1,7 +1,8 @@
 const event = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));
-
 const options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
 
+//NOTE: The "expected output" depends on timezone, and may not be correct on your browser
```
wbamberg, line 4: "not be correct" is a bit strong.

```suggestion
//NOTE: Depending on your local timezone, your output may differ from "expected output"
```

hamishwillee:
Thanks for review. I accepted your change.

> Sorry to be a pain, but if we're going to add a generic comment here, it looks weird to have a specific comment at line 12 (now 13). So I suggest we should remove the " (varies according to default locale)" bit from line 12 (now 13)

The specific comment is actually relevant and correct - ie the string format changes depending on locale settings, but only for the last case where you are using the default locale. We could do something like below, but I'm concerned people might not get the point that locale only affects the last case.

```
//NOTE: Depending on your local timezone and locale, your output may differ from "expected output"
```

And  now you see why I wanted to just update the doc content!

wbamberg:
Then could we update all three comments:

```js
// expected output: Donnerstag, 20. Dezember 2012 (varies according to local timezone)

// expected output:  الخميس، ٢٠ ديسمبر، ٢٠١٢ (varies according to local timezone)

// expected output: Thursday, December 20, 2012 (varies according to local timezone and default locale)
```

hamishwillee:
@wbamberg I did not think of that. I've done this, but with the comments like `// expected output (varies according to local timezone: Donnerstag, 20. Dezember 2012)`

This was mostly because in the middle case we switch from a left-to-right to a right-to-left language. Switching again seemed cruel. I also don't think it hurts to note the result might differ before showing it.

hamishwillee:
This appears to be failing due to a change I have not made.

wbamberg (approved):
Thanks for making this change, looks great to me.

wbamberg:
The warning is due to https://github.com/mdn/interactive-examples/pull/1782, because our version of ESLint doesn't support `hasIndices`. But it looks like the new version does! https://eslint.org/blog/2021/10/eslint-v8.0.0-released. so I guess we should upgrade.

hamishwillee:
Thanks for the review and the help. Good to get this done.

wbamberg:
Thank you for your patience with my nit-picking :)

wbamberg:
Oh and I see we have https://github.com/mdn/interactive-examples/pull/1944 for the ESLint upgrade, it looks scary though!

PR #1954, opened 2021-11-21
Change RegExp variable with lastIndex property for expected output

AnilSeervi:
Update RegExp variable for expected output.

- Fixes mdn/content#10678

### DIFF
--- a/live-examples/js-examples/regexp/regexp-prototype-test.js
+++ b/live-examples/js-examples/regexp/regexp-prototype-test.js
@@ -6,7 +6,7 @@ const globalRegex = new RegExp('foo*', 'g');
 console.log(regex.test(str));
 // expected output: true
 
-console.log(globalRegex.lastIndex);
+console.log(regex.lastIndex);
 // expected output: 0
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for this PR, @AnilSeervi , I appreciate it but I don't think this is right, and I don't think the original issue is either.

As I understand it `lastIndex` isn't very interesting unless `/g` is set. The purpose of referring to `lastIndex` here is to show how multiple calls to `test` advance through the string from the last match, when `/g` is set, eventually returning `false`. So the first call at line 9 is to establish the initial value of `lastIndex`, so we can see how it advances after the call to `test`.

AnilSeervi:
> The purpose of referring to `lastIndex` here is to show how multiple calls to `test` advance through the string from the last match, when `/g` is set, eventually returning `false`.

Sure. My first impression was also the same understanding. But the invocation of the property twice without any context can lead to confusions as the property returns same value at that line of execution.

Perhaps a comment can make it clearer.

Issue #1922, opened 2021-10-04
Remove eslint ignored files


Rumyra:
eslint does not support static initialization blocks in js classes ... yet. This meant two js examples were failing tests - they have both been added to the `.eslintignore` file. They are:

* live-examples/js-examples/classes/classes-static-initialization.js
* live-examples/js-examples/classes/classes-static.js

Once [this issue](https://github.com/eslint/eslint/issues/15016) has been resolved and eslint supports static blocks, we can remove these files from the ignore file

ddbeck:
The actual changes are:

- https://github.com/mdn/interactive-examples/commit/c35c3c653a645a9440819751690101dc772098a0
- https://github.com/mdn/interactive-examples/commit/3f6ed18a5d629f5e32d19f5a46f88efd55a7ebc8

PR #1959, opened 2021-11-25
Update ESLint config for version 8

wbamberg:
Now we have ESLint 8 (thanks @schalkneethling ) we should be able to unpick some of the tweaks we had to make:

* undo changes to stop ESLint complaining about static classes, now it supports them (fixes https://github.com/mdn/interactive-examples/issues/1922)
* fix actual lint errors in those examples
* undo https://github.com/mdn/interactive-examples/pull/1798 now ESLint supports `hasIndices`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,2 @@
 js/lib/*
 js/mode/*
-
-# js class static blocks is currently ignored as it fails test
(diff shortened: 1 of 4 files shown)


schalkneethling (approved):
thanks, @wbamberg

PR #1963, opened 2021-12-02
Update expressions-right-shift.js

JMarkMurphy:
Corrected binary for -5 and -2


schalkneethling:
@JMarkMurphy, thanks so much for your contribution. When doing the conversion using [RapidTables](https://www.rapidtables.com/convert/number/hex-to-binary.html), it seems the current representation is correct. I am not confident, though so @wbamberg can you have a look as well? - Thanks!

Rumyra (approved):

Rumyra:
@JMarkMurphy thank you!

@schalkneethling from using your link I do think this change is correct, so I'll merge 👍

PR #1971, opened 2021-12-08
Fix nesting depth in Array.prototype.flat example

bodograumann:
Nesting is reduced one level by array.flat. I have removed one level of nesting from the input to fix the semantics and at the same time make it more readable, while still illustrating the point.

### DIFF
--- a/live-examples/js-examples/array/array-flat.js
+++ b/live-examples/js-examples/array/array-flat.js
@@ -3,7 +3,12 @@ const arr1 = [0, 1, 2, [3, 4]];
 console.log(arr1.flat());
 // Expected output: Array [0, 1, 2, 3, 4]
 
-const arr2 = [0, 1, 2, [[3, 4]]];
+const arr2 = [0, 1, [2, [3, [4, 5]]]];
 
+// flatten with default depth 1
+console.log(arr2.flat());
+// expected output: Array [0, 1, 2, Array [3, Array [4, 5]]]
+
+// flatten with depth 2 (same as calling .flat() two times)
 console.log(arr2.flat(2));
-// Expected output: Array [0, 1, 2, Array [3, 4]]
+// expected output: Array [0, 1, 2, 3, Array [4, 5]]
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for the PR, @bodograumann , and I'm sorry to take so long to review this PR.

I'm not sure what "fix the semantics" means in this case, but I do think the current example is showing something that this version is not: namely that the argument is flattening to exactly the level given and not any further, so the array is not fully flattened. This seems like a useful thing to show in the example.

bodograumann:
After another look, it seems I overlooked the `depth` parameter.
I’ll need to think about this some more.

bodograumann:
What do you think about this approach, @wbamberg ?

wbamberg:
This misses the case where the argument is > the nesting level, but I don't think that is so important to show, and I don't want this demo to be longer.

bodograumann:
I like the idea a lot. Is the output represenation usual for MDN examples? In the firefox console I see:
Screenshot_20220610-161850

wbamberg:
> I like the idea a lot. Is the output represenation usual for MDN examples? In the firefox console I see: Screenshot_20220610-161850

For the interactive examples we override the Firefox console implementation so as to log results into the example output pane instead of the own console. The output isn't always formatted in the same way - there's no particular good reason for this, but the form we give in `//expected output` should match the form people see when they click the "Run" button.

(you'll also note differences between the Firefox console and the Chrome console - the way consoles format output is not standardised).

bodograumann:
How about using `Infinity` instead of `3` for the last variant?

bodograumann:
I’ve commited your proposal modulo the `Infinity` example, which I think is the second most-common use-case after not using any `depth` parameter.

bodograumann:
I have rebased on main.

bsmth (approved):
Looks good to me, leaving a +1

Any objections @wbamberg?

wbamberg (approved):

PR #1975, opened 2021-12-15
🔥 Retire firefox vendor prefix from example

AnilSeervi:
### Summary

Moves up the vendor prefix `-moz-tab-size` so that the main property gets applied if it is supported by the browser.


wbamberg (approved):
@AnilSeervi you are a ⭐.

AnilSeervi:
> @AnilSeervi you are a ⭐.

Thank you @wbamberg

Issue #1648, opened 2020-08-09
JS examples: need example for Intl.ListFormat.resolvedOptions()


wbamberg:
See https://wiki.developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/resolvedOptions.

The interactive example could follow a similar pattern to the example for [Intl.NumberFormat.resolvedOptions()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/resolvedOptions).

IestynGage:
I'll have a look at this 😄

IestynGage:
hey @wbamberg whats the next step? 😄

wbamberg:
Apparently the next step is "ping me because I forgot" :). Sorry! I'll look at it now.

NiedziolkaMichal:
Closed via #1986

PR #1986, opened 2022-01-05
 Add example for Intl.listFormat.prototype.resolvedOptions()

IestynGage:
First time contributing to this project 😄 

#1648

Edit: Oops included the wrong number 🤦‍♂️

### DIFF
--- a/live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
+++ b/live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
@@ -1,4 +1,4 @@
-const deListFormatter1 = new Intl.ListFormat('de-DE', { style: 'short', type: 'disjunction' });
+const deListFormatter1 = new Intl.ListFormat('de-DE', { type: 'disjunction' });
 const options1 = deListFormatter1.resolvedOptions();
 
 console.log(options1.locale);
@@ -6,7 +6,7 @@ console.log(options1.locale);
 // expected output (Chrome): "de"
 
 console.log(options1.style);
-// expected output: "short"
+// expected output: "long"
 
 console.log(options1.type);
 // expected output: "disjunction"
--- a/live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
+++ b/live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
@@ -1,0 +1,12 @@
+const deListFormatter1 = new Intl.ListFormat('de-DE', { style: 'short', type: 'disjunction' });
+const options1 = deListFormatter1.resolvedOptions();
+
+console.log(options1.locale);
+// expected output (Firefox / Safari): "de-DE"
+// expected output (Chrome): "de"
+
+console.log(options1.style);
+// expected output: "short"
+
+console.log(options1.type);
+// expected output: "disjunction"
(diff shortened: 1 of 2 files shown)


live-examples/js-examples/intl/meta.json
```diff
@@ -78,6 +78,12 @@
             "title": "JavaScript Demo: Intl.GetCanonicalLocales",
             "type": "js"
         },
+        "intlListFormatPrototypeResolvedOptions": {
+            "exampleCode": "./live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js",
+            "fileName": "intl-listformat-prototype-resolvedoptions.html",
+            "title": "JavaScript Demo: Intl.ListFormat.prototype.resolvedOptions()",
+            "type": "js"
+        },
```
wbamberg, line 86: this should go after "intlListFormat" (really we should have CI that checks these JSON files are alphabetised).

live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
```diff
@@ -0,0 +1,12 @@
+const deListFormatter1 = new Intl.ListFormat('de-DE', { style: 'short', type: 'disjunction' });
```
wbamberg, line 1:
```suggestion
const deListFormatter1 = new Intl.ListFormat('de-DE', { type: 'disjunction' });
```

maybe, so we can show the defaults too? (of course you'd need to change the expected output as well)

live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
```diff
@@ -0,0 +1,12 @@
...
+const options1 = deListFormatter1.resolvedOptions();
+
+console.log(options1.locale);
+// expected output (Firefox / Safari): "de-DE"
+// expected output (Chrome): "de"
```
wbamberg, line 6: I don't really like calling out specific browsers in here. Maybe we could avoid this issue by choosing `"fr"` instead? If so we should change the name of the variable of course.

wbamberg (changes requested):
Thanks! I had a few little comments.

live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
wbamberg: Oh I just saw that the other one uses `de`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/resolvedOptions. Well maybe keep it then, so we are consistent.

live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
IestynGage: Alright added :) plus changed the comment to long for the style console log.

live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js
IestynGage: I'll leave the language as it is then 👍

IestynGage:
I removed the spurious character that I accidently added when re-ordering `meta.json`, I'll also make a pull request to [mdn/content](https://github.com/mdn/content)

wbamberg:
> I removed the spurious character that I accidently added when re-ordering `meta.json`, I'll also make a pull request to [mdn/content](https://github.com/mdn/content), when this gets merged 😄

Sounds great! Please ask if you have any q's.

PR #1987, opened 2022-01-05
Initial set of live examples for Intl.Segmenter et al.

SphinxKnight:
As a companion for https://github.com/mdn/content/pull/8402

### DIFF
--- a/live-examples/js-examples/intl/intl-segmenter-prototype-resolvedoptions.js
+++ b/live-examples/js-examples/intl/intl-segmenter-prototype-resolvedoptions.js
@@ -1,0 +1,8 @@
+const segmenter1 = new Intl.Segmenter('fr-FR');
+const options1 = segmenter1.resolvedOptions();
+
+console.log(options1.locale);
+// expected output: "fr-FR"
+
+console.log(options1.granularity);
+// expected output: "grapheme"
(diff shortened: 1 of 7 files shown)


teoli2003 (approved):
👍

wbamberg (commented):
@SphinxKnight , these examples look great, but I think it's worth considering if it's possible to make some of them shorter.

We'd like to keep examples to 12 lines or less, then they fit in the "standard" editor and generally nicely in the page. We do go over that when we have to (e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/preventExtensions) but it's not ideal.

In this PR 4/6 examples are "big". Can we make any of them shorter without compromising clarity or illustrative power? For example, https://github.com/mdn/interactive-examples/pull/1987/files#diff-58e8adc19b4b1ddad231f391c5d74f8fb16a43bb29c18b58bb1ae73551f5cf07 uses 7 lines for the output comment. Perhaps we could instead do something like:

```js
const segmenterFr = new Intl.Segmenter('fr', { granularity: 'word' });
const string1 = 'Que ma joie demeure';

const iterator1 = segmenterFr.segment(string1)[Symbol.iterator]();

for (const segment of iterator1) {
  if (segment.segment.length > 4) {
    console.log(segment.segment);
  }
}
// expected output: "demeure"
```

? Or maybe something better.

SphinxKnight:
@wbamberg sorry for slacking here, I'll get back to this in the coming days :x

wbamberg (approved):
👍 thank you @SphinxKnight !

PR #1992, opened 2022-01-16
Add an example of slice without arguments

redrambles:
Adding a `slice()` example to demonstrate a full shallow copy of the array.


wbamberg (approved):
This is a useful addition, thank you @redrambles !

PR #1996, opened 2022-01-21
Improve Array.prototype.reduce intro example

jenweber:
Goal: `reduce` is often a confusing method for new learners, and we hope that this more explicit example will help build understanding.

Other array methods (`map`, `sort`, `filter`, `forEach`) all use inline arrow functions for their intro examples, but reduce does not. This PR makes the first example shown for `reduce`  follow that same convention. It also uses variable naming to communicate what has changed between the two examples.

The downside of this change is that the example is longer, but we think it's easier to understand now for beginners.

**Alternatives considered**

Examples would be shorter and more readable if we dropped `Value` from the parameter names (`currentValue` becomes `current`). However, that would not match the rest of the page's prose and examples, so we left the variable names as they were.

There are some other improvements to the page that we are brainstorming, and will open an issue for those.

Co-authored-by: [Kennie Boulin](Kennie-create)

### DIFF
--- a/live-examples/js-examples/array/array-reduce.js
+++ b/live-examples/js-examples/array/array-reduce.js
@@ -1,19 +1,19 @@
 const array1 = [1, 2, 3, 4];
 
 // 1 + 2 + 3 + 4
-const result = array1.reduce(
+const sum = array1.reduce(
   (previousValue, currentValue) => previousValue + currentValue
 );
 
-console.log(result);
+console.log(sum);
 // expected output: 10
 
 // 5 + 1 + 2 + 3 + 4
 const initialValue = 5;
-const resultWithInitial = array1.reduce(
+const sumWithInitial = array1.reduce(
   (previousValue, currentValue) => previousValue + currentValue,
   initialValue
 );
 
-console.log(resultWithInitial);
+console.log(sumWithInitial);
 // expected output: 15
--- a/live-examples/js-examples/array/array-reduce.js
+++ b/live-examples/js-examples/array/array-reduce.js
@@ -1,10 +1,17 @@
-const array1 = [1, 2, 3, 4];
-const reducer = (previousValue, currentValue) => previousValue + currentValue;
+const myArray = [1, 2, 3, 4];
 
 // 1 + 2 + 3 + 4
-console.log(array1.reduce(reducer));
+const result = myArray.reduce(
+  (previousValue, currentValue) => previousValue + currentValue
+);
+console.log(result);
 // expected output: 10
 
 // 5 + 1 + 2 + 3 + 4
-console.log(array1.reduce(reducer, 5));
+const initialValue = 5;
+const resultWithInitial = myArray.reduce(
+  (previousValue, currentValue) => previousValue + currentValue,
+  initialValue
+);
+console.log(resultWithInitial);
 // expected output: 15
--- a/live-examples/js-examples/array/array-reduce.js
+++ b/live-examples/js-examples/array/array-reduce.js
@@ -1,17 +1,19 @@
-const myArray = [1, 2, 3, 4];
+const array1 = [1, 2, 3, 4];
 
 // 1 + 2 + 3 + 4
-const result = myArray.reduce(
+const result = array1.reduce(
   (previousValue, currentValue) => previousValue + currentValue
 );
+
 console.log(result);
 // expected output: 10
 
 // 5 + 1 + 2 + 3 + 4
 const initialValue = 5;
-const resultWithInitial = myArray.reduce(
+const resultWithInitial = array1.reduce(
   (previousValue, currentValue) => previousValue + currentValue,
   initialValue
 );
+
 console.log(resultWithInitial);
 // expected output: 15
--- a/live-examples/js-examples/array/array-reduce.js
+++ b/live-examples/js-examples/array/array-reduce.js
@@ -1,6 +1,6 @@
 const array1 = [1, 2, 3, 4];
 
-// 5 + 1 + 2 + 3 + 4
+// 0 + 1 + 2 + 3 + 4
 const initialValue = 0;
 const sumWithInitial = array1.reduce(
   (previousValue, currentValue) => previousValue + currentValue,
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/array/array-reduce.js
```diff
@@ -1,10 +1,19 @@
...
 
 // 5 + 1 + 2 + 3 + 4
-console.log(array1.reduce(reducer, 5));
+const initialValue = 5;
+const resultWithInitial = array1.reduce(
+  (previousValue, currentValue) => previousValue + currentValue,
+  initialValue
+);
+
+console.log(resultWithInitial);
```
jneurock, line 18: These are nice changes. It helps better explain the initial value example quite a lot and the removal of the overly-generic (IMO) `reducer` function lets the reader focus more on what's important. I see opportunity for more clarity, too.

Given the preceding sentence for this code example:
> Perhaps the easiest-to-understand case for reduce() is to return the sum of all the elements in an array:

The reader already has the word "sum" in mind but the code only connects to that word via the `+` operator.

WDYT of changing `result` to `sum` and `resultWithInitial` to `sumWithInitialValue`?

I also think it's hard for new learners to make the connection between the callback arguments and the array itself. It _might_ also help to connect the dots between the `previousValue` and `currentValue` arguments to the actual array if the array were named `values`. WDYT?

Thanks for working to improve this!

jenweber: I love the sum variable naming idea. I’ll update this PR.

wbamberg:
I'm sorry to be slow to look at this @jenweber . I did see it but wasn't sure what to do (so did nothing, which isn't very helpful :). Honestly I'm conflicted about it. I do think your version is clearer than the original and I appreciate it being made consistent with the others.

The only doubt I have is that it makes the example longer. There are three fixed sizes of JS editor:

* [small](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf) (<7 lines)
* [medium](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) (7-12 lines)
* [large](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/preventExtensions) (>12 lines)

We aim to have the large majority of examples fit in the "medium" editor, so they don't take up so much vertical space, and generally fit better in the page. Here's the last time I did the size-distribution analysis, I don't expect it has changed much: https://github.com/mdn/interactive-examples/issues/1530#issuecomment-595508978 .

So when an example grows over 12 lines we always have to ask, is it worth it? For some APIs, like [`Proxy/preventExtensions`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/preventExtensions), it's really hard to see how to make a usable example fit into 12 lines, so we go over.

In this case we have an example in 12 lines that's _OK_, but the longer example is better. Maybe that's "better" is enough justification and I should obsess less about this. Maybe we should think a bit about whether there's a way to have it more readable but still fit into 12 lines. If we remove the second example with `initialValue`, it fits. Is it worth losing that extra example?

So given that extra context I'd be interested to hear what you think. If you still think this is optimal I'll go ahead and merge it, but if you had any other suggestions for how we could get the best of both worlds, I'd love to know.

wbamberg:
> Thanks for sharing your thoughts! What about if we remove the first example? And maybe we set the initial value of the accumulator to 0? Then we can keep the example shorter and still teach the accumulator/initial value handling.
> 
> My reasoning is that I think not specifying the accumulator is a hidden trap for new coders. Often times people use reduce to sum up things inside of objects, so the thing being iterated over is an object, but then they return a number from the method. Then all the object methods throw type errors in the next iteration of the loop. This is the problem I saw firsthand while teaching someone to use reduce, and what motivated the PR.
> 
> Later down in the docs, people can learn the default behavior of the accumulator.
> 
> I’ll post a couple different iterations later tonight to make this more concrete.

Thanks! I'd be happy with this, and also happy to accept your judgement either way: I just wanted to provide the context.

On general fixes to the`reduce()` page, that would be great! @hamishwillee [made some improvements here recently](https://github.com/mdn/content/pull/10555) but there's definitely unfinished business, and I would love to replace that big quote from the spec.

jenweber:
## Option 1, show one example

This is what I think the best option may be, especially given the past iterations on these docs.

```
const array1 = [1, 2, 3, 4]

// 0 + 1 + 2 + 3 + 4
const initialValue = 0;
const sum = array1.reduce(
  (previousValue, currentValue) => previousValue + currentValue,
  initialValue
);

console.log(sum);
// expected output: 10
```

## Option 2, show one example and shorten variable names

not much gained here

```
// 0 + 5 + 1 + 2 + 3 + 4
const initialValue = 0;
const sum = array1.reduce(
  (previous, current) => previous + current,
  initialValue
);

console.log(sum);
// expected output: 15
```

## The problem

Here's an example of a problem that happens if someone new to reduce doesn't specify a starting value, and then they have the wrong mental model:

```
const fruits = [ 
  { item: 'apples', quantity: 1 },
  { item: 'grapes', quantity: 2 },
  { item: 'kiwis', quantity: 3 },
  { item: 'bananas', quantity: 4 }
];

// Problem use
const sum = fruits.reduce((previous, current) => previous.quantity + current.quantity )
// Returns NaN

const corrected =  fruits.reduce((total, current) => total + current.quantity, 0)
```

## Option 3, show a more real-world example

This throws naming out the window and instead goes for a more concrete example. The downside is, someone needs to understand objects in order to understand this example, which I think of as a strong downside, and it's why there are more examples. I don't like this option, but it is an option.

```
const fruits = [ 
  { item: 'apples', quantity: 1 },
  { item: 'grapes', quantity: 2 },
  { item: 'kiwis', quantity: 3 },
];

const initial = 0
const sum =  fruits.reduce((total, current) => total + current.quantity, initial)

console.log(sum)
// expected output: 6
```

jenweber:
I updated this PR to show just one example, using the initialValue.

wbamberg (approved):
Thanks for your PR, @jenweber ! I agree that this looks like the best option.

jenweber:
Here's an attempt at fixing the Description section https://github.com/mdn/content/pull/12949

PR #2001, opened 2022-01-26
Remove use of Number.parseFloat in toPrecision example

sophiebits:
Number.parseFloat here serves only to distract from the use of toPrecision.


Elchi3 (approved):
Thank you! Looks good to me 👍

PR #2014, opened 2022-02-15
Array.groupBy() example

hamishwillee:
FF98 nightly adds support for Array.prototype.groupBy(). This is a basic example. Rest of docs that depend on this linked from https://github.com/mdn/content/issues/12579

PR #2015, opened 2022-02-15
Map.groupBy() interactive example

hamishwillee:
FF98 nightly adds support for `Array.prototype.groupByToMap()`. This is a basic example. Rest of docs that depend on this linked from https://github.com/mdn/content/issues/12579

This is the same as #2014 except that the grouping key can be any object, rather than something that can be coerced to a string.
This is a basic example showing the same grouping, but passing in a key and then using that key to get one of the groups. 

Note that as this uses a Map the comparison of whether two keys are equal are same-value-zero equality. From [this table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#a_model_for_understanding_equality_comparisons) you can see two objects that happen to have the same property names and values are not considered the same. 
That's why the _same_ key is needed to set and get the group.


wbamberg:
Running this locally, I get an error like this:

> MDN-BOB: No template found for template type wat

...and then no editor for the examples:

*****************

Screen Shot 2022-04-27 at 10 30 09 AM

*****************

Does anyone else see this? Does anyone know what the problem might be here? Perhaps some mismatch between interactive-examples and bob?

hamishwillee:
I have run this locally and not got that message.

The message "MDN-BOB: No template found for template type wat". indicates to me that BOB perhaps things this is a template of type "wat" which might be used by web assembly (it is the format for the human readable format).

bsmth:
@hamishwillee this is now https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy right?

Related pull requests:

- [x] https://github.com/mdn/content/pull/13171
- [x] https://github.com/mdn/content/pull/28424

hamishwillee:
Yes. I'll look at this next week!

hamishwillee:
@Josh-Cena This was the old interactive example I originally did for Array.prototype.groupByToMap(). I've updated it to the same basic example that is already given for `Map.groupBy()`. 

Is this OK to merge or do you want to close it because you have other ideas? Note that I appreciate that whether or not you "like" this, it is not yet time to add it to MDN.

Josh-Cena:
I think this is good, content-wise.

hamishwillee:
@bsmth This can be merged, but we might not put it into content for a while (Thanks Josh for comment).

bsmth:
> @bsmth This can be merged, but we might not put it into content for a while (Thanks Josh for comment).

Very good, merging shortly 👍🏻

bsmth (approved):

PR #2020, opened 2022-02-22
Fix Wasm trunc example

fearnoeval:
- While browsing the [Truncate docs on MDN](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Numeric/Truncate), I noticed that the example was not calling `f32.trunc`, but rather it was calling `f32.ceil`
- This pull request simply fixes the call


wbamberg (approved):
👍 thank you @fearnoeval , even I can see that this is better :)

PR #2023, opened 2022-02-24
Changes language marker from flags to text

teoli2003:
Using country flags for languages is an anti-pattern. We shouldn't do this in interactive examples.

This PR replaces the flags with some text describing the language.

Fixes mdn/content#13238

Note: this is my first PR in interactive-examples.


wbamberg (commented):
Thanks @teoli2003!

It looks like these .png files aren't used anywhere else, so could you delete them?

teoli2003:
Oh good catch. Thanks!

wbamberg (approved):
👍 thanks Jean-Yves!

PR #2029, opened 2022-03-03
Fixed text color in dark mode for CSS border

NiedziolkaMichal:
I have made changes to most of CSS border properties, to fix their visibility on new Dark Theme.

In border-color, border-image & border-width, text is now black.
In border-style properties, text color is now black and border-color is brown.
In border shorthands, text color and default border color is now darkmagenta, border-color of last example was changed from violet to lemon.

### DIFF
--- a/live-examples/css-examples/backgrounds-and-borders/border-bottom.html
+++ b/live-examples/css-examples/backgrounds-and-borders/border-bottom.html
@@ -29,7 +29,7 @@
     </div>
 
     <div class="example-choice">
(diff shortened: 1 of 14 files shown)


caugner (approved):

caugner:
LGTM, as all examples where color was added already had a background-color, but I will let @sideshowbarker have a look at it as well, just to make sure. 👍

sideshowbarker (approved):
Nice work

PR #2036, opened 2022-03-04
Fix for text-overflow in Dark Mode

NiedziolkaMichal:
text-overflow should now be visible in both dark and light mode. It was reported in #2031 and #2030


caugner (approved):

caugner:
Thank you @NiedziolkaMichal for your contributions! 👏

PR #2039, opened 2022-03-05
Fix Dark Mode in grid, flexbox & box-alignment properties

NiedziolkaMichal:
I changed all interactive examples in box-alignment, flexbox & grid directories, so they are visible well on both light and dark mode.
image
image

The only exception is order property, in which I removed 'width: 80%' and border, so all boxes are in interactive-examples container and border is not distracting.
image

### DIFF
--- a/live-examples/css-examples/box-alignment/align-content.css
+++ b/live-examples/css-examples/box-alignment/align-content.css
@@ -1,7 +1,5 @@
 #example-element {
-    background-color: #eee;
-    border: .75em solid;
-    padding: .75em;
+    border: 1px solid #c5c5c5;
     display: grid;
     grid-template-columns: 60px 60px;
     grid-auto-rows: 40px;
--- a/live-examples/css-examples/box-alignment/align-items.css
+++ b/live-examples/css-examples/box-alignment/align-items.css
@@ -1,7 +1,5 @@
 #example-element {
-    background-color: #eee;
-    border: .75em solid;
-    padding: .75em;
+    border: 1px solid #c5c5c5;
     display: grid;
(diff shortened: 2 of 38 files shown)


caugner (approved):
LGTM, but I will request another review to have a second opinion.

caugner:
@sideshowbarker Feel free to pass the review to someone else. 

(I'm still fairly new to the project and admittedly don't have the insight yet of who else could be a good `interactive-examples` reviewer for a second opinion here.)

sideshowbarker:
> (I'm still fairly new to the project and admittedly don't have the insight yet of who else could be a good `interactive-examples` reviewer for a second opinion here.)

I’m always happy to do reviews here, and I’ll take a look at this one today.

We don’t really have a dedicated set of reviewers for this repo — and definitely not a systematic reviewer-assignment mechanism set up for this repo in the way we do for the https://github.com/mdn/content/ repo. Maybe we should try to set up something — but I guess we haven’t need it so far, because we haven’t often needed to make changes until very recently, after the redesign. I guess things will settle down more once we get some bulk changes like this one made, to iron out the wrinkles that got exposed by the redesign.

NiedziolkaMichal:
Introducing dark mode, unfortunately, made many changes necessary. When the text color became white, it was no longer readable on the background that matches the light theme. It is a temporary problem though. There are about 50 more properties that require some changes, that I will try to solve in the coming days. After that, there won't be many changes required.

sideshowbarker (approved):

sideshowbarker:
More nice work — it’s great to see these fixes getting made

PR #2047, opened 2022-03-15
Removed vendor prefix from position: sticky

NiedziolkaMichal:
In [position](https://developer.mozilla.org/en-US/docs/Web/CSS/position) property, having -webkit-sticky seems unnecessary. It clutters example, while non-prefixed version seems to be working everywhere anyway.


teoli2003 (approved):
Yes. since Safari 13, prefixing is no more necessary.

PR #2055, opened 2022-03-21
Add stretch to justify-content

NiedziolkaMichal:
Simply adding stretch to [justify-content](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)


wbamberg:
For the [same reason as `position: fixed`](https://github.com/mdn/interactive-examples/pull/2068#issuecomment-1085219437), I don't think we should add this case.

PR #2059, opened 2022-03-24
using `2's complement` format for negative number

### LINKED ISSUE #2057: using `2's complement` format for negative number in `expressions-unsigned-right-shift`
In [this document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unsigned_right_shift), the `EmbedInteractiveExample` is using "binary with symbol" format, but the [content in description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unsigned_right_shift#description) is using "two's complement" format.

Should we update the `EmbedInteractiveExample` [here](https://github.com/mdn/interactive-examples/blob/29ea8cfd0d52b1c486f027a155a94ad901f0e066/live-examples/js-examples/expressions/expressions-unsigned-right-shift.js) to keep the same format with the main content.

## changes

```diff
const a = 5;          //  00000000000000000000000000000101
const b = 2;          //  00000000000000000000000000000010
- const c = -5;         // -00000000000000000000000000000101
+ const c = -5;         //  11111111111111111111111111111011

console.log(a >>> b); //  00000000000000000000000000000001
// expected output: 1

console.log(c >>> b); //  00111111111111111111111111111110
// expected output: 1073741822
```

Reference: mdn/translated-content#3228

yin1999:
Fixes #2057


wbamberg (approved):
👍 thank you, I agree it is good to be consistent with the page.

PR #2065, opened 2022-03-27
Rewrite void operator example

NiedziolkaMichal:
Fixes https://github.com/mdn/content/issues/14361.
As discussed in issue [#14361](https://github.com/mdn/content/issues/14361), exception output currently shows "Object {  }" which causes misunderstandings. Change to "e.toString()" will print "ReferenceError: test is not defined"


NiedziolkaMichal:
I have rewritten whole example. It now shows that:
- void always returns undefined
- its expression is normally evaluated
- it's possible to create IIFE with it
- declared function is not placed in surrounding scope

sideshowbarker (approved):

PR #2066, opened 2022-03-27
appearance

NiedziolkaMichal:
As requested in Issue [#11278](https://github.com/mdn/content/issues/11278), appearance examples now have visual effect. Current standard contains only 4 keywords: none, auto, menulist-button & textfield. I have changed example to button, on which none and auto have an effect(border changes). Unfortunately I don't know any element on which menulist-button and textfield have any visual impact, so I have removed them from examples.

### DIFF
--- a/live-examples/css-examples/basic-user-interface/appearance.html
+++ b/live-examples/css-examples/basic-user-interface/appearance.html
@@ -1,8 +1,6 @@
 <section id="example-choice-list" class="example-choice-list">
   <div class="example-choice">
     <pre><code class="language-css">
-    -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
Thanks @NiedziolkaMichal ! This is definitely better than what we have now, so I'll merge it. I do wonder though if we might not just remove this example. It's not very clear how helpful it is (or tbh how helpful this whole page, or the feature itself, is).

NiedziolkaMichal:
I think example is still helpful, even though this property really isn't. User might quickly see that this feature isn't as powerful as it once was.

Issue #1900, opened 2021-09-05
Content: position CSS does not have "fixed" example


shapkarin:
Hello. As you can see at the page https://developer.mozilla.org/en-US/docs/Web/CSS/position interactive example does not have "fixed" example. I created PR https://github.com/mdn/interactive-examples/pull/1896 to fix that.

### Current version
<img width="1022" alt="Screenshot 2021-09-05 at 07 23 14" src="https://user-images.githubusercontent.com/1463086/132115222-d60bd8ba-7668-4d6f-ae26-ec090d0cc767.png">

### Version from #1896 
<img width="1002" alt="Screenshot 2021-09-05 at 07 26 45" src="https://user-images.githubusercontent.com/1463086/132115261-f5b3ee5d-4739-4e70-8fa8-9ec1f1a10807.png">

I hope that PR #1896 done well and it fits nice to the developer.mozilla.org website page. Please recheck if it looks nice with the iframe.

wbamberg:
Thanks for filing this issue, @shapkarin , but it's intentional that `fixed` is omitted. We have a limited amount of space for CSS interactive examples, often not enough space for all the options. If we add `fixed`, then the set of examples will overflow and we will get a scrollbar, and I don't want that.

Interactive examples don't have to show all the syntax: we have the whole page to go into as much detail as we want.

seregadushka:
> we will get a scrollbar, and I don't want that.
Is that the only reason? Scrolling? Is that why it is necessary to cut off a significant part of knowledge? 
Then I suggest throwing out half of the CSS, because the font is the wrong color. 
Do you think that with position:fixed everything fixed already known? You're wrong. 
https://qna.habr.com/q/1363594 --15 comment screens of what you just cropped. 
The number of these comments is your fault.

Maykon153:
E mesmo 😏

Em qui., 1 de ago. de 2024, 19:53, seregadushka ***@***.***>
escreveu:

seregadushka:
> E mesmo :)

Are you normal? Decided to play Spanish Agatha Christie? And become my personal biographer?
I posted these links myself, I am writing about this topic myself. Or rather , about her absence from Mozilla. 
There are still not enough of your toothy smiles here.

Maykon153:
👏

Em qui., 1 de ago. de 2024, 20:46, seregadushka ***@***.***>
escreveu:

seregadushka:
https://github.com/hamishwillee
https://github.com/queengooborg

[...]

PR #2068, opened 2022-03-27
Added position: fixed

NiedziolkaMichal:
It's a fix for #1900

### DIFF
--- a/live-examples/css-examples/positioned-layout/position.html
+++ b/live-examples/css-examples/positioned-layout/position.html
@@ -22,6 +22,13 @@ top: 40px; left: 40px;</code></pre>
         </button>
     </div>
 
+    <div class="example-choice">
+        <pre><code class="language-css">position: fixed;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
     <div id="choice-sticky" class="example-choice">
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for filing this PR @NiedziolkaMichal , but I don't think we should do this. We have a limited amount of space for CSS interactive examples, often not enough space for all the options. If we add `fixed`, then the set of examples will overflow and we will get a scrollbar, and I don't want that.

Interactive examples don't have to show all the syntax: we have the whole page to go into as much detail as we want.

I should have said this before in the issue and I'm sorry for wasting your time. I'll go comment there now.

NiedziolkaMichal:
@wbamberg If content overflow and displayed scrollbar wouldn't be an issue, do you really think removing an example is the best choice in this case? I strongly believe showing a clear illustration of property values, can swiftly convey their purpose without going into details. It's quite easy to make an iframe of interactive examples that have adjusted height to their inner content, so why wouldn't we just show all the values and make the inner window have a dynamic size?

wbamberg:
We don't currently have the ability on MDN to make the iframe dynamically resize according to the content height.

Even if we did though, it is the intention of these examples to be short, and not necessarily to show all possible ways of using a feature. Interactive examples are somewhat rude, by occupying the top of the page. To make up for that we want to make them compact.

It's nice if we can show all the ways of using a feature in a small space, but there are always going to be features where that isn't practical. We do have the entire page to show the complete usage of the feature.

PR #710, opened 2018-04-13
Add <aside> example

wbamberg:

PR #2069, opened 2022-03-29
Fix setHours example for JS Date

wtlin1228:
In this page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours

Current `Try it` example's output doesn't match the expected output.

Screen Shot 2022-03-29 at 2 48 33 PM

After this fix

Screen Shot 2022-03-29 at 2 48 12 PM

### DIFF
--- a/live-examples/js-examples/date/date-sethours.js
+++ b/live-examples/js-examples/date/date-sethours.js
@@ -1,11 +1,11 @@
 const event = new Date('August 19, 1975 23:15:30');
 event.setHours(20);
 
-console.log(event);
+console.log(event.toString());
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
In case issue [#710](https://github.com/mdn/bob/issues/710) gets fixed, there will be a global solution and this PR will be unnecessary.

wbamberg:
> In case issue [#710](https://github.com/mdn/bob/issues/710) gets fixed, there will be a global solution and this PR will be unnecessary.

Yes, I have a fix for this at https://github.com/mdn/bob/pull/745.

wbamberg:
Thanks for the PR anyway, @wtlin1228 !

PR #2073, opened 2022-04-01
Remove -webkit-text-emphasis as it's now unprefixed

mrego:
See https://github.com/mdn/content/pull/14542

The example at https://github.com/mdn/interactive-examples/blob/d9e7dcae46e1dbbd0207f89628f8660328402978/CONTRIBUTING-CSS.md#supporting-prefixed-properties might need to be reworded or updated, as `text-emphasis` won't be longer a good example for that case.


wbamberg (approved):
👍 thank you @mrego for taking care of this update!

I'll update the bit of the meta-docs you pointed out.

PR #2074, opened 2022-04-01
docs: Change prefixed property example in meta-docs

wbamberg:
See https://github.com/mdn/interactive-examples/pull/2073 - `text-emphasis` is no longer prefixed, so it's not a good example. This PR switches it for `box-decoration-break`, which is still prefixed.


schalkneethling (approved):
Thanks, @wbamberg 🎉

Issue #706, opened 2018-04-12
HTML examples: output escapes from its pane


wbamberg:
CSS added to the CSS pane are applied to the entire page, not just the output:

![html-examples-escape](https://user-images.githubusercontent.com/432915/38700234-3c396ab2-3e4f-11e8-9db2-ae22e9faddd1.gif)

Similarly, clicking links loads the page into the entire example, not just the output:

![html-examples-escape-2](https://user-images.githubusercontent.com/432915/38700256-4f063b0c-3e4f-11e8-9dce-7a3adcf19eb4.gif)

a2sheppy:
Couldn't this be solved by just making it so the editing is in one iframe and the output is in a second iframe, and the editor periodically updates the HTML or CSS contained in the iframe? That shouldn't be hard to do, in theory, although I don't know details of the implementation already in place.

chrisdavidmills:
Or could you do it so that the CSS applied to the output is put through an intermediary stage where it has a namespacing selector put at the start of the chain in each case, like .output-pane ?

chrisdavidmills:
@a2sheppy yeah, I get it does get tricky at that point. I would love to find time to play with this.

schalkneethling:
@a2sheppy I was thinking another iframe as well at first but, that means a different doc to use as the source. We also need to be sure we do not run into scripting issues where we are prevented from updating the CSS and HTML. Currently updates to this happens almost instantly(200ms delay once you stop typing) as you type.

While it is not impossible to generate two separate docs, it will make authoring and testing these locally... hard ;)

@chrisdavidmills I also thought of preprocessing but, once someone chains a couple of selectors i.e. `.class .another .another, .someother { ... }` things can quickly become fragile. We also need to constantly be aware of over all byte size to avoid performance issues.

One solution that would work great is using shadow DOM for the output but, unfortunately only Chrome has support for it currently :(

schalkneethling:
The link portion of this is easy thankfully. I am simply going to attach an event listener to the output container and listen for any clicks on an anchor link that has a value that starts with at least `http`.

Once this happens, `preventDefault` and call `window.open(url);`

chrisdavidmills:
> One solution that would work great is using shadow DOM for the output but, unfortunately only Chrome has support for it currently :(

Could you use Polymer to polyfill it? I guess performance is then an issue.

schalkneethling:
[...]

PR #2089, opened 2022-04-16
feat(css): Add example of `@counter-style` and all of its descriptors

NiedziolkaMichal:
It adds interactive examples to @counter-style and all its descriptors except for [speak-as](https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style/speak-as), which doesn't have any visual impact.

image

image

image

image

image

image

image

image

image

image


wbamberg:
Thanks for this PR, @NiedziolkaMichal ! I don't have time to review now, but:

>     2. They don't work on firefox. It looks like latest version of firefox has an issue, with @counter-style and possibly other at-rules, be not operational, while being placed in shadow-root.

I remembered this after our conversation: https://github.com/mdn/interactive-examples/issues/887 . Maybe the same thing?

NiedziolkaMichal:
Yes, it looks to be the same issue, although @font-face doesn't work on both firefox & chrome, while @counter-style works fine on chromium based browsers. Is there any way to make it work?

wbamberg:
I don't know, @NiedziolkaMichal , sorry, @schalkneethling , do you have any ideas about this? I'm afraid that this is quite likely a blocker for this idea.

There is more information here:

* https://github.com/w3c/csswg-drafts/issues/5693
* https://bugzilla.mozilla.org/show_bug.cgi?id=1672302
* https://wiki.csswg.org/spec/css-scoping

schalkneethling:
@NiedziolkaMichal thank you for your continued contributions. I believe your recent pull request on mdn/bob will address the issues mentioned here? If so, let’s keep this one around until that one is merged. Thanks!

NiedziolkaMichal:
@schalkneethling No, this PR will still not work on Firefox, because output is presented in shadow tree. Recently I have been prototyping on replacing shadow DOM with iframe in HTML interactive examples. It would make it possible to use all at-rules and WebAPI examples, without much effort. Is there any important reason why you have chosen shadow dom instead?

schalkneethling:
> Is there any important reason why you have chosen shadow dom instead?

I honestly cannot remember :) I believe it could have been because we were, and still are, considering moving interactive examples into content without it being hosted in an `iframe` at all. We needed to prevent styling and Javascript "leaking" into the rest of the page.

NiedziolkaMichal:
@schalkneethling Making HTML examples be operational without either iframe or shadow DOM would be tricky. There are plenty of examples, like [p](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p), using selectors that would affect rest of the layout.

wbamberg:
There are I think 2 levels here:
- the complete example, including the UI, code editors, and output
- the output pane

At the moment the complete example is in an iframe, of course, and the output pane is in shadow DOM to stop it leaking into the editor UI. So even if the complete example was "inline" in the page, we'd need to isolate the output.

> Recently I have been prototyping on replacing shadow DOM with iframe in HTML interactive examples.

That's interesting! So we'd have an iframe inside an iframe? I might have mentioned this a few thousand times before, but I'd love to have Web API interactive examples on MDN.

wbamberg:
> @schalkneethling No, this PR will still not work on Firefox, because output is presented in shadow tree. Recently I have been prototyping on replacing shadow DOM with iframe in HTML interactive examples. It would make it possible to use all at-rules and WebAPI examples, without much effort. Is there any important reason why you have chosen shadow dom instead?

Hey, @NiedziolkaMichal , I have been trying to write this whole thing up and found this: https://github.com/mdn/interactive-examples/issues/706 which is where we decided to use shadow DOM. It was also suggested there that an extra iframe could be a solution.

Just in case this bit of history could help :).

NiedziolkaMichal:
@wbamberg In the [iframe PR](https://github.com/mdn/bob/pull/779), I believe that those issues don't exist. Applied CSS rules, don't select anything other than output. Clicking on links opens new browser tab. My solution also don't generate two separate documents, output exists only in memory and is applied via srcdoc attribute. On my device, I also didn't notice any performance issues.

live-examples/css-examples/@counter-style/counter-style.css
```diff
@@ -0,0 +1,11 @@
+@counter-style black-white-squares {
+    system: symbolic;
+    symbols: ◆ ◇;
+    suffix: ' ';
```
estelle:
```suggestion
    suffix: " ";
```

live-examples/css-examples/@counter-style/negative.css
```diff
@@ -0,0 +1,9 @@
+@counter-style negative-alphabet {
+    system: numeric;
+    symbols: 'A' 'B' 'C' 'D' 'E' 'F';
+    negative: '→' '←';
```
estelle:
```suggestion
    symbols: "A" "B" "C" "D" "E" "F";
    negative: "→" "←";
```

live-examples/css-examples/@counter-style/pad.css
```diff
@@ -0,0 +1,9 @@
+@counter-style binary-with-pad {
+    system: numeric;
+    symbols: '0' '1';
+    pad: 4 '_';
```
estelle:
```suggestion
    symbols: "0" "1";
    pad: 4 "_";
```

live-examples/css-examples/@counter-style/prefix.css
```diff
@@ -0,0 +1,9 @@
+@counter-style heads {
+    system: cyclic;
+    symbols: 👧 🧒 🧑 👱;
+    prefix: '🖐🏻';
```
estelle:
```suggestion
    symbols: 👧 🧒 🧑 👱;
    prefix: "🖐🏻";
```

live-examples/css-examples/@counter-style/range.css
```diff
@@ -0,0 +1,9 @@
+@counter-style missing-shapes {
+    system: cyclic;
+    symbols: '♥︎' '♤';
```
estelle:
```suggestion
    symbols: "♥︎" "♤";
```

live-examples/css-examples/@counter-style/suffix.css
```diff
@@ -0,0 +1,9 @@
+@counter-style octal {
+    system: numeric;
+    symbols: '0' '1' '2' '3' '4' '5' '6' '7';
+    suffix: ': ';
```
estelle:
```suggestion
    symbols: "0" "1" "2" "3" "4" "5" "6" "7";
    suffix: ": ";
```

live-examples/css-examples/@counter-style/symbols.css
```diff
@@ -0,0 +1,8 @@
+@counter-style cool-symbols {
+    system: symbolic;
+    symbols: '*' '\21C9'▲ 🐉 '🥲' ✹ ❎ 💓;
```
estelle:
```suggestion
    symbols: "*" "\21C9" ▲ 🐉 "🥲" ✹ '"' ❎ 💓;
```
i added a double quote here to show that quotes can be included

estelle (commented):
Looks good. My suggestions include adding white space, using double quotes, and adding an example of a using a quote as a symbol. 

I created a PR to add a comment about white space and quotes; so our example follow our best practices, but the content indicates that if you don't it should still work. 

https://github.com/mdn/content/pull/26677

bsmth:
> My suggestions include adding white space, using double quotes

The formatter doesn't seem to like this, so I've added in https://github.com/mdn/interactive-examples/pull/2089/commits/f4c78a49bca1949f36a2bd8a0dc58b0a2f7464b9 to make things green, I hope that's okay.

live-examples/css-examples/@counter-style/additive-symbols.css
```diff
@@ -0,0 +1,21 @@
...
+    400 cd,
+    100 c,
+    90 xc,
+    50 l,
+    40 xl,
+    10 x,
+    9 ix,
+    5 v,
+    4 iv,
+    1 i;
```
bsmth:
```suggestion
  additive-symbols: 50 l, 40 xl, 10 x, 9 ix, 5 v, 4 iv, 1 i;
```

bsmth (approved):
One suggestion open to prevent the line wrapping, but looks good aside from that, leaving a +1 👍🏻

bsmth:
Thanks everyone for the work! Merging shortly 🚢

PR #2093, opened 2022-04-19
Adds matrix & matrix3d functions

NiedziolkaMichal:
I added examples to [matrix](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix) and [matrix3d](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/matrix3d) functions. 

image

image

### DIFF
--- a/live-examples/css-examples/transforms/function-matrix3d.html
+++ b/live-examples/css-examples/transforms/function-matrix3d.html
@@ -11,10 +11,11 @@
 </div>
 
 <div class="example-choice" initial-choice="true">
-<pre><code class="language-css">transform: matrix3d(0.5, 0, -0.866025, 0,
-0.595877, 1.2, -1.03209, 0,
-0.866025, 0, 0.5, 0,
-25.9808, 0, 15, 1);</code></pre>
+<pre><code class="language-css">transform: matrix3d(
+  0.5,        0,      -0.866025, 0,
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/transforms/function-matrix3d.html
```diff
@@ -0,0 +1,27 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="transform">
+<div class="example-choice">
+<pre><code class="language-css">transform: matrix3d(-0.6, 1.34788, 0, 0,
+-2.34788, -0.6, 0, 0,
+0, 0, 1, 0,
+0, 0, 10, 1);</code></pre>
```
wbamberg, line 6: Do you think it would be worth aligning columns for readability? Like:

```suggestion
<pre><code class="language-css">transform: matrix3d(
  -0.6,       1.34788, 0,        0,
  -2.34788,  -0.6,     0,        0,
   0,         0,       1,        0,
   0,         0,      10,        1);</code></pre>
```

?

wbamberg (commented):
Thanks @NiedziolkaMichal ! I had a formatting suggestion, what do you think?

NiedziolkaMichal:
@wbamberg That's much better :)

wbamberg (approved):

PR #2096, opened 2022-04-21
Pseudo-elements after & before

NiedziolkaMichal:
Those are interactive examples for pseudo-elements [::before](https://developer.mozilla.org/en-US/docs/Web/CSS/::before) and [::after](https://developer.mozilla.org/en-US/docs/Web/CSS/::after).
I made new directory "pseudo-elements" for them, since there aren't any other examples for pseudo-elements or pseudo-classes.

image

image

@wbamberg Could you take a look please?


wbamberg (commented):
I think these are great!

I guess they are dependent on https://github.com/mdn/bob/pull/758/, so we can show the CSS tab by default? @schalkneethling , could you review that or find a reviewer for it?

I wonder if we could do something about naming and build location. At the moment these are getting built to, say, /pages/tabbed/before.html . It would be better if they ended up under /css with the other CSS examples, and maybe with a clear prefix to show that they are different? We use `function-thing` for CSS functions, so perhaps `pseudo-element-after`?

It's a relatively minor thing but would make it a bit easier for people working with these things.

But I love the idea and the content of the examples looks fine to me.

NiedziolkaMichal:
Current PR doesn't use property from [#758](https://github.com/mdn/bob/pull/758), so current commit is independent, but html tab is opened by default. I was thinking that if bob PR will be merged, then I will simply make another commit, changing those 2 pseudo elements, [@keyframes](https://github.com/mdn/interactive-examples/pull/2094) and other similar examples.

Currently, [bob](https://github.com/mdn/bob/blob/main/lib/pageBuilder.js) uses the "type" property to select the kind of interactive example to create, and then adds created page to a directory with the same name as that type.
image

I could create new type called "css-tabbed", so we can place in there all kinds of css properties/functions/classes that are shown in HTML interactive example window.

wbamberg:
> Current PR doesn't use property from [#758](https://github.com/mdn/bob/pull/758), so current commit is independent, but html tab is opened by default. I was thinking that if bob PR will be merged, then I will simply make another commit, changing those 2 pseudo elements, [@keyframes](https://github.com/mdn/interactive-examples/pull/2094) and other similar examples.
> 

Let's see if we can get that bob commit merged.

> Currently, [bob](https://github.com/mdn/bob/blob/main/lib/pageBuilder.js) uses the "type" property to select the kind of interactive example to create, and then adds created page to a directory with the same name as that type. image
> 
> I could create new type called "css-tabbed", so we can place in there all kinds of css properties/functions/classes that are shown in HTML interactive example window.

Actually maybe it's OK to keep them in "tabbed", if we just call the examples "pseudo-element-thing.html". That should be clear enough for maintainers.

schalkneethling:
> Let's see if we can get that bob commit merged.

I have asked engineering for review.

NiedziolkaMichal:
I have added defaultTab and changed file names to "pseudo-element-after.html" & "pseudo-element-before.html"

wbamberg:
Testing this, I get errors:

```
[2022-04-27T20:32:23.548Z]  "GET /css/editor-tabbed.css?v=1ab6a06" Error (404): "Not found"
[2022-04-27T20:32:23.548Z]  "GET /js/editor-tabbed.js?v=1ab6a06" Error (404): "Not found"
```

...and the tabbed editor UI isn't shown. This is for all tabbed examples, not just these ones. Perhaps this is because interactive-examples needs to know about the new bob version? @caugner , do you know the answer to this?

NiedziolkaMichal:
@wbamberg It seems like problem with compilation. I think you should checkout latest BOB and then run it by going to(with `cd`) Interactive Examples repository on your hard drive, then installing it by running this command:
```
npx install-local ~/path/to/repo/bob && node node_modules/.bin/mdn-bob
```
as it is said in [bob readme](https://github.com/mdn/bob#testing-bob-locally-inside-interactive-examples).

For me command above throws an exception(maybe because I work on windows), so I execute following steps when I test BOB:
1. cd C:/Repo/mdn/interactive-examples
2. npx install-local ../bob && node C:/Repo/mdn/interactive-examples/node_modules/mdn-bob/lib/mdn-bob.js
3. yarn start

schalkneethling:
@NiedziolkaMichal the most recent version of BoB was merged here https://github.com/mdn/interactive-examples/pull/2113
I believe if you merge main into this branch everything should work.

NiedziolkaMichal:
@schalkneethling Oh, alright. It's done now.

wbamberg (approved):
All right, these work fine for me now for some reason, and they look great. Thank you @NiedziolkaMichal , love them.

PR #2101, opened 2022-04-23
All inset properties

NiedziolkaMichal:
Those are interactive examples for [inset](https://developer.mozilla.org/en-US/docs/Web/CSS/inset), [inset-block](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block), [inset-block-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-end), [inset-block-start](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-start), [inset-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline), [inset-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end), [inset-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-start) properties.

Example text and style is copied from [top](https://developer.mozilla.org/en-US/docs/Web/CSS/top).

image

image

image

image

image

image

image

### DIFF
--- a/live-examples/css-examples/positioned-layout/inset-block-end.html
+++ b/live-examples/css-examples/positioned-layout/inset-block-end.html
@@ -1,0 +1,42 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="inset-block-end">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">inset-block-end: 20px;
+writing-mode: horizontal-tb;</code></pre>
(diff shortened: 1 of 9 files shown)


wbamberg (approved):

NiedziolkaMichal:
@wbamberg Thank you for review. Could you please also check [#2093](https://github.com/mdn/interactive-examples/pull/2093) and [#2066](https://github.com/mdn/interactive-examples/pull/2066)?

PR #2102, opened 2022-04-23
user-select

NiedziolkaMichal:
I have added interactive example for [user-select](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) property.

image

I have also changed 3 names in this meta.json, which used "-" as a word separator instead of using camelCase.


wbamberg (approved):
👍 thanks @NiedziolkaMichal !

PR #2103, opened 2022-04-23
Improve scroll-margin-inline

NiedziolkaMichal:
I added better examples to [scroll-margin-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-inline), so it is ready to be added to mdn/content.

image


wbamberg (approved):

PR #2104, opened 2022-04-23
aspect-ratio

NiedziolkaMichal:
Interactive example for [aspect-ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio).

image


NiedziolkaMichal:
I removed attr(width) / attr(height) example. It's not supported by chrome or firefox.

wbamberg (approved):
👍 that looks awesome. Thanks for the new example @NiedziolkaMichal !

PR #2111, opened 2022-04-27
defaultTab info

NiedziolkaMichal:
It adds info about usage of "defaultTab", added in [PR #758](https://github.com/mdn/bob/pull/758)

### DIFF
--- a/CONTRIBUTING-HTML.md
+++ b/CONTRIBUTING-HTML.md
@@ -308,6 +308,22 @@ Usage is as follows. When adding the meta information for your example, set the 
 }
 ```
 
+### Changing tab opened by default
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING-HTML.md
```diff
@@ -308,6 +308,22 @@ Usage is as follows. When adding the meta information for your example, set the
 }
 ```
 
+### Changing tab opened by default
+
+When HTML interactive example is loaded, HTML tab will be opened by default. We can change it to either `css` or `js`, by setting that value in `defaultTab` property. For example:
```
wbamberg, line 313:
```suggestion
When an HTML interactive example is loaded, the HTML tab will be opened by default. We can change the default open tab to either `css` or `js`, by setting that value in the `defaultTab` property. For example:
```

wbamberg (approved):
Just a little copy-editing, but thank for you updating the docs, @NiedziolkaMichal ! 👍

PR #2116, opened 2022-05-02
text-combine-upright

NiedziolkaMichal:
Interactive Example for property [text-combine-upright](https://developer.mozilla.org/en-US/docs/Web/CSS/text-combine-upright).
image

### DIFF
--- a/live-examples/css-examples/writing-modes/meta.json
+++ b/live-examples/css-examples/writing-modes/meta.json
@@ -7,13 +7,20 @@
             "title": "CSS Demo: direction",
             "type": "css"
         },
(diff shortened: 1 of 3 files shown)


wbamberg (commented):
This works but the effect is pretty subtle, I had to look a few times to see the difference. Do you think it's worth trying to make the effect a bit more obvious?

NiedziolkaMichal:
That was good suggestion. I have changed example and it is now shown like that:
image
image

wbamberg (approved):
Nice, thank you @NiedziolkaMichal !

PR #2117, opened 2022-05-02
Fix text-emphasis-position

NiedziolkaMichal:
Currently firefox requires both keywords( [ over | under ] && [ right | left ] ) to be specified in [text-emphasis-position](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position), so I added them to example and also added vertical writing mode examples so meaning of right/left is shown. Chrome recently added support for this property along with [text-emphasis-color](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-color) & [text-emphasis-style](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style), so we will be able to add all 3 of them to content.

image

### DIFF
--- a/live-examples/css-examples/text-decoration/text-emphasis-position.html
+++ b/live-examples/css-examples/text-decoration/text-emphasis-position.html
@@ -1,6 +1,6 @@
 <section id="example-choice-list" class="example-choice-list" data-property="text-emphasis-position">
     <div class="example-choice" initial-choice="true">
-        <pre><code class="language-css">text-emphasis-position: over left;</code></pre>
+        <pre><code class="language-css">text-emphasis-position: over right;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


wbamberg (commented):
Thanks @NiedziolkaMichal ! Very nice.

What do you think about changing the values so it's easier to see which of the keywords is making the difference in each writing mode? For example:

```
text-emphasis-position: over right;

text-emphasis-position: under right;

text-emphasis-position: over left;
writing-mode: vertical-rl;

text-emphasis-position: over right;
writing-mode: vertical-rl;
```

...is showing us that `left | right` is not used in horizontal mode, and `over | under` is not used in vertical mode.

NiedziolkaMichal:
That's a good idea, thank you. I have changed values, to be exactly the same as you have suggested.

wbamberg (approved):

PR #2118, opened 2022-05-02
repeating-conic-gradient

NiedziolkaMichal:
Interactive Example for [repeating-conic-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-conic-gradient).

image

### DIFF
--- a/live-examples/css-examples/images/function-repeating-conic-gradient.html
+++ b/live-examples/css-examples/images/function-repeating-conic-gradient.html
@@ -1,0 +1,36 @@
+<section id="example-choice-list" class="example-choice-list large">
+    <div class="example-choice">
+        <pre><code class="language-css">background: repeating-conic-gradient(red 0%, orange 15%, yellow 30%, green 40%, red 50%);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">background: repeating-conic-gradient(yellow 0% 25%, coral 0% 50%) 0% / 50px 50px;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">background: repeating-conic-gradient(from 3.1416rad at 10% 50%, #e66465, #9198e5 5% 10%);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">background: repeating-conic-gradient(red 15deg, orange 15deg 30deg, yellow 30deg 45deg,
+    green 45deg 60deg, blue 60deg 75deg, purple 75deg 90deg);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+</section>
+
+<div id="output" class="output large hidden">
+    <section id="default-example" class="display-block">
+        <div id="example-element"></div>
+    </section>
+</div>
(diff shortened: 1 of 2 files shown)


wbamberg (commented):
Thanks for this PR, @NiedziolkaMichal !

I think interactive examples work best when there's a very clear connection between the code and the result. That's really the value that they provide. So for me they succeed when, by looking at the example, I can quickly grasp something about the syntax of the feature. I think this is much more important than the effect being impressive. That is, they're syntax examples, not demos.

In this case, although the effects are impressive, it's hard to understand the syntax and how it's producing the effect. This is especially the case since the MDN redesign, where we get so little width to show the code, so in a case like this really all we're seeing is the name of the property, we don't see the syntax at all.

I think in cases where the syntax is complex, like this, it's best to do two things to try to make it more comprehensible:
- have simpler examples. We have a whole page where we can go into all the detail we like.
- have fewer examples, and split them across multiple lines. This helps with readability and also with the width problem.

For example, here we could have just two examples, and just cover the basics, like:

Screen Shot 2022-05-11 at 12 27 33 PM

To me, although these are simpler effects, they're teaching me more about how the syntax on the left produces the output.

I don't think it necessarily has to be exactly this, but something like this.

What do you think?

NiedziolkaMichal:
For sure examples written with proper indentation and line breaks, would improve readability. However change proposed here and in path function, contradicts style applied in pretty much every css example. Those are examples with very long property values: [background-image](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image), [background](https://developer.mozilla.org/en-US/docs/Web/CSS/background), [border-image-source](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-source), [border-image](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image), [conic-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient), [linear-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient), [repeating-linear-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-linear-gradient), [radial-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient), [repeating-radial-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-radial-gradient), [shape-image-threshold](https://developer.mozilla.org/en-US/docs/Web/CSS/shape-image-threshold), [basic-shape](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape). Line breaks usually occurs when another property is about to be set.

wbamberg:
> For sure examples written with proper indentation and line breaks, would improve readability. However change proposed here and in path function, contradicts style applied in pretty much every css example. Those are examples with very long property values: [background-image](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image), [background](https://developer.mozilla.org/en-US/docs/Web/CSS/background), [border-image-source](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image-source), [border-image](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image), [conic-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient), [linear-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient), [repeating-linear-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-linear-gradient), [radial-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/radial-gradient), [repeating-radial-gradient](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-radial-gradient), [shape-image-threshold](https://developer.mozilla.org/en-US/docs/Web/CSS/shape-image-threshold), [basic-shape](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape). Line breaks usually occurs when another property is about to be set.

Sorry, I know this is not consistent with our past practice. But I think in the cases you've listed here our past practice doesn't really serve our users well, and this has got worse since the redesign - examples that were marginally OK before are not any more.

NiedziolkaMichal:
Alright, how about those 2 examples?
image

image

wbamberg (approved):
These look great to me. Thank you for your understanding @NiedziolkaMichal !

PR #2119, opened 2022-05-03
Functions circle, ellipse, inset, polygon, path

NiedziolkaMichal:
Interactive examples for basic-shape functions: [circle](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/circle), [ellipse](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/ellipse), [inset](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/inset) and [polygon](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/polygon). Theme is the same as in [basic-shape](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape).

image
image
image
image

### DIFF
--- a/live-examples/css-examples/values-and-units/function-circle.html
+++ b/live-examples/css-examples/values-and-units/function-circle.html
@@ -1,0 +1,44 @@
+<section id="example-choice-list" class="example-choice-list large">
+
+    <div class="example-choice">
+        <pre><code class="language-css">clip-path: circle(50px);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">clip-path: circle(6rem at right center);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">clip-path: circle(10% at 2rem 90%);</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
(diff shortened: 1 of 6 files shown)


NiedziolkaMichal:
I have also added [path](https://developer.mozilla.org/en-US/docs/Web/CSS/path)
image

wbamberg (commented):
Thanks @NiedziolkaMichal !

I think `circle()`, `ellipse()`, and `inset()` are fine (although some of them overflow a bit) but as with `repeating-conic-gradient`, for especially `path()` and `polygon`, I worry that the syntax is not very visible or comprehensible. Maybe again it's worth having fewer simpler examples and using multiple lines to keep it readable, like

```
clip-path: path(
 'M  20 240  L  20  80
  L 160  80  L 160  20
  L 280 100  L 160 180
  L 160 120  L  60 120
  L  60 240 Z');

clip-path: path(
 'M 20 240
  C 20 0 300 0 300 240 Z');
```

?

NiedziolkaMichal:
@wbamberg Unfortunately I don't think it can be done so easily. Line break is incorrectly interpretted in middle of a string and the whole example is not applied:
image

It would require placing `\` before every line ending, just like that:
image

What do you think?

wbamberg:
Hm, nasty. Personally I think the version with the `\` line breaks is still better than the version on one line. I mean, the one-line version is really unusable, and at least like that we can see all the syntax.

wbamberg:
Thank you for the update @NiedziolkaMichal !

What do you think about doing this for `polygon()`:

```css
clip-path: polygon(10px 10px,
                  110px 10px,
                   60px 60px);

clip-path: polygon(50% 0%,
                  100% 50%,
                   50% 100%, 
                    0% 50%);
```

...I think this helps to show the syntax of `polygon()`, as a series of coordinates.

NiedziolkaMichal:
Sure. I have updated example and now it looks like this:
image

wbamberg (approved):
👍 I like the arrow version, much more interesting than mine. These are great @NiedziolkaMichal .

PR #2122, opened 2022-05-03
layer at-rule

NiedziolkaMichal:
Interactive Example for [@layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer). I have tested it on Firefox & Chrome, both of them work correctly.

I have added "at-rule-" prefix in file name. I guess we could include it in every at-rule, so there will be no confusion. What do you think @wbamberg ?

image


dipikabh (commented):
Hello @NiedziolkaMichal, like you mentioned in the PR, should the file names be updated to `at-rule-layer.css` and  `at-rule-layer.html`?

NiedziolkaMichal:
@dipikabh That's correct. Thank you

NiedziolkaMichal:
Thanks a lot

PR #2125, opened 2022-05-05
Fix font-synthesis example

NiedziolkaMichal:
Current version of font-synthesis usually doesn't give any visual effect, because used fonts had bold, italic and small-caps styles. I have attached fonts from fonts.google.com, that don't have those styles, so they will always be shown as intended. Also text "italic" was changed to "small-caps".
image
image

Unfortunately on Chrome, only small-caps value seems to be working, even though all of them are recognized. Example has visual effect, but bold and italic styles are always active:
image

### DIFF
--- a/live-examples/css-examples/fonts/font-synthesis.html
+++ b/live-examples/css-examples/fonts/font-synthesis.html
@@ -39,8 +39,8 @@
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
       <div id="example-element" class="transition-all">
-        <p>Most Western fonts include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">italic</span> variants but CJK fonts tend not to.</p>
-        <p>中文排版通常不运用<span class="bold">粗体</span>或<span class="italic">斜体</span>。</p>
+        <p class="english">Most Western fonts include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">small-caps</span> variants but CJK fonts tend not to.</p>
+        <p class="chinese">中文排版通常不运用<span class="bold">粗体</span>或<span class="italic">斜体</span>。</p>
       </div>
     </section>
 </div>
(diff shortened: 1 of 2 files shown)


live-examples/css-examples/fonts/font-synthesis.html
```diff
@@ -39,8 +39,8 @@
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
       <div id="example-element" class="transition-all">
-        <p>Most Western fonts include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">italic</span> variants but CJK fonts tend not to.</p>
-        <p>中文排版通常不运用<span class="bold">粗体</span>或<span class="italic">斜体</span>。</p>
+        <p class="english">Most Western fonts include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">small-caps</span> variants but CJK fonts tend not to.</p>
```
wbamberg, line 42:
```suggestion
        <p class="english">Most Western fonts include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">small-caps</span> variants but CJK fonts tend not to.</p>
```
```suggestion
        <p class="english">This font does not include <span class="bold">bold</span>, <span class="italic">italic</span>, and <span class="small-caps">small-caps</span> variants.</p>
```

I think the original idea of this example was that the English bit was the "Western font" that does include the variants, and the Chinese bit was the other font. So when the "Western font" actually doesn't include the variants, the wording feels contradictory.

wbamberg (changes requested):
Wording suggestion, otherwise but otherwise I like this change!

live-examples/css-examples/fonts/font-synthesis.html
NiedziolkaMichal: That seems better, thank you

wbamberg (approved):

PR #2127, opened 2022-05-05
Improve example of property all

NiedziolkaMichal:
Current example for property [all](https://developer.mozilla.org/en-US/docs/Web/CSS/all) doesn't show any difference between values inherit, unset, revert and revert-layer. I have improved example so difference between inherit, unset and revert is visible. Still revert and revert-layer will look the same, because rever-layer requires cascade layers in order to have any impact. Maybe it's better to remove it from this example and make example in layers in [revert-layer](https://developer.mozilla.org/en-US/docs/Web/CSS/revert-layer) page? Scrollbar would be gone too.

image

### DIFF
--- a/live-examples/css-examples/cascading-and-inheritance/all.css
+++ b/live-examples/css-examples/cascading-and-inheritance/all.css
@@ -1,8 +1,11 @@
 #example-element {
-    border: 2px dashed #999;
     color: red;
     padding: 10px;
     font-size: 1.5rem;
     text-align: left;
     width: 100%;
+}
+
+.example-container {
+    border: 2px dashed #999;
 }
--- a/live-examples/css-examples/cascading-and-inheritance/all.html
+++ b/live-examples/css-examples/cascading-and-inheritance/all.html
@@ -33,13 +33,6 @@
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
     </div>
-
-    <div class="example-choice">
-        <pre><code class="language-css">all: revert-layer;</code></pre>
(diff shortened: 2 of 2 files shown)


wbamberg (commented):
Yes, I think we should remove `revert-layer`. Otherwise, this is definitely better than what we had.

It's not going to be obvious to people that we have a paragraph inside a div with a border., which is kind of key to understanding what's going on. I'm tempted to cheat, and make the paragraph text contain this information:

> This paragraph has a font size of 1.5rem and a color of red. It's inside a `<div>`, and the `<div>` has a dashed gray border. `font-size` and `color` inherit, but `border` does not.

Just a suggestion.

NiedziolkaMichal:
How about this:

> This paragraph has a font size of 1.5rem and a color of red. It also has 1rem of vertical margin set by user-agent. Parent of the paragraph is `<div>` with a dashed gray border. 

It's important to point out vertical margin, because it is shown in `revert`, but not in `unset`. Also `font-size` and `color` don't inherit in this case, so last sentence is not needed.

wbamberg:
Copy edits:

> This paragraph has a font size of 1.5rem and a color of red. It also has 1rem of vertical margin set by the user-agent. The parent of the paragraph is a <div> with a dashed gray border.

...but otherwise, yes!

NiedziolkaMichal:
image

Looks good, thank you for suggestions.

wbamberg (approved):
👍 looks great, thank you, @NiedziolkaMichal !

PR #2129, opened 2022-05-05
font-variant-east-asian

NiedziolkaMichal:
Interactive Example for [font-variant-east-asian](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-east-asian).
image

Text comes from [page](https://www.wdic.org/w/WDIC/JIS%20C%206226-1978) about jp78 form.

### DIFF
--- a/live-examples/css-examples/fonts/font-variant-east-asian.css
+++ b/live-examples/css-examples/fonts/font-variant-east-asian.css
@@ -1,0 +1,5 @@
+#output section {
+    font-family: 'Yu Gothic', sans-serif;
+    margin-top: 10px;
+    font-size: 1.5em;
+}
--- a/live-examples/css-examples/fonts/font-variant-east-asian.html
+++ b/live-examples/css-examples/fonts/font-variant-east-asian.html
@@ -1,0 +1,37 @@
+<section id="example-choice-list" class="example-choice-list" data-property="font-variant-east-asian">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">font-variant-east-asian: normal;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">font-variant-east-asian: ruby;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">font-variant-east-asian: traditional;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">font-variant-east-asian: proportional-width;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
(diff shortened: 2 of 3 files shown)


mfuji09:
The text is taken from a proprietary (normal copyrighted) page which isn't an open source nor creative commons. Is it okay to add in MDN?

NiedziolkaMichal:
@mfuji09 I don't really know. Should I try to change it to something else?

mfuji09:
@NiedziolkaMichal 
Shall I create a suitable example sentence to test this property?

NiedziolkaMichal:
That would be great. I don't speak any asian language, so it's not easy to form something. Thank you

mfuji09:
Here is my example, and it can be provided as public domain.
JIS78とJIS83以降では、檜と桧、籠と篭など、一部の文字の入れ替えが行われている。また、「唖然」や「躯体」などの書体が変更されている。

And can you use `jis78` instead of `traditional`? The `traditional` value is useless in Japanese because it produces incorrect usage of traditional kanji in some cases.
For example, if you use `font-variant-east-asian: traditional`  for a text 弁護士 (lawer), it produces 辨護士, but the correct traditional kanji is 辯護士.
(It works fine in traditional Chinese.)

NiedziolkaMichal:
I have changed example to what you suggested. Everything seems to be great now. Thank you

mfuji09 (approved):
Thank you!

wbamberg:
Thanks for your PR @NiedziolkaMichal and your help @mfuji09 !

When running this example and switching between I don't see a difference, should I?

NiedziolkaMichal:
Yes, difference in each displayed value is significant. Could you please check which font is being used, by using firefox font tab? For me it is Yu Gothic Regular.

wbamberg:
Yes, for me it's PingFang SC. Perhaps because I'm using a Mac? Yu Gothic looks like it might be a Windows font: https://docs.microsoft.com/en-us/typography/font-list/yu-gothic#licensing-and-redistribution-info .

NiedziolkaMichal:
I have checked all preinstalled windows fonts and only Yu Gothic is affected by font-variant-east-asian property. Unfortunately it cannot be legally placed on MDN. I have also checked all japanese fonts on https://fonts.google.com/ by changing values of this property in dev tools, but none of the fonts shown me any effect. On [fonts.com](https://www.fonts.com/search/all-fonts?SortColumn=relevancy&searchtext=&ShowAllFonts=All&SearchType=AllFonts&page=1&FontFeature=Ruby%20Notation%20Forms) I see that there are some fonts with Ruby Notation Forms and property works fine for them, but none of them are free :/

wbamberg (approved):
👍 thanks so much for researching this, @NiedziolkaMichal . Then I think the only thing to do is merge this PR, but when we add it to the page add a note saying that for the example to show a difference between the options, the system must have a particular font installed (and perhaps even referencing Yu Gothic, as the page already does, further down).

mfuji09:
Thank you for your researching, @NiedziolkaMichal!
I will add some information I know.
"Yu Gothic" is also available in macOS and used in many web pages.
Its name might be "YuGothic" (without space) in macOS. Some blogs say so.
I have sometimes heard, however, it can be vanished in some cases, even if the main language was set in Japanese locale.

So, it may be better to use `font-family: "Yu Gothic Medium”, YuGothic, sans-serif;`.
(I have added "Medium" because I heard "Yu Gothic Medium" is closest to "YuGothic" of macOS.)

NiedziolkaMichal:
@mfuji09 I have found [list of fonts on macOS](https://support.apple.com/en-us/HT212587), `YuGothic Medium` and `YuGothic Bold` are there. I will make another PR with font-family set, maybe it will fix the problem.

PR #2130, opened 2022-05-06
Few tree-structural pseudo-classes

NiedziolkaMichal:
[:first-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child):
image

[:first-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type):
image

[:last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child):
image

[:last-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type):
image

[:nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child):
image

[:nth-last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-child):
image

[:nth-last-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-last-of-type):
image

[:nth-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type):
image

### DIFF
--- a/live-examples/css-examples/pseudo-class/first-child.css
+++ b/live-examples/css-examples/pseudo-class/first-child.css
@@ -3,5 +3,5 @@ p {
 }
 
 li:first-child {
-    background-color: gold;
+    border: 2px solid orange;
 }
--- a/live-examples/css-examples/pseudo-class/first-child.css
+++ b/live-examples/css-examples/pseudo-class/first-child.css
@@ -1,0 +1,7 @@
+p {
+    font-weight: bold;
+}
+
+li:first-child {
+    background-color: gold;
+}
--- a/live-examples/css-examples/pseudo-class/first-child.html
+++ b/live-examples/css-examples/pseudo-class/first-child.html
@@ -1,0 +1,12 @@
+<p>NBA players with most championships:</p>
+<ul>
+    <li>Bill Russell</li>
+    <li>Sam Jones</li>
+    <li>Tom Heinsohn</li>
+    <li>K. C. Jones</li>
+    <li>Satch Sanders</li>
+    <li>John Havlicek</li>
+    <li>Jim Loscutoff</li>
+    <li>Frank Ramsey</li>
+    <li>Robert Horry</li>
+</ul>
--- a/live-examples/css-examples/pseudo-class/first-of-type.css
+++ b/live-examples/css-examples/pseudo-class/first-of-type.css
@@ -1,0 +1,13 @@
+dt {
+    font-weight: 700;
+}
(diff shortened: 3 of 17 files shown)


OnkarRuikar:
Small suggestions:
- It would be great if we use same highlighting for the selected element in all examples. First I saw the post I immediately started to look for patterns like background for `*-child` and border for `*-type` etc. But for last-child you used text decoration. I like the golden border. In case of multiple selectors we can use same group of highlights.
- The rules `*last-child` say select the last child but output selects item with index '1'. Technically it's the last element but logically it's the first item. This may create confusion among beginners. Also, in this case only the indexes are inverted actual elements are not. So, the selector will always select `Peshtigo Fire` regardless of `reversed` property.

NiedziolkaMichal:
@OnkarRuikar Those are great suggestions, thank you. I have applied them and updated pictures in the first comment. Could you check if everything seems ok now?

wbamberg:
These look great! But I still see errors when I try to run it locally:

```
[2022-05-12T02:30:23.801Z]  "GET /css/editor-tabbed.css?v=f41847f" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:102.0) Gecko/20100101 Firefox/102.0"
[2022-05-12T02:30:23.804Z]  "GET /js/editor-tabbed.js?v=f41847f" Error (404): "Not found"
[2022-05-12T02:30:23.808Z]  "GET /css/editor-tabbed.css?v=f41847f" Error (404): "Not found"
[2022-05-12T02:30:23.890Z]  "GET /js/editor-tabbed.js?v=f41847f" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:102.0) Gecko/20100101 Firefox/102.0"
[2022-05-12T02:30:23.890Z]  "GET /js/editor-tabbed.js?v=f41847f" Error (404): "Not found"
MDN-BOB: Bundles built successfully
```

...and the examples look like this:

Screen Shot 2022-05-11 at 7 31 38 PM

I guess this is to do with the bob dependency, like https://github.com/mdn/interactive-examples/pull/2096 (where I also still see this error).

I expect I could fix it using one of the methods in https://github.com/mdn/interactive-examples/pull/2096#issuecomment-1111473510, but it bothers me that the version I'm trying to test doesn't just work. Does it work for you?

wbamberg:
Oh no hang on, it works now :).

wbamberg (approved):
👍 these are awesome.

PR #2168, opened 2022-05-29
Fix syntax error (put imports first)

pufferfish101007:
Fixes an error when running the live example at https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Variables/Global
```
> Error: parseWat failed:
8:20: error: imports must occur before all non-import definitions
  (global $from_js (import "env" "from_js") i32)
                   ^
```
This moves the imported global to above the created global.


MendyBerger:
LGTM

Thanks @pufferfish101007!

PR #2172, opened 2022-06-08
Added wasm conversion instructions

MendyBerger:
@sunfishcode please review whenever you get a chance, thanks!

Related to https://github.com/mdn/content/pull/17103


live-examples/wat-examples/numeric/extend.wat
```diff
@@ -0,0 +1,13 @@
+(module
+  (import "console" "log" (func $log (param i64)))
+  (func $main
+
+    i32.const 10 ;; push an i32 onto the stack
+
+    i64.extend_i32_u ;; extend from i32 to i64
```
sunfishcode, line 7: I suggest saying "zero-extend from i32 to i64", to emphasize how it's different from the `_s` form.

live-examples/wat-examples/numeric/trunc_float_to_int.wat
```diff
@@ -0,0 +1,13 @@
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+
+    f32.const 10.5 ;; push an i32 onto the stack
+
+    i32.trunc_f32_s ;; convert from signed f32 to i32 (.5 will be lost)
```
sunfishcode, line 7: The `signed` should be on the `i32` rather than on the `f32`. And I suggest adding "rounding towards zero" to be explicit how the conversion works.

live-examples/wat-examples/numeric/wrap.wat
```diff
@@ -0,0 +1,12 @@
+(module
+  (import "console" "log" (func $log (param i32)))
+  (func $main
+
+    i64.const 10 ;; push an i64 onto the stack
+
+    i32.wrap_i64 ;; wrap from i64 to i32
```
sunfishcode, line 7: "Wrap" may not be a familiar term to everyone, so I suggest giving a brief description of what this does. One can think of it either as reducing the value [mod](https://en.wikipedia.org/wiki/Modular_arithmetic) 2<sup>32</sup>, or as discarding the high 32 bits to produce a value containing just the low 32 bits.

live-examples/wat-examples/numeric/extend.wat
MendyBerger: Accepted your suggestion but changed to signed as I suspect that more people will understand `sign-extend` than `zero-extend`

live-examples/wat-examples/numeric/trunc_float_to_int.wat
MendyBerger: Accepted

live-examples/wat-examples/numeric/wrap.wat
MendyBerger: Added the explanation to the wrap page in https://github.com/mdn/content/pull/17103 as this pr is just for the example.

I've used almost your exact words as I think it summarises what wrap does really well

PR #2174, opened 2022-06-09
Few pseudo elements

NiedziolkaMichal:
Added interactive examples to pseudo elements [::file-selector-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button), [::first-letter](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter), [::first-line](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-line), [::placeholder](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) and [::selection](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection).

image

image

image

image

image

### DIFF
--- a/live-examples/css-examples/pseudo-element/file-selector-button.css
+++ b/live-examples/css-examples/pseudo-element/file-selector-button.css
@@ -1,0 +1,11 @@
+input {
+    margin-top: 1rem;
+}
+
+input::file-selector-button {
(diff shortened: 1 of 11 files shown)


wbamberg (approved):
👍 these look great @NiedziolkaMichal .

PR #2175, opened 2022-06-11
feat(css): Add example of `::backdrop`

NiedziolkaMichal:
This PR adds an interactive example for a [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop).

image

image


wbamberg:
Does `::backdrop` depend on the shadow DOM fix? The button doesn't work when I push it, and I get an error on the console:

> Uncaught TypeError: can't access property "addEventListener", showDialogBtn is null

NiedziolkaMichal:
Yes :/ I was testing them on BOB with iframe and didn't realize that it will fail on a current build.

I think all examples with JS tab will fail, so could you skip those PRs for now? I will try to move pseudo classes without any dependency, like a :marker, to a separate PRs.

estelle (approved):
this is working for me. Thanks!

estelle:
@wbamberg is this ok to merge?

wbamberg:
I haven't tested it lately but it ought to work now @NiedziolkaMichal 's fix to avoid shadow DOM is in, so if it works for you I'm sure it will be fine.

pcjmfranken:
The `::marker` example first removed by 294cd88 and later restored by f137050 has sadly fallen victim to squash merging: https://developer.mozilla.org/en-US/docs/Web/CSS/::marker (RIP in pieces).

This method can seriously mess things up in several annoying and unobvious ways. In this particular case it has allowed for an older change to overwrite a more recent conflicting one because their parent branches/PRs were merged in the opposite order.

Re-merging PR #2200 would probably produce the cleanest and best traceable fix for this. It might also be a good idea to verify nothing else was unintentionally deleted or overwritten!

PR #2179, opened 2022-06-12
Focus, hover, active

NiedziolkaMichal:
Those are interactive examples for pseudo-classes [:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus), [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible), [:focus-within](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within), [:hover](https://developer.mozilla.org/en-US/docs/Web/CSS/:hover), [:active](https://developer.mozilla.org/en-US/docs/Web/CSS/:active).

image

image

image

image

image


live-examples/css-examples/pseudo-class/hover.css
```diff
@@ -0,0 +1,14 @@
...
+    border: 2px solid burlywood;
+    border-radius: 10px;
+    font-weight: bold;
+    color: chocolate;
+    cursor: pointer;
+}
+
+.joinBtn:hover {
+    background-color: #ffd6a5;
+}
```
wbamberg: Maybe consider better contrast for the button text and background?

wbamberg (changes requested):
These work great and are fine examples, but asking for gender should not offer only Male/Female. See for example: https://www.ruth-ng.co.uk/how-to-ask-about-gender-in-forms-respectfully . 

Maybe consider something silly instead, like favourite flavor of ice cream?

NiedziolkaMichal:
@wbamberg I have changed Gender to Flavor as requested, also contrast of :hover was improved.

Thank you for reviewing.

wbamberg (approved):

PR #2180, opened 2022-06-13
MacOS fix in font-variant-east-asian

NiedziolkaMichal:
This commit was proposed by @mfuji09 in #2129 to fix [font-variant-east-asian](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-east-asian) example on MacOS. @wbamberg Could you please check if you see any visual difference between property values now?

On windows it's like that:
image

image

image

image


wbamberg (approved):
👍 thanks @NiedziolkaMichal and @mfuji09 for your work on this! I tried it and while itdidn't work out of the box, I could install "YuGothic" and then it worked fine. I still think this should have a note in the page but this is definitely an improvement.

PR #2184, opened 2022-06-14
only-child, only-of-type

NiedziolkaMichal:
Those are interactive examples for pseudo-classes [:only-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-child) & [:only-of-type](https://developer.mozilla.org/en-US/docs/Web/CSS/:only-of-type)

image

image


wbamberg (approved):
👍 thank you @NiedziolkaMichal , looks good to me!

PR #2189, opened 2022-06-17
valid, invalid, read-only, read-write, placeholder-shown

NiedziolkaMichal:
Those are interactive examples for pseudo-classes [:valid](https://developer.mozilla.org/en-US/docs/Web/CSS/:valid), [:invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid), [:read-only](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only), [:read-write](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-write), [:placeholder-shown](https://developer.mozilla.org/en-US/docs/Web/CSS/:placeholder-shown).

image

image

image

image

image


live-examples/css-examples/pseudo-class/read-write.html
```diff
@@ -0,0 +1,14 @@
+<p>Please fill your details:</p>
```
wbamberg:
```suggestion
<p>Please fill in your details:</p>
```

wbamberg (approved):

PR #2191, opened 2022-06-18
Autofill, checked, default

NiedziolkaMichal:
Those are interactive examples for pseudo-classes [:autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill), [:checked](https://developer.mozilla.org/en-US/docs/Web/CSS/:checked), [:default](https://developer.mozilla.org/en-US/docs/Web/CSS/:default).

image

image

image


NiedziolkaMichal:
I removed :indeterminate, because it depends on Shadow DOM PR.

wbamberg (approved):
👍 thank you @NiedziolkaMichal , looks great!

PR #2192, opened 2022-06-20
empty, lang

NiedziolkaMichal:
Those are interactive examples for pseudo-classes [:empty](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty), [:lang()](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang).

image

image


NiedziolkaMichal:
I removed :defined because it depends on Shadow DOM PR.

live-examples/css-examples/pseudo-class/empty.html
```diff
@@ -0,0 +1,8 @@
+<p>Element with no content:</p>
+<div></div>
+<p>Element with comment:</p>
+<div><!-- Simple Comment --></div>
+<p>Element with whitespace:</p>
+<div>    </div>
+<p>Element with nested empty element:</p>
+<div><p></p></div>
```
wbamberg:
```suggestion
<p>Element with no content:</p>
<div></div>

<p>Element with comment:</p>
<div><!-- Simple Comment --></div>

<p>Element with whitespace:</p>
<div>    </div>

<p>Element with nested empty element:</p>
<div><p></p></div>
```

wbamberg (approved):
👍 thank you @NiedziolkaMichal ! I pushed a whitespace change to make the HTML more readable.

PR #2196, opened 2022-06-24
feat(css): Add example of `:scope` and `:fullscreen`

NiedziolkaMichal:
Interactive Examples for [:scope](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope) and [:fullscreen](https://developer.mozilla.org/en-US/docs/Web/CSS/:fullscreen).

image

image

image


live-examples/css-examples/pseudo-class/scope.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <p id="special-paragraph">The love of my life and the most beautiful woman in the whole world is <s>Alice</s> Ava.</p>
+
+    <p>From the moment I saw her in a <s>coffeehouse</s> bar, I knew she has to be mine.</p>
```
estelle:
```suggestion
        <p id="special-paragraph">The love of my life and the most intriguing woman in the whole world is <s>Alice</s> Ava.</p>

    <p>From the moment I saw her in a <s>coffeehouse</s> bar, I knew she completed me.</p>
```
original quote made me a bit uncomfortable (women aren't owned. looks fade.)

estelle (approved):
Thanks! 🎉

PR #2197, opened 2022-06-27
Remove hgroup descendant selector from CSS

joshcrain:
The descendant selector `hgroup > h1` does not exist in the example HTML so styling is not applied to the h1 element. To apply the margin to the h1 we need only select it. Also, hgroup should not be used because no assistive technology supports it.

### DIFF
--- a/live-examples/html-examples/content-sectioning/css/section.css
+++ b/live-examples/html-examples/content-sectioning/css/section.css
@@ -1,3 +1,3 @@
-hgroup > h1, h2 {
+h1, h2 {
     margin: 0;
 }


teoli2003:
Hi @joshcrain. Do you have the URL where this example shows on MDN. To review the change, it would be useful to have the textual context where it appears.

joshcrain:
@teoli2003 Here is the URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#try_it

teoli2003 (approved):
Thank you!

PR #2199, opened 2022-06-30
feat(css): Add example of `:indeterminate` and `:defined`

NiedziolkaMichal:
Those are interactive examples for [:indeterminate](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate) and [:defined](https://developer.mozilla.org/en-US/docs/Web/CSS/:defined).

image

image


live-examples/css-examples/pseudo-class/defined.html
```diff
@@ -0,0 +1,9 @@
+<div class="all-elements">
+    <p>This is ordinary element <b>p</b>.</p>
+
+    <invalid>This is not defined element <b>invalid</b>.</invalid>
+
+    <not-registered>This is not defined element <b>not-registered</b>.</not-registered>
+
+    <registered-custom>This is defined element <b>registered-custom</b>.</registered-custom>
+</div>
```
estelle:
```suggestion
<div class="all-elements">
    <p>This is an ordinary <b>&;lt;p&gt;</b> element.</p>

    <invalid>This <b>&;lt;invalid&gt;</b> element is not defined, but has no hyphen.</invalid>

    <not-registered>This <b>&;lt;not-registered&gt;</b> element is neither defined nor registered.</not-registered>

    <registered-custom>This <b>&;lt;registered-custom&gt;</b> element is registered.</registered-custom>
</div>
```
Not sure if it was clear that elements without hyphens will match `:defined` even if they don't exist

live-examples/css-examples/pseudo-class/indeterminate.html
```diff
@@ -0,0 +1,12 @@
+<form>
+    <p>How did you find out about us?</p>
+    <label><input name="origin" type="radio" value="google"> Google</label>
+    <label><input name="origin" type="radio" value="facebook"> Facebook</label>
```
estelle:
```suggestion
    <label><input name="origin" type="radio" value="duck"> Duck-duck-go</label>
    <label><input name="origin" type="radio" value="mastodon"> Mastodon</label>
```
wanted to give a little love to the smaller players

estelle (approved):
thank you.

PR #2200, opened 2022-06-30
marker

NiedziolkaMichal:
Interactive Example for pseudo-element [::marker](https://developer.mozilla.org/en-US/docs/Web/CSS/::marker).

image


wbamberg (approved):

PR #2201, opened 2022-07-01
Removed .output rules from tabbed examples

### LINKED ISSUE #2171: <select> example not working
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

The default image is hamster and changing the animal with the select menu does nothing.

Using Chrome Version 102.0.5005.61 (Official Build) (x86_64) on macOs 12.4

### MDN metadata

<details>

* Folder: `en-us/web/html/element/select`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/select/index.md
* Last commit: https://github.com/mdn/content/commit/23c16d3e3f5229cc673a8a1f58d2b06533e3cebd
* Document last modified: 2022-05-15T03:56:08.000Z

</details>

NiedziolkaMichal:
This PR removes all rules in tabbed(html) examples refering to class "output". This class is used by HTML element living outside of the example and which is wrapping whole example output content. It is not shown in HTML tab, therefore it can be found only by using dev tools. I believe it might be confusing to the users, as they might be searching for this element and they will not be able to find it.

Elements embed, iframe, object, picture & source are setting #eee background color by using .output. Removing those rules doesn't have any visual impact, as #eee is very close to #fff used in other examples.

Element select is setting background image. Removing this rule makes this example look similar to [option](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option) example and fixes #2171 

Element wbr is setting background color to grey. This has significant impact, as edges of the inner white element can be easily seen. I have removed grey background and created dashed border instead, just like in CSS example [line-break](https://developer.mozilla.org/en-US/docs/Web/CSS/line-break).

Other examples are just changing font family by declaration "font: 1rem 'Fira Sans', sans-serif;". Removing this has very low impact and makes font family uniform with other examples. In global attribute tabindex and contenteditable I had to set font-weight to bold, to keep previews style.

Examples accesskey, dir, enterkeyhint, spellcheck, style don't have any content in CSS tab now. If that's a problem I can add something to them or if [#812](https://github.com/mdn/bob/pull/812) gets passed, we could remove css tab from those examples.


NiedziolkaMichal:
I also removed hamser image, which is no longer used by any example.

wbamberg:
> I also removed hamser image, which is no longer used by any example.

😭 well I'm going to miss the little guy, but it's about time he went I suppose.

wbamberg (approved):
👍 nice change @NiedziolkaMichal .

PR #2202, opened 2022-07-01
not, is, where

NiedziolkaMichal:
Interactive examples for pseudo-classes [:not](https://developer.mozilla.org/en-US/docs/Web/CSS/:not), [:is](https://developer.mozilla.org/en-US/docs/Web/CSS/:is), [:where](https://developer.mozilla.org/en-US/docs/Web/CSS/:where).

image

image

image


wbamberg (approved):

PR #2203, opened 2022-07-03
content, attr()

NiedziolkaMichal:
Those are interactive examples for [content](https://developer.mozilla.org/en-US/docs/Web/CSS/content) and [attr()](https://developer.mozilla.org/en-US/docs/Web/CSS/attr).

image

image


wbamberg (approved):

NiedziolkaMichal:
@wbamberg Thank you for all reviews.

PR #2206, opened 2022-07-05
Array/TypedArray .findLast and findLastInded examples

hamishwillee:
This adds examples for the Array and TypedArray methods `findL:ast()` and `findLastIndex()`. The methods are pretty much the same as find and findIndex, and the examples therefore follow the pattern of those examples.

This is part of documenting the methods for FF103, which can be tracked in https://github.com/mdn/content/issues/17473

### DIFF
--- a/live-examples/js-examples/array/array-findlast.js
+++ b/live-examples/js-examples/array/array-findlast.js
@@ -1,6 +1,6 @@
 const array1 = [5, 12, 50, 130, 44];
 
-const found = array1.findLast(element => element > 45);
+const found = array1.findLast((element) => element > 45);
 
 console.log(found);
 // expected output: 130
--- a/live-examples/js-examples/array/array-findlast.js
+++ b/live-examples/js-examples/array/array-findlast.js
@@ -1,0 +1,6 @@
+const array1 = [5, 12, 50, 130, 44];
+
+const found = array1.findLast(element => element > 45);
+
+console.log(found);
+// expected output: 130
--- a/live-examples/js-examples/typedarray/typedarray-findlast.js
+++ b/live-examples/js-examples/typedarray/typedarray-findlast.js
@@ -1,0 +1,8 @@
+function isNegative(element, index, array) {
+  return element < 0;
+}
+
+const int8 = new Int8Array([10, 0, -10, 20, -30, 40, 50]);
+
+console.log(int8.find(isNegative));
+// expected output: -30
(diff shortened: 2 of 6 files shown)


live-examples/js-examples/array/array-findlast.js
```diff
@@ -0,0 +1,6 @@
+const array1 = [5, 12, 50, 130, 44];
+
+const found = array1.findLast(element => element > 45);
```
teoli2003: Nit. A matter of taste, but also to be coherent with the rest of this PR.
```suggestion
const found = array1.findLast((element) => element > 45);
```

live-examples/js-examples/typedarray/typedarray-findlastindex.js
```diff
@@ -0,0 +1,8 @@
+function isNegative(element, index, array) {
```
teoli2003: I don't know the convention here, so this is more of a question. But shouldn't we have (eventually with a comment saying `index` and `array` are not needed in this case):
```suggestion
function isNegative(element) {
```
This would fix the test failure.

teoli2003 (changes requested):
1 nit and 1 comment. First time reviewing new stuff in `interactive-examples`.

live-examples/js-examples/typedarray/typedarray-findlastindex.js
hamishwillee: Thanks @teoli2003. I'm going to instead comment the unused elements. They are almost never used, but I think it is worth making the point, even here, that they are supplied. If you strongly disagree I can strip them out - they are almost never used.

```suggestion
function isNegative(element /*, index, array */) {
```

live-examples/js-examples/typedarray/typedarray-findlast.js
```diff
@@ -0,0 +1,8 @@
+function isNegative(element, index, array) {
```
hamishwillee:
```suggestion
function isNegative(element /*, index, array */) {
```

live-examples/js-examples/array/array-findlast.js
hamishwillee: THanks!

live-examples/js-examples/typedarray/typedarray-findlastindex.js
teoli2003: No, I think this is the good way of doing it.

teoli2003 (approved):
👍

PR #2207, opened 2022-07-05
scroll-snap-align

NiedziolkaMichal:
This is interactive example for property [scroll-snap-align](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-align).

image


estelle (approved):
thank you

PR #2208, opened 2022-07-06
namespace

NiedziolkaMichal:
This is interactive example for [@namespace](https://developer.mozilla.org/en-US/docs/Web/CSS/@namespace)

image


wbamberg (approved):

PR #2209, opened 2022-07-07
supports

NiedziolkaMichal:
This is interactive example for [@supports](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports).

image


wbamberg (approved):
It's tricky (impossible?) to make a good example out of this because the reader can't really toggle `@supports`. But this is fine :).

NiedziolkaMichal:
Yes, unfortunately user is unable to see full capability of this feature, without testing it on some other/older browser. I was thinking to add another @supports with something clearly invalid in there, like @supports (invalid-property: invalid-value), but whole example would get too long.

PR #2210, opened 2022-07-09
media

NiedziolkaMichal:
This is interactive example for [@media](https://developer.mozilla.org/en-US/docs/Web/CSS/@media).

image


wbamberg:
I had a couple of questions about this one. The first is that a rule like this:

```css
@media (hover: hover) {
    abbr:hover {
        color: limegreen;
        transition-duration: 1s;
    }
}
```

...is strange to me. If hover isn't supported, then `abbr:hover` would have no effect anyway, would it? So isn't this just equivalent to:

```css
abbr:hover {
    color: limegreen;
    transition-duration: 1s;
}
```

? I would have thought this:

```css
@media (hover: hover) {
  abbr {
    color: limegreen;
  }
}
```

...would make more sense, as an example?

This is like https://github.com/mdn/content/issues/12400, which was marked fixed by https://github.com/mdn/content/pull/13689, but I also don't understand why that fixed it.

Also with this rule:

```css
@media not all and (hover: hover) {
    abbr::after {
        content: ' (' attr(title) ')';
    }
}
```

...I can't get this rule to fire at all. But if I do this:

```css
@media screen and (hover: hover) {
    abbr::after {
        content: ' (' attr(title) ')';
    }
}
```

...then I can see the effect (I see the `content` on the screen but not in print preview).

Am I confused here?

NiedziolkaMichal:
Many touch screen devices simulate hover state, even though they don't show cursor on the screen. For example my iphone 13 will activate :hover pseudo-class when I press my finger on a link and hold it for few seconds, but when I release it, :hover will still be active. The same devices return value "none" in "hover" media feature, that's why it is quite common to use media query from my example: [https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/](https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/)

Do you think it's better to replace it with something else?

wbamberg (approved):
Thanks for the clarification, let's merge this one :).

PR #2211, opened 2022-07-10
color-scheme

NiedziolkaMichal:
This is interactive example for [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme).

image


wbamberg (approved):
👍 thank you!

PR #2212, opened 2022-07-11
feat(html): Add example of `<dialog>`

NiedziolkaMichal:
It's interactive example for [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element.

image

image

image


live-examples/html-examples/interactive-elements/dialog.html
```diff
@@ -0,0 +1,13 @@
+<div class="card">
+  <p class="question">Would you like to try your luck in throwing a dice?</p>
```
bsmth:
```suggestion
  <p class="question">Would you like to try your luck in a throw of the dice?</p>
```

bsmth (approved):
Thanks @NiedziolkaMichal! Merging shortly 👍🏻

PR #2213, opened 2022-07-12
menu

NiedziolkaMichal:
This is interactive example for [menu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) element.

image


wbamberg (approved):
👍 nice work as always @NiedziolkaMichal !

PR #2217, opened 2022-07-16
Add CSS break-before, break-after, break-inside properties

NiedziolkaMichal:
Those are interactive examples for CSS properties [break-before](https://developer.mozilla.org/en-US/docs/Web/CSS/break-before), [break-after](https://developer.mozilla.org/en-US/docs/Web/CSS/break-after), [break-inside](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside), [page-break-before](https://developer.mozilla.org/en-US/docs/Web/CSS/page-break-before), [page-break-after](https://developer.mozilla.org/en-US/docs/Web/CSS/page-break-after), [page-break-inside](https://developer.mozilla.org/en-US/docs/Web/CSS/page-break-inside).

image
When button is pressed and value page or left is selected, first page looks like that:
image
Second page looks like that:
image

image

image

image

image

image

In break-inside & page-break-inside, middle box has very big height when document is in print preview. Value avoid looks like that:
image

### DIFF
--- a/live-examples/css-examples/fragmentation/break-after.html
+++ b/live-examples/css-examples/fragmentation/break-after.html
@@ -27,9 +27,9 @@
             <p>The effect of this property can be noticed when the document is being printed or a preview of a print is displayed.</p>
             <button id="print-btn">Show Print Preview</button>
             <div class="box-container">
-                <div class="box">Content Before</div>
-                <div class="box" id="example-element">Content with break-after</div>
-                <div class="box">Content After</div>
+                <div class="box">Content before the property</div>
+                <div class="box" id="example-element">Content with 'break-after'</div>
+                <div class="box">Content after the property</div>
             </div>
         </div>
     </section>
--- a/live-examples/css-examples/fragmentation/break-after.html
+++ b/live-examples/css-examples/fragmentation/break-after.html
@@ -12,13 +12,6 @@
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
     </div>
-
-    <div class="example-choice">
-        <pre><code class="language-css">break-after: left;</code></pre>
-        <button type="button" class="copy hidden" aria-hidden="true">
-            <span class="visually-hidden">Copy to Clipboard</span>
-        </button>
-    </div>
 </section>
 
 <div id="output" class="output large hidden">
--- a/live-examples/css-examples/fragmentation/break-after.html
+++ b/live-examples/css-examples/fragmentation/break-after.html
@@ -1,0 +1,36 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="break-after">
+    <div class="example-choice">
+        <pre><code class="language-css">break-after: auto;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">break-after: page;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">break-after: left;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+</section>
+
+<div id="output" class="output large hidden">
+    <section id="default-example">
+        <div>
+            <p>The effect of this property can be noticed when the document is being printed or a preview of a print is displayed.</p>
+            <button id="print-btn">Show Print Preview</button>
+            <div class="box-container">
+                <div class="box">Content Before</div>
+                <div class="box" id="example-element">Content with break-after</div>
+                <div class="box">Content After</div>
+            </div>
+        </div>
+    </section>
+</div>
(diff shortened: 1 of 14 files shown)


live-examples/css-examples/fragmentation/break-before.html
```diff
@@ -0,0 +1,36 @@
...
+
+<div id="output" class="output large hidden">
+    <section id="default-example">
+        <div>
+            <p>The effect of this property can be noticed when the document is being printed or a preview of a print is displayed.</p>
+            <button id="print-btn">Show Print Preview</button>
+            <div class="box-container">
+                <div class="box">Content Before</div>
+                <div class="box" id="example-element">Content with break-before</div>
+                <div class="box">Content After</div>
```
dipikabh: How about rephrasing to the following to add more clarity:

```suggestion
                <div class="box">Content before the property</div>
                <div class="box" id="example-element">Content with 'break-before'</div>
                <div class="box">Content after the property</div>
```

This suggestion can be applied to other html files in this pull request

live-examples/css-examples/fragmentation/break-inside.html
```diff
@@ -0,0 +1,36 @@
...
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">break-inside: avoid-page;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
dipikabh: The description for the two values `avoid` and `avoid-page` [reads](https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside#values) quite similar and the output of this example here also does not help to clarify the difference in behavior. 
Do you think filling in dummy content would help to bring out the difference between the two?

live-examples/css-examples/fragmentation/break-before.html
```diff
@@ -0,0 +1,36 @@
...
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">break-before: left;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
dipikabh: Currently, there is no difference visually in the print preview for `page` and `left`.
- Would the contrast be more visible if you instead demo the `page` and `right` values? And also perhaps add one more "box-container" div if needed to highlight the difference?
- In the print preview, can the `Pages per sheet` option be set to `2` to show the pages side-by-side?

live-examples/css-examples/fragmentation/break-after.html
```diff
@@ -0,0 +1,36 @@
...
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">break-after: left;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
```
dipikabh: Same comment as in 'break-before' to highlight the difference in the print preview between `page` and `left`.

dipikabh (changes requested):
Apologies for the delay on getting back on this, @NiedziolkaMichal. I've added a few comments for your consideration.

dipikabh:
Hi @NiedziolkaMichal, hope you're good. Pinging to see if you had the chance to see the comments.

live-examples/css-examples/fragmentation/break-before.html
NiedziolkaMichal: I am afraid that values `left` and `right` are not really working on Firefox or Chrome: https://caniuse.com/css-page-break
Should I remove them to avoid confusion?
I also don't see any way to affect the default value of `Pages per sheet`.

live-examples/css-examples/fragmentation/break-inside.html
NiedziolkaMichal: Those values work exactly the same way, except for the fact, that `avoid` is also supposed to work in a multi-column layout. In the future, we could also show how multi-column layout is affected, but currently, it's not supported well on firefox. Do you think it's better to remove one of those values?

live-examples/css-examples/fragmentation/break-before.html
NiedziolkaMichal: That's better. I have applied your suggestion to all examples. Thanks

NiedziolkaMichal:
@dipikabh Sorry for the delay. I have responded to all of your suggestions.

live-examples/css-examples/fragmentation/break-inside.html
dipikabh: Yes, I think in that case it might be better to remove one

live-examples/css-examples/fragmentation/break-before.html
dipikabh: We can keep the example if you think it will not cause confusion or make people wonder about the outputs. Otherwise, let's add these examples later when we're able to better demonstrate them.

dipikabh:
Thanks, @NiedziolkaMichal! I've responded to your questions.

live-examples/css-examples/fragmentation/break-inside.html
NiedziolkaMichal: Alright, I have removed `avoid-page`.

live-examples/css-examples/fragmentation/break-before.html
NiedziolkaMichal: I think it's better to keep this example. I removed `left` values, so there will be no confusion.

dipikabh (approved):
Thanks, @NiedziolkaMichal!

NiedziolkaMichal:
Thanks a lot for reviewing.

PR #2218, opened 2022-07-16
Fix property names in meta.json to camelCase

NiedziolkaMichal:
This PR simply changes few property names in meta.json which use '-' as word separator, to use camelCase instead.


wbamberg (approved):
👍 thank you!

PR #2219, opened 2022-07-17
feat(html): Add example of `<template>`

NiedziolkaMichal:
This is interactive example for html element [template](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).

image


wbamberg:
I think this depends on the shadow DOM fix, doesn't it?

NiedziolkaMichal:
It does :/ I hope we can get someone to review shadow dom pr soon

estelle (approved):

PR #2220, opened 2022-07-18
feat(html): Add example of  `draggable` attribute

NiedziolkaMichal:
This is an interactive example of a global attribute [draggable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable).

image


wbamberg (commented):
With this one, do you think it is worth waiting until we can have JS, and implementing a basic drag-and-drop example? There doesn't seem much point in having something draggable if you can't drop it :)

NiedziolkaMichal:
@wbamberg Yes, that's a good idea

wbamberg (approved):
👍 looks great to me (and sorry to be so slow at reviewing these).

PR #2224, opened 2022-07-19
feat(html): Add example of `exportparts` attribute

NiedziolkaMichal:
This is interactive example for [exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts) attribute.

image

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-exportparts.html
+++ b/live-examples/html-examples/global-attributes/attribute-exportparts.html
@@ -1,0 +1,21 @@
+<template id="shadowStructure">
+    <p part="paragraph">
+        This paragraph is in Shadow Tree and can be reached because part attribute is set.
+    </p>
+    <p class="not-exported-paragraph">
+        This paragraph is in Shadow Tree and cannot be reached because part attribute is NOT set.
+    </p>
+
+    <template id="nestedShadowStructure">
+        <p part="nested-paragraph">
+            This paragraph exists in nested Shadow Tree and can be reached because its part is set and exported.
+        </p>
+        <p part="skipped-nested-paragraph">
+            This paragraph exists in nested Shadow Tree but its part is NOT exported.
+        </p>
+    </template>
+
+    <div id="nestedShadowHost" exportparts="nested-paragraph"></div>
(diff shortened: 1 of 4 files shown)


live-examples/html-examples/global-attributes/attribute-exportparts.html
```diff
@@ -0,0 +1,21 @@
...
+    <p class="not-exported-paragraph">
+        This paragraph is in Shadow Tree and cannot be reached because part attribute is NOT set.
+    </p>
+
+    <template id="nestedShadowStructure">
+        <p part="nested-paragraph">
+            This paragraph exists in nested Shadow Tree and can be reached because its part is set and exported.
+        </p>
+        <p part="skipped-nested-paragraph">
+            This paragraph exists in nested Shadow Tree but its part is NOT exported.
```
sideshowbarker:
```suggestion
        This paragraph is in a shadow tree and can be reached because the <code>part</code> attribute is set.
    </p>
    <p class="not-exported-paragraph">
        This paragraph is in a shadow tree and cannot be reached because the <code>part</code> attribute is <em>not</em> set.
    </p>

    <template id="nestedShadowStructure">
        <p part="nested-paragraph">
            This paragraph exists in a nested shadow tree and can be reached because its <code>part</code> attribute is set and exported.
        </p>
        <p part="skipped-nested-paragraph">
            This paragraph exists in a nested shadow tree but its <code>part</code> attribute is <em>not</em> exported.
```

live-examples/html-examples/global-attributes/css/attribute-exportparts.css
```diff
@@ -0,0 +1,18 @@
+#shadowHost::part(paragraph) {
+    color: green;
+}
+
+.not-exported-paragraph {
+    /* This value will not be applied, because selector
+       cannot find elements present in Shadow DOM */
```
sideshowbarker:
```suggestion
    /* This value will not be applied, because the selector
       cannot find elements present in the shadow DOM */
```

sideshowbarker (approved):

PR #2225, opened 2022-07-19
Fix invalid data-property values

NiedziolkaMichal:
This PR fixes invalid value of "data-property" attribute in two css properties. That attribute is used by BOB to determine whether browser supports presented property, so it can show examples instead of a fallback dialog.


wbamberg (approved):

PR #2227, opened 2022-07-21
offset

### LINKED ISSUE #1643: CSS examples: need example for offset
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/offset

This is a shorthand property for the properties used to animate an element along a path.

This is a relatively complex example to implement, but there are some nice interactive examples for some of the longhand properties, that could be adapted for this one. In particular https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/offset-path seems like a reasonable way to approach this. Note that this example uses a "Play/Pause" button to start and stop the animation, and this might be a good approach here rather than playing the animation automatically. The example for https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/offset-anchor uses the same approach.

NiedziolkaMichal:
This is interactive example for [offset](https://developer.mozilla.org/en-US/docs/Web/CSS/offset).

image

Fixes #1643

### DIFF
--- a/live-examples/css-examples/motion-path/meta.json
+++ b/live-examples/css-examples/motion-path/meta.json
@@ -1,35 +1,43 @@
 {
-  "pages": {
-    "offsetAnchor": {
-        "cssExampleSrc": "./live-examples/css-examples/motion-path/offset-anchor.css",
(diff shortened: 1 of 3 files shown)


wbamberg (approved):
👍 looks great.

PR #2228, opened 2022-07-21
updates hgroup example

scottaohara:
per recent changes to `hgroup` element, reflected in the updated content of the [`hgroup`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup) MDN doc, the try it example needed to be updated as well.

replaced `h2` element with `p` element

### DIFF
--- a/live-examples/html-examples/content-sectioning/hgroup.html
+++ b/live-examples/html-examples/content-sectioning/hgroup.html
@@ -1,5 +1,5 @@
 <hgroup>
     <h1>Calculus I</h1>
-    <h2>Fundamentals</h2>
+    <p>Fundamentals</p>
 </hgroup>
 <p>This course will start with a brief introduction about the limit of a function. Then we will describe how the idea of derivative emerges in the Physics and Geometry fields. After that, we will explain that the key to master calculus is &hellip;</p>


jasikpark:
++ for this change, the description below the example describes that a single heading element is permitted, while the current example has two.

image
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup#try_it

jasikpark:
same info here: https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model

scottaohara:
@estelle per the updates i made to the `hgroup` article the other day, this example needed to be updated as well.  Is there a separate team that reviews changes to this repo?

estelle:
@scottaohara The interactive examples are in a separate repo: https://github.com/mdn/interactive-examples. 

I am not sure who generally reviewed these. I was added to that repo last week, so have been reviewing a backlog. if you mention me with @estelle, I will see it in my notifications, which I tend to check at least every other day.

estelle (approved):
thank you

estelle:
i happen to be in there right now, so I'll make the edit.

jasikpark:
woot 🥳

PR #2230, opened 2022-07-22
overflow-anchor

NiedziolkaMichal:
This is interactive example for [overflow-anchor](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor).

image


wbamberg (approved):
Very illustrative example, nice work @NiedziolkaMichal .

PR #2233, opened 2022-07-24
display

NiedziolkaMichal:
This is interactive example for [display](https://developer.mozilla.org/en-US/docs/Web/CSS/display).

image

### DIFF
--- a/live-examples/css-examples/basic-box-model/display.css
+++ b/live-examples/css-examples/basic-box-model/display.css
@@ -1,0 +1,15 @@
+.example-container {
+    width: 100%;
+    height: 100%;
+}
+
+#example-element {
+    border: 1px solid #c5c5c5;
+}
+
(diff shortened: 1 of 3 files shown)


wbamberg (commented):
`display` is a tricky one because it feels more like a meta-property these days. But this is nice!

However I wonder why you have given the prominent border to the children and the subtle one to the container, when the container is the example element. To me that implies that it's the children we are changing.

What do you think about just swapping the border styles:

Screen Shot 2022-09-28 at 7 41 26 PM

----------------

Screen Shot 2022-09-28 at 7 44 36 PM

?

NiedziolkaMichal:
You are right. On light mode this example is hard to understand. I have changed colors, so now they look like this:
image
image

Rumyra (approved):

PR #2234, opened 2022-07-25
feat(html): Add example of `is` global attribute

NiedziolkaMichal:
This is interactive example for [is](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is) attribute.
This PR depends on Shadow DOM fix.

image

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-is.html
+++ b/live-examples/html-examples/global-attributes/attribute-is.html
@@ -1,5 +1,5 @@
 <p>Please describe why would you like to work for us:</p>
 
 <div class="length-textarea-container">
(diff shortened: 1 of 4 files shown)


live-examples/html-examples/global-attributes/attribute-is.html
```diff
@@ -0,0 +1,5 @@
+<p>Please describe why would you like to work for us:</p>
+
+<div class="length-textarea-container">
+    <textarea is="length-left-textarea" maxLength="150">I would like to work for you because ...</textarea>
```
sideshowbarker:
```suggestion
    <textarea is="length-left-textarea" maxLength="150">I would like to work for you because…</textarea>
```

sideshowbarker (approved):

PR #2237, opened 2022-07-27
rgb, rgba, hsl, hsla, hwb

NiedziolkaMichal:
Those are interactive examples for color values [rgb](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb), [rgba](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgba), [hsl](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl), [hsla](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsla), [hwb](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hwb).

I also removed 1 example from [color](https://developer.mozilla.org/en-US/docs/Web/CSS/color), so there is no vertical scrollbar.

image

image

image

image

image

### DIFF
--- a/live-examples/css-examples/color/color-function.css
+++ b/live-examples/css-examples/color/color-function.css
@@ -1,0 +1,5 @@
+#example-element {
+    min-width: 100%;
+    min-height: 100%;
+    padding: 10%;
+}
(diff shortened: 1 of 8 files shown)


teoli2003 (approved):
👍

PR #2238, opened 2022-07-28
WebAssembly | Control flow - update `return` and `select`

ghost:
Cooperative PR to https://github.com/mdn/content/pull/18838; these may be merged out of sync/order.

Updates the `return` control flow instruction example to show discarding values from the Wasm value stack.

Adds an example for `select t`, using WebAssembly external reference types.

### DIFF
--- a/live-examples/wat-examples/statements/return.wat
+++ b/live-examples/wat-examples/statements/return.wat
@@ -1,11 +1,14 @@
 (module
-
   (func (export "get_90") (result i32)
-
     ;; load 90 onto the stack and return it
     i32.const 90
     return
-
   )
-
+  (func (export "get_second_value") (param i32 i32) (result i32)
+    ;; load both arguments onto the stack
+    local.get 0
+    local.get 1
+    ;; return the second argument; the first is discarded
+    return
+  )
 )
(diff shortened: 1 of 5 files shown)


live-examples/wat-examples/statements/return.wat
```diff
@@ -1,11 +1,14 @@
...
   )
-
+  (func (export "get_second_value") (param i32 i32) (result i32)
+    ;; load both arguments onto the stack
+    local.get 0
+    local.get 1
+    ;; return the second argument; the first is discarded
+    return
+  )
 )
```
MendyBerger, line 10: I don't think two examples are need here, this having one example with 2 values on the stack is enough.

MendyBerger (changes requested):
Should not be merged before https://github.com/mdn/bob/pull/842

ghost:
@schalkneethling May we have have a review here, as you merged the blocker PR?

live-examples/wat-examples/statements/select.wat
```diff
@@ -1,15 +1,18 @@
 (module
-  (import "console" "log" (func $log (param i32)))
-  (func $main
+  (func (export "select_10_or_20") (result i32)
```
MendyBerger: Maybe rename the functions to something like `select_simple` and `select_externref`?

MendyBerger (approved):

ghost:
@Josh-Cena Seeing as you assigned the reviewer, may we get someone to review for merging?

Josh-Cena (approved):
Oh—let's go then

PR #2239, opened 2022-07-28
Proper contrast in min() & max()

NiedziolkaMichal:
This PR changes image used in [min()](https://developer.mozilla.org/en-US/docs/Web/CSS/min) and [max()](https://developer.mozilla.org/en-US/docs/Web/CSS/max) examples. First I thought to change color of the image, but it's also used in other places.

Before:
image

After:
image

### DIFF
--- a/live-examples/css-examples/values-and-units/function-max.html
+++ b/live-examples/css-examples/values-and-units/function-max.html
@@ -24,7 +24,7 @@
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
       <div id="example-element" class="transition-all">
-        <img src="../../media/examples/dino.svg" alt="MDN Dino logo" class="logo">
+        <img src="../../media/examples/firefox-logo.svg" alt="MDN Dino logo" class="logo">
       </div>
     </section>
 </div>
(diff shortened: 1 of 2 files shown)


live-examples/css-examples/values-and-units/function-max.html
```diff
@@ -24,7 +24,7 @@
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
       <div id="example-element" class="transition-all">
-        <img src="../../media/examples/dino.svg" alt="MDN Dino logo" class="logo">
+        <img src="../../media/examples/firefox-logo.svg" alt="MDN Dino logo" class="logo">
```
estelle:
```suggestion
        <img src="../../media/examples/firefox-logo.svg" alt="Firefox logo" class="logo">
```

live-examples/css-examples/values-and-units/function-min.html
```diff
@@ -24,7 +24,7 @@
 <div id="output" class="output large hidden">
     <section id="default-example" class="default-example">
       <div id="example-element" class="transition-all">
-        <img src="../../media/examples/dino.svg" alt="MDN Dino logo" class="logo">
+        <img src="../../media/examples/firefox-logo.svg" alt="MDN Dino logo" class="logo">
```
estelle:
```suggestion
        <img src="../../media/examples/firefox-logo.svg" alt="Firefox logo" class="logo">
```

NiedziolkaMichal:
Thank you for the correction and for the review

PR #2242, opened 2022-07-29
accept, capture

NiedziolkaMichal:
Those are interactive examples for attributes [accept](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept) & [capture](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture).

image

image


wbamberg (approved):
👍 thank you, @NiedziolkaMichal !

I like the formatting of the HTML source here. Makes me wonder if we shouldn't go back through some of the others.

PR #2243, opened 2022-07-30
minlength, maxlength

NiedziolkaMichal:
Those are interactive examples for HTML attributes [minlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/minlength) & [maxlength](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength).

image

image


wbamberg (approved):

PR #2247, opened 2022-08-09
pattern

NiedziolkaMichal:
This is interactive example for html attribute [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).

image


wbamberg (commented):
Maybe it's just me and regular expressions, but I think the second example is too complex to be a good example. I think just the first example should be fine - given the principle that you can uses regexes, it follows that you can make them as complicated as you like.

Also [our docs say](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#usability) "Additionally, include a [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) attribute which gives a description of the pattern." - although I'm not sure if that is really good advice.

NiedziolkaMichal:
I think example would look a bit empty with just a single box, so I replaced password with PIN:

image

I believe it's better to place description of the pattern in the label, because content of title attribute is shown only after placing mouse over textbox for at least few seconds, so users could miss it.

Rumyra:
I'm good with this - what do you think @wbamberg ?

wbamberg (approved):
👍 sorry to be slow again @NiedziolkaMichal .

PR #2248, opened 2022-08-10
readonly

NiedziolkaMichal:
This is interactive example for HTML attribute [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly).

image


wbamberg (approved):
👍

PR #2249, opened 2022-08-11
multiple

NiedziolkaMichal:
This is interactive example for HTML attribute [multiple](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/multiple).

image


wbamberg (approved):
👍 great example, very illustrative.

PR #2253, opened 2022-08-19
feat: showcase basic array functionality

hirehamir:
The goal of this is to add an interactive example to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array.

I did something similar for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map with #1855.


hamirmahal:
## Demo after `yarn run build` and `yarn start`

https://user-images.githubusercontent.com/43425812/185542756-ca51552a-ff40-436b-becd-74d69657eb28.mp4

hamirmahal:
I'm not sure how big of an issue it is, but the result of

```
git diff ca3a22dd3666e30454197f81eeaa61257a3c1bdb
```

includes this output.

```diff
     "pages": {
+        "array": {^M
+            "exampleCode": "./live-examples/js-examples/array/array.js",^M
+            "fileName": "array.html",^M
+            "title": "JavaScript Demo: Array",^M
+            "type": "js"^M
+        },^M
         "arrayAt": {
             "exampleCode": "./live-examples/js-examples/array/array-at.js",
             "fileName": "array-at.html",
```

I don't know how much of an issue the `^M`s will be.

hamirmahal:
@Rumyra let me know if there is anything I can do to improve this pull request.

NiedziolkaMichal:
Hey! What do you think about improving this example a little bit, so more features of an array are presented in a short example? I am thinking about following features, but you choose differently:
1. Arrays are zero-indexed and its values can be retrieved with []
2. Arrays can contain mix of different data types
3. It might also be good to show usage of length property, which is very common.

Rumyra:
Sorry to take a while to get back - yes I love the idea of an example of the array page and agree with @NiedziolkaMichal here 

>Arrays are zero-indexed and its values can be retrieved with []
>Arrays can contain mix of different data types
>It might also be good to show usage of length property, which is very common.

How do you feel about that @hamirmahal ?

hamirmahal:
@Rumyra that sounds excellent! I just implemented those changes.

hamirmahal:
> Hey! What do you think about improving this example a little bit, so more features of an array are presented in a short example? I am thinking about following features, but you choose differently:
> 
> 1. Arrays are zero-indexed and its values can be retrieved with []
> 2. Arrays can contain mix of different data types
> 3. It might also be good to show usage of length property, which is very common.

Excellent suggestion! My recent changes should address this.

hamirmahal:
> Hey! What do you think about improving this example a little bit, so more features of an array are presented in a short example? I am thinking about following features, but you choose differently:
> 
> 1. Arrays are zero-indexed and its values can be retrieved with []
> 2. Arrays can contain mix of different data types
> 3. It might also be good to show usage of length property, which is very common.

Excellent suggestions! My recent changes should address this.

hamirmahal:
@Rumyra what changes would you like to see here before merging this pull request?

Josh-Cena (changes requested):
This seems overly loaded to me. IMO (but I'm not too keen with interactive examples anyway) the point of these examples are to give a "taste" of the feature, so the third example:

```js
const array2 = [123, 'abc', { objectWithAge: 67 }, null];
const nestedArrays = [
  array2,
  [],
  [
    [1, 2, 3],
    [4, 5, 6],
  ],
];
console.log(nestedArrays.length); // 3
```

Seems unnecessary. I think the first two are fine.

hamirmahal:
That third example was to address previous comments saying to showcase arrays showing different types.

> Hey! What do you think about improving this example a little bit, so more features of an array are presented in a short example? I am thinking about following features, but you choose differently:
> 
> 1. Arrays are zero-indexed and its values can be retrieved with []
> 2. Arrays can contain mix of different data types
> 3. It might also be good to show usage of length property, which is very common.



> Sorry to take a while to get back - yes I love the idea of an example of the array page and agree with @NiedziolkaMichal here
> 
> > Arrays are zero-indexed and its values can be retrieved with []
> > Arrays can contain mix of different data types
> > It might also be good to show usage of length property, which is very common.
> 
> How do you feel about that @hamirmahal ?

hamirmahal:
That third example was to address previous comments saying to showcase arrays having different types.

> Hey! What do you think about improving this example a little bit, so more features of an array are presented in a short example? I am thinking about following features, but you choose differently:
> 
> 2. Arrays can contain mix of different data types



> Sorry to take a while to get back - yes I love the idea of an example of the array page and agree with @NiedziolkaMichal here
> 
> > Arrays can contain mix of different data types
> 
> How do you feel about that @hamirmahal ?

Josh-Cena:
To showcase "Arrays can contain mix of different data types", we can simply add a `names.push(null)` or something similar if you wish, but to me it's not highly necessary. WDYT?

hamirmahal:
I think for now, we can leave this as is, but anyone who wants to make an
improvement to this in the future is welcome to do so.

Josh-Cena (approved):
Okay, let's do this. Thank you for sticking with us after so long!

hamirmahal:
You’re welcome! I’m glad I could help.

PR #2254, opened 2022-08-21
Include audio/video fallback downloads.

dracos:
As per e.g. the MDN audio help, “it's a good practice to provide some
content (such as the direct download link) as a fallback for viewers”
so include this good practice in the examples.

Links with https://github.com/mdn/content/pull/19790 to update other similar content on the page.


NiedziolkaMichal (approved):
That looks great. I have just fixed 2 extensions from webm to mp4. Thanks a lot!

PR #2262, opened 2022-08-31
Fix lint error

wbamberg:
Fix lint error introduced by https://github.com/mdn/interactive-examples/commit/c6b5dc61c6b558c58091b8d974e752e64fc040d2.


teoli2003 (approved):
I'm fine for this now, but note that, by default, Prettier uses `""`. It will be necessary to uniform this, but not now.

PR #2268, opened 2022-09-14
Add clarity to assignment behavior

shanimal:
Use a strict equality comparison to emphasize that `returnedTarget` is the same as the original `target` parameter.

### DIFF
--- a/live-examples/js-examples/object/object-assign.js
+++ b/live-examples/js-examples/object/object-assign.js
@@ -8,3 +8,6 @@ console.log(target);
 
 console.log(returnedTarget);
 // expected output: Object { a: 1, b: 4, c: 5 }
+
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
Hey!

Since you are showing that both target & returnedTarget are equal, it's pretty obvious that using console.log on them will print same result. What do you think about removing line 9 & 10, so example looks like this?

image

shanimal:
> Hey!
> 
> Since you are showing that both target & returnedTarget are equal, it's pretty obvious that using console.log on them will print same result. What do you think about removing line 9 & 10, so example looks like this?
> 
> image

Nice catch! Will do!

NiedziolkaMichal (commented):
That looks great, thank you for this change.

PR #2271, opened 2022-09-19
Removing closing tags in `source` and `img` tags in `picture` example

StatsProIO:
Example is visible on this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture


NiedziolkaMichal (changes requested):
I believe it's better to go other way around and remove closing tag from  element instead. Both of those elements don't require closing tag and most of their occurences in other files don't have them. Could you please remove closing slash from both elements?

NiedziolkaMichal (approved):

NiedziolkaMichal:
@MarbleMetrics That's a good change. Thanks a lot!

PR #2273, opened 2022-09-20
Remove closing tag from void elements

NiedziolkaMichal:
I have removed closing tags from all void elements, to make it consistent across all html files. All void elements were checked, but there were problems only with: br, hr, img, track, input, area

### DIFF
--- a/live-examples/css-examples/positioned-layout/position.html
+++ b/live-examples/css-examples/positioned-layout/position.html
@@ -41,8 +41,8 @@ top: 20px;</code></pre>
             <p class="clear">To see the effect of <code>sticky</code> positioning, select the <code>position: sticky</code> option and scroll this container.</p>
             <p>The element will scroll along with its container, until it is at the top of the container (or reaches the offset specified in <code>top</code>), and will then stop scrolling, so it stays visible.</p>
             <p>The rest of this text is only supplied to make sure the container overflows, so as to enable you to scroll it and see the effect.</p>
-            <hr/>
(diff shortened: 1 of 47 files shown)


live-examples/css-examples/positioned-layout/position.html
```diff
@@ -41,8 +41,8 @@
             <p class="clear">To see the effect of <code>sticky</code> positioning, select the <code>position: sticky</code> option and scroll this container.</p>
             <p>The element will scroll along with its container, until it is at the top of the container (or reaches the offset specified in <code>top</code>), and will then stop scrolling, so it stays visible.</p>
             <p>The rest of this text is only supplied to make sure the container overflows, so as to enable you to scroll it and see the effect.</p>
-            <hr/>
             <p>Far out in the uncharted backwaters of the unfashionable end of the western spiral arm of the Galaxy lies a small unregarded yellow sun. Orbiting this at a distance of roughly ninety-two million miles is an utterly insignificant little blue green planet whose ape-descended life forms are so amazingly primitive that they still think digital watches are a pretty neat idea.</p> 
+            <hr>
```
wbamberg: Why did you move this?

wbamberg (commented):
Nice!

Just one question.

Also, I like this and will merge it, but also see: https://github.com/mdn/mdn-community/discussions/242 - if we ever run Prettier on this repo we'll have a problem here.

live-examples/css-examples/positioned-layout/position.html
NiedziolkaMichal: It looks to be a mistake, thanks for noticing. I have made a commit with a fix.

wbamberg (approved):

PR #2274, opened 2022-09-20
Update CONTRIBUTING-CSS

bsmth:
This PR adds languages to code blocks, formatting fixes.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
     "singleQuote": true,
-    "tabWidth": 4
+    "tabWidth": 4,
+    "printWidth": 120
 }
--- a/CONTRIBUTING-CSS.md
+++ b/CONTRIBUTING-CSS.md
@@ -6,21 +6,23 @@ You start off by creating a new file inside the subfolder `live-examples\css-exa
 
 Inside this newly created file, copy and paste the following code:
 
-```
-<section id="example-choice-list" class="example-choice-list large" data-property="border-radius">
-
+```html
+<section
+    id="example-choice-list"
+    class="example-choice-list large"
+    data-property="border-radius"
+>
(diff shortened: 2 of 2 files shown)


NiedziolkaMichal:
I think it's better to keep &lt;section> elements in the single line. This page is was written to teach structure of the interactive example file, but also to allow contributors to easily copy-paste the entire fragment. All interactive examples have &lt;section> written without any line breaks, so guide should teach to do the same.

I am also not sure about this fragment:
```html
<section
    id="..."
    data-property="box-decoration-break -webkit-box-decoration-break"
></section>
```
This fragment is now larger and it's missing class attribute. I feel it's not worth to make it `html` type, for the sake of simplicity.

Overall you have done a great job. Could you do the same for other .md files?

bsmth:
hi @NiedziolkaMichal let me have another look, there might be a better way to handle this.

>Could you do the same for other .md files?

I can indeed, I propose to follow up with a separate PR once we land this one, how does that sound?

bsmth:
I've set the max line length to 120 for Prettier in https://github.com/mdn/interactive-examples/pull/2274/commits/5a1ba0cdd654f5ec40d8b79cc1abaec07bf8aba5 which seems to work better, what do you think?

bsmth:
>I am also not sure about this fragment:

```html
<section
    id="..."
    data-property="box-decoration-break -webkit-box-decoration-break"
></section>
```

I see your point, we could revert this if you think readers will know where to put it:

```
data-property="box-decoration-break -webkit-box-decoration-break"
```

bsmth:
@NiedziolkaMichal do you want to have another review?

NiedziolkaMichal:
It looks good as it is, thank you for this PR.

bsmth:
Thanks @NiedziolkaMichal

PR #2275, opened 2022-09-20
Add 3D translate, rotate, scale examples

NiedziolkaMichal:
Those are interactive examples for CSS properties [translate](https://developer.mozilla.org/en-US/docs/Web/CSS/translate), [rotate](https://developer.mozilla.org/en-US/docs/Web/CSS/rotate), [scale](https://developer.mozilla.org/en-US/docs/Web/CSS/scale) based on 3d functions.

image

image

image


bsmth:
Thanks @NiedziolkaMichal, taking a look 👀

live-examples/css-examples/transforms/meta.json
```diff
@@ -7,175 +7,196 @@
...
+        "perspective": {
+            "cssExampleSrc": "./live-examples/css-examples/transforms/transform3d.css",
+            "exampleCode": "./live-examples/css-examples/transforms/perspective.html",
+            "fileName": "perspective.html",
+            "title": "CSS Demo: perspective",
+            "type": "css"
+        },
+        "perspectiveOrigin": {
+            "cssExampleSrc": "./live-examples/css-examples/transforms/perspective-origin.css",
+            "exampleCode": "./live-examples/css-examples/transforms/perspective-origin.html",
```
bsmth, line 154: Looks unrelated, want to remove this?

live-examples/css-examples/transforms/meta.json
```diff
@@ -7,175 +7,196 @@
...
+        "functionTranslateZ": {
             "cssExampleSrc": "./live-examples/css-examples/transforms/transform3d.css",
             "exampleCode": "./live-examples/css-examples/transforms/function-translateZ.html",
             "fileName": "function-translateZ.html",
             "title": "CSS Demo: translateZ()",
             "type": "css"
+        },
+        "perspective": {
+            "cssExampleSrc": "./live-examples/css-examples/transforms/transform3d.css",
+            "exampleCode": "./live-examples/css-examples/transforms/perspective.html",
```
bsmth, line 147: Same as below, are these unrelated changes?

live-examples/css-examples/transforms/meta.json
```diff
@@ -7,175 +7,196 @@
...
-        },
-        "rotate": {
+        "functionRotate": {
             "exampleCode": "./live-examples/css-examples/transforms/function-rotate.html",
             "fileName": "function-rotate.html",
             "title": "CSS Demo: rotate()",
             "type": "css"
         },
-        "rotate3d": {
+        "functionRotate3d": {
```
bsmth, line 35: I think the rename makes sense, but are there any side-effects of renaming these keys?

bsmth:
From the changes, we have the following 

__Additions:__

* `live-examples/css-examples/transforms/rotate.html`
* `live-examples/css-examples/transforms/scale.html`
* `live-examples/css-examples/transforms/translate.html`

__Updates:__
* `live-examples/css-examples/transforms/meta.json`

Could you describe a bit of the context of the changes in the meta.json file?

bsmth:
The new additions look really nice, well done. <strike>Is there anything we could do with the file and example names to disambiguate from 2d transforms?</strike>

In https://github.com/mdn/interactive-examples/pull/2275#discussion_r1041146449, these transforms will do 2 or 3D operations depending on the number of passed values.

live-examples/css-examples/transforms/meta.json
```diff
@@ -7,175 +7,196 @@
...
+            "exampleCode": "./live-examples/css-examples/transforms/transform-style.html",
+            "fileName": "transform-style.html",
+            "title": "CSS Demo: transform-style",
+            "type": "css"
+        },
+        "translate": {
+            "cssExampleSrc": "./live-examples/css-examples/transforms/transform3d.css",
+            "exampleCode": "./live-examples/css-examples/transforms/translate.html",
+            "fileName": "translate.html",
+            "title": "CSS Demo: translate",
```
bsmth, line 198: Does it make sense to give it the title `CSS Demo: 3D translate`? Same for the other ones.

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: No, those keys don't matter at all, as long as they are unique in the same file.

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: perspective and perspectiveOrigin did already exist, but they were at the beginning of the file. I moved them to keep alphabetical order, like in the rest of meta.json files.

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: In all CSS examples, the title contains just the name of a property or a function, so I think adding a "3D" prefix could be misleading.

live-examples/css-examples/transforms/meta.json
bsmth: Oh, I see, thanks.

live-examples/css-examples/transforms/meta.json
bsmth: That's true. What about in the file name? We have `transform3d` above, what do you think about `translate3d.html` or `translate-3d.html` or similar?

live-examples/css-examples/transforms/meta.json
bsmth: I had to check again how we're embedding in the content repo, for example for `border` CSS prop we're using:

```
{{EmbedInteractiveExample("pages/css/border.html")}}
```

My concern was how we reference them and if this renaming had any impact, so I guess this is fine :)

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: I don't quite understand. The property name is called [translate](https://developer.mozilla.org/en-US/docs/Web/CSS/translate), so we want to reference it by its name like that:
```
{{EmbedInteractiveExample("pages/css/translate.html")}}
```
while a function called [translate3d](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translate3d) is called like that:
```
{{EmbedInteractiveExample("pages/css/function-translate3d.html")}}
```
Those 3 new properties might be similar to 3d functions, but they have differences too. They all allow less than 3 values to be placed and the syntax of the scale property is totally different from scale3d.

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: Currently, the path is built by always placing "pages/" on the front, then the value of property "type" is included, then "/" and then the value of fileName property. So it also matters that file names don't collide in the same type.

live-examples/css-examples/transforms/meta.json
bsmth: Thanks, I had a read through the docs to get some more context:

>The [translate](https://w3c.github.io/csswg-drafts/css-transforms-2/#propdef-translate) property accepts 1-3 values, each specifying a translation against one axis, in the order X, Y, then Z. When the second or third values are missing, they default to 0px.
> If the third value is omitted or zero, this specifies a 2d translation, equivalent to the [translate()](https://w3c.github.io/csswg-drafts/css-transforms-1/#funcdef-transform-translate) function. Otherwise, this specifies a 3d translation, equivalent to the [translate3d()](https://w3c.github.io/csswg-drafts/css-transforms-2/#funcdef-translate3d) function.

I was confused because I thought the properties are for 3D transforms only due to the 3D box we're using.

live-examples/css-examples/transforms/translate.html
```diff
@@ -0,0 +1,42 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="translate">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">translate: 50%;</code></pre>
```
bsmth: I think it might be useful to show `none` also. And to demonstrate how it works with 1-3 values. How about:

* `translate: none;`
* `translate: 50%;` one value
* `translate: 50% 2rem;` two values
* `translate: 50% 2rem -20px;` three values
* ... (optional if we want to limit the height of the example)

What do you think?

live-examples/css-examples/transforms/meta.json
bsmth: FYI, I'm considering adding [something like this codepen](https://codepen.io/bsmth/pen/bGKzVzq) to the [examples section](https://developer.mozilla.org/en-US/docs/Web/CSS/translate#examples) as to show that you can do 2D transforms with regular elements without special handling.

live-examples/css-examples/transforms/translate.html
NiedziolkaMichal: That sounds good. I have updated example to match your suggestions.

NiedziolkaMichal: That sounds good. I have updated the example to match your suggestions.

NiedziolkaMichal: That sounds good. I have updated the example to match your suggestions.

live-examples/css-examples/transforms/meta.json
NiedziolkaMichal: I see, that's ok.

live-examples/css-examples/transforms/rotate.html
```diff
@@ -0,0 +1,49 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="rotate">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">rotate: -45deg;</code></pre>
```
bsmth, line 10: I would again recommend adding `none` before the first value so we can see it unset.

live-examples/css-examples/transforms/scale.html
```diff
@@ -0,0 +1,42 @@
+<section id="example-choice-list" class="example-choice-list large" data-property="scale">
+    <div class="example-choice" initial-choice="true">
+        <pre><code class="language-css">scale: 1.5;</code></pre>
```
bsmth, line 10: Same comment about adding `none` as a value

live-examples/css-examples/transforms/scale.html
```diff
@@ -0,0 +1,42 @@
...
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
+        <pre><code class="language-css">scale: 1.7 0.5;</code></pre>
```
bsmth: Maybe we can add a percentage value for illustration:

```suggestion
        <pre><code class="language-css">scale: 1.7 60%;</code></pre>
```

bsmth (approved):

bsmth:
Thanks @NiedziolkaMichal. I've committed some changes in 3a0a790a0a80239a0f7d13972b1065fda11c1299 do you agree? If not, feel free to revert the commit and we can have another look, but this is ready to go in my opinion 👍🏻

NiedziolkaMichal:
Yes, that looks great. Sorry for the delay.

bsmth:
Thanks, merging now

PR #2277, opened 2022-09-22
Improve existing example checkbox.html

valentin-p:
Improve an existing example.

I am fixing the inline html markup for "checked" in the checkbox example.

I noticed the issue in this site.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

Before
image

After
image

Have a nice day.


NiedziolkaMichal:
Thanks a lot, that's a good change.

PR #2285, opened 2022-09-27
fix: malformed code flagged by Prettier

nschonni:
Ran prettier across the repo, and these were a few of the files it choked on because of parsing issues in the code.
There is also an issue with the rtc.html. That one changed upstream because the tab no longer is specified https://github.com/mdn/content/pull/20869


wbamberg (approved):
👍 nice, thank you @nschonni .

PR #2287, opened 2022-09-28
ci: use built-in Yarn caching

nschonni:
Some jobs were using the built-in caching and custom, so changed them all to use the setup-node caching alone


schalkneethling (approved):
Thanks so much, @nschonni 🎉

PR #2288, opened 2022-09-28
style: Enable ESLint and Prettier for all files

nschonni:
This is a little big, so I can try and break it down a bit, if the end result is desired:
- Remove ESLint rules that overlap with Prettier/EditorConfig
- Remove Prettier rules overlapping with EditorConfig
- Replace Prettier-eslint with plugin/config versions
- Set Environment to ES2022 and remove globals that are not required
- Remove babel parser, since covered by ES2022
- Ignore WASM folders because of top level awaits
- Rename script commands to match mdn/content


NiedziolkaMichal:
I can see few problems with this commit:

- Unfortunately it's going to change height of multiple examples. For example [Array.reduceRight](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduceRight) will be converted from `smaller` to `medium`. Basic solution is to make separate PR in content repository, in which we revisit specified heights of all interactive examples modified in this PR. Alternative solution is to wait for [#839](https://github.com/mdn/bob/pull/839) &  [#7679](https://github.com/mdn/yari/pull/7679) to be deployed, which will make it no longer necessary to update content repo.

- Comments in examples of bitwise operators are no longer aligned. For example take a look at [&=](https://github.com/mdn/interactive-examples/pull/2288/files#diff-31598641d5185066ca77a7e8c6771dc5f60f37a874d64de8919c4f8adee7d6b1). Currently it's very easy to see how individual bits are affected, which makes those operators easier to be understood.

- Forced line breaks in arrays will make some examples huge. For example take a look at [extends](https://github.com/mdn/interactive-examples/pull/2288/files#diff-84674961bf986583e845e31f6ec69a58501a39c8a56d4044e36825b98ba6a915) or [NumberFormat](https://github.com/mdn/interactive-examples/pull/2288/files#diff-04a1fe27ed5fde264c5ac3ad64f2bcb173a10e104b7a1fd550879b1103000071).

- I feel code style would be better without line wrapping like in examples [TypedArray.at](https://github.com/mdn/interactive-examples/pull/2288/files#diff-f6209e5badc0cdbaa35139429d54d42508d6849e9ff9d2602f8a683cbeebc808) or [Symbol.split](https://github.com/mdn/interactive-examples/pull/2288/files#diff-8abb5382c2ebce7b045069761ee0dc10110250004f10811a1375d5f8e23c92d8), but that's just my opinion.

- I feel unnecessary double parenthesis in examples like [exponentiation assignment](https://github.com/mdn/interactive-examples/pull/2288/files#diff-f564d06fd0de4a4bd29ecd6734fe4b525176eeeb0361f7433c0ee6000ce70b28) might be confusing to the user.

Except for that I can see clear benefits of this PR. Most of the examples would look better with those changes applied.

schalkneethling:
Hey @nschonni, thoughts on the comments here https://github.com/mdn/interactive-examples/pull/2288#issuecomment-1328350043

schalkneethling:
Hey @nschonni and @NiedziolkaMichal,

First, thank you for the work you have done here, Nick. There were a couple of items brought up by Michal here https://github.com/mdn/interactive-examples/pull/2288#issuecomment-1328350043

Please review them when you have a moment. Also, there are a couple of conflicts at the moment, mainly in `package.json`. Thanks so much.

nschonni:
The config does have overrides for particular files https://github.com/mdn/interactive-examples/blob/2315bc6a8f9e8f0dc09cff73c88b2d85f7854523/.eslintrc#L71
If you've got suggestions for those, then I can add them, but I didn't see anything actionable for me from the comments above

schalkneethling:
Hey @nschonni and @NiedziolkaMichal,

We had a chat about this one. The work here is great and something we want but, the following topic raised by Michal is a blocker at the moment.

> Unfortunately it's going to change height of multiple examples. 

We have brought this to the attention of engineering and will follow up on this in the new year. We are also discussing whether this can be something the community can assist with if we have low-bandwidth internally.

We hope this makes sense and is acceptable to you all. Thanks again for the work done.

schalkneethling:
@NiedziolkaMichal, @bsmth, anyone opposed to us landing this one?

bsmth:
I think the changes are good, but it might be best if these two PRs land first so we don't introduce display issues:

- [x] https://github.com/mdn/bob/pull/839
- [ ] https://github.com/mdn/yari/pull/7679

schalkneethling:
> I think the changes are good, but it might be best if these two PRs land first so we don't introduce display issues:
> 
> * [x]  [Adds height-data.json bob#839](https://github.com/mdn/bob/pull/839)
> * [ ]  [feat(macros/EmbedInteractiveExample): use height from height-data.json yari#7679](https://github.com/mdn/yari/pull/7679)

I have assigned reviewers for these pull requests.

PR #2289, opened 2022-09-28
ci: remove Jenkinsfile and build scripts

nschonni:
Seems like everything is GitHub Actions now


nschonni:
@wbamberg is this setup still being used?

wbamberg:
I think we can remove it but I'm not sure. @schalkneethling , @NiedziolkaMichal  do you know if we can remove this Jenkins stuff?

NiedziolkaMichal:
I think it's up to @schalkneethling. Maybe he prefers to keep it as a backup or there is a secret usage of this build that we are not aware of.

bsmth:
I'll raise this one with the team today

schalkneethling:
Hey @nschonni, @wbamberg, @NiedziolkaMichal, @bsmth, I do think we still need these. I say that with two caveats:

1. I am not sure we still need `build.sh` and `slack-notify.sh` - The first is replaced by a GitHub Action, and the second is a nice to have. I would vote for keeping the Slack notify as Jenkins does not report back on GitHub.
2. What we definitely still need is `s3-sync.sh`. Without this, the built files will never end up in production. There is surely an Action we can use to do the Slack notify and the S3 sync, but for now it might be safer to keep these.

The one thing I am a little curious about is what is kicking of the Jenkinsfile. I will have to ask our SRE's about that. Seeing that filies do end up in production and we are notified on Slack, it is definitely running. My first thought was a webhook, but there is no webhook in this repository although, there might be one on Jenkins?

nschonni:
I'm seeing 2 jobs that maybe are supposed to replace that (but only push to an orphan branch right now)
- https://github.com/mdn/interactive-examples/blob/main/.github/workflows/prod-deploy.yml
- https://github.com/mdn/interactive-examples/blob/main/.github/workflows/publish-release.yml

schalkneethling:
Hey @nschonni and @bsmth, I have asked our site reliability engineers (SRE) to have a look at this PR and clear up any confusions. Once we know the answer, we should 100% document it. :+1:

schalkneethling:
@nschonni, I spoke with the SREs, and we still need these files. I will summarize the feedback and add it to the README.

PR #2291, opened 2022-09-28
hyphenate-character

NiedziolkaMichal:
This is interactive example for [hyphenate-character](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphenate-character) CSS property.

image


wbamberg (approved):

PR #2293, opened 2022-09-29
fix: Improve array.keys

elo1lson:
* use let declaration
* add more exemple


wbamberg (commented):
Thanks for the PR, @elo1lson ! But I don't think this is an improvement.

- `const` is better than `let` here
- using an iterator like this is not going to be as clear or as common a use case as using `for...of`
- this example exhausts the iterator, so the `for...of` loop does nothing

PR #2299, opened 2022-10-06
Add CSS import at-rule

NiedziolkaMichal:
This is interactive example for [@import](https://developer.mozilla.org/en-US/docs/Web/CSS/@import).

image


bsmth:
Thanks @NiedziolkaMichal. Usually we're expecting the reader to be able to click through a few options and see the result, but I'm not sure we're adding anything to the article with this. Do you think `@import` deserves an interactive example? I would lean towards leaving it out.

NiedziolkaMichal:
I think you are right. What if I would add media query or `supports` to the example, so more functionalities are displayed? Do you think it would make interactive example useful to the user?

bsmth:
> What if I would add media query or `supports`

We could do, if we're doing something with [conditionals](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#importing_css_rules_conditionally), supports or layers it might be interesting. Is there anything that's [in the examples section](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#examples) that would be better demonstrated by user interaction? I'm not sure 🤔

bsmth:
There are some good options we can include from the examples like:


```css
@import url("whatever.css") supports(font-tech(color-COLRv1));

@import "theme.css" layer(utilities);

```

bsmth:
> Usually we're expecting the reader to be able to click through a few options and see the result, but I'm not sure we're adding anything to the article with this. Do you think `@import` deserves an interactive example? I would lean towards leaving it out.

I'm going to close this one for now with this in mind unless there's something I'm missing that's really useful with interactivity in mind. Thank you :)

PR #2305, opened 2022-10-10
Fix JavaScript sample for wat/call

dsyer:
The `meta.json` is pointing at the wrong JS file.


NiedziolkaMichal:
Thank you for this fix

PR #2310, opened 2022-10-13
Attribute size

NiedziolkaMichal:
This PR adds interactive example for attribute [size](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size).

image

### DIFF
--- a/live-examples/html-examples/form-attributes/attribute-size.html
+++ b/live-examples/html-examples/form-attributes/attribute-size.html
@@ -1,0 +1,17 @@
+<label for="firstName">First Name:</label>
+<input name="firstName"
+       type="text"
+       size="10">
+
+<label for="lastName">Last Name:</label>
+<input name="lastName"
+       type="text"
+       size="20">
+
+<label for="shirtSize">Size of T-Shirt:</label>
+<select name="shirtSize"
+        size="2">
+	<option>Small</option>
+	<option>Medium</option>
+	<option>Large</option>
+</select>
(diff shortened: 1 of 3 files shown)


Rumyra:
Hey @NiedziolkaMichal ! Apologies if I'm late to the party on this - but I assume anything that does *not* have the label `depends on shadow dom fix` can be reviewed & merged?

NiedziolkaMichal:
@Rumyra Yes. I just need to make changes to imputmode example, but except for that, everything should work fine in production.

Rumyra:
Thanks @NiedziolkaMichal - I'll find sometime this week to check these out and get them merged 👍

NiedziolkaMichal:
That would be great, thank you

Rumyra:
@NiedziolkaMichal I've done a first pass

For this one I wonder whether 'tshirt size' might confuse users as we're showing the `size` attribute - maybe something like 'Favourite fruit' (Orange, banana, apple) means the content doesn't conflict with the code (if that makes sense)

NiedziolkaMichal:
@Rumyra I think you are right. I replaced "Size of T-Shirt" with "Faviourite fruit" as you have suggested.

Rumyra (approved):

Rumyra:
Amazing - thanks so much @NiedziolkaMichal !

NiedziolkaMichal:
Thanks a lot for all reviews. Those PRs have been waiting for a while.

PR #2314, opened 2022-10-17
feat(html): Add `crossorigin` attribute

NiedziolkaMichal:
This PR adds interactive example for [crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) attribute. I chose to show how it works with script element, because showing how tainted canvas work would require long and complex JS code, while crossorigin in `<link>` effect can be seen mostly in network devtools tab.

**Currently my private host is used to store cross-origin JS code, so this PR shouldn't be merged.** I won't be able to store it indefinitely, so we need server owned by Mozilla with `Access-Control-Allow-Origin:  *` HTTP header which can store following JS file:
```js
setTimeout(() => {
    throw new Error('Secret message of error inside cross-origin script.');
}, 500);
```

image

### DIFF
--- a/live-examples/html-examples/other-attributes/attribute-crossorigin.html
+++ b/live-examples/html-examples/other-attributes/attribute-crossorigin.html
@@ -1,0 +1,6 @@
+<script src="http://ddvdiow.cluster031.hosting.ovh.net/throwError.js"></script>
+
+<script src="http://ddvdiow.cluster031.hosting.ovh.net/throwError.js" crossorigin></script>
+
+<label for="logBox">Caught Errors:</label>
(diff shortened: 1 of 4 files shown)


bsmth:
I think it's a good addition. It would be great to get some feedback or thoughts from the team as to how we host it, adding `needs infra` as we might need some additional setup here. Tagging @LeoMcA for some input

bsmth (changes requested):
We need to resolve where crossorigin resources are hosted

bsmth:
I'm going to close for now due to inactivity. We may reopen again at some point if supporting infra is available.

PR #2316, opened 2022-10-18
feat(css): Add example of `:modal`

NiedziolkaMichal:
This PR adds interactive example to [:modal](https://developer.mozilla.org/en-US/docs/Web/CSS/:modal).

image

### DIFF
--- a/live-examples/css-examples/pseudo-class/modal.html
+++ b/live-examples/css-examples/pseudo-class/modal.html
@@ -1,4 +1,4 @@
-<p>Would you like to see new random number?</p>
+<p>Would you like to see a new random number?</p>
 <button id="showNumber">Show me</button>
 
 <dialog id="favDialog">
--- a/live-examples/css-examples/pseudo-class/modal.html
+++ b/live-examples/css-examples/pseudo-class/modal.html
@@ -1,0 +1,9 @@
+<p>Would you like to see new random number?</p>
+<button id="showNumber">Show me</button>
+
+<dialog id="favDialog">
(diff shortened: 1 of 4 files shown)


live-examples/css-examples/pseudo-class/modal.html
```diff
@@ -0,0 +1,9 @@
+<p>Would you like to see new random number?</p>
```
sideshowbarker:
```suggestion
<p>Would you like to see a new random number?</p>
```

sideshowbarker (approved):

sideshowbarker:
At https://github.com/mdn/interactive-examples/actions/runs/5632002837/job/15259572637?pr=2316#step:6:11 there is a CI failure:

>   8:1  ✖  Unexpected unknown pseudo-class selector ":modal"  selector-pseudo-class-no-unknown

…but the reason for that is, https://github.com/mdn/interactive-examples/blob/1d69481229781284b699d4e7ea118b8dae218939/package.json#L50C24-L50C31 shows that we may be using an old version of `stylelint` — version [14.16.1 from Dec 29, 2022](https://github.com/stylelint/stylelint/releases/tag/14.16.1) — and we instead need to be using at least version [15.6.1 from May 3, 2023](https://github.com/stylelint/stylelint/releases/tag/15.6.1), which contains this fix:

> Fixed: selector-pseudo-class-no-unknown false positive for :modal ([#6811](https://github.com/stylelint/stylelint/pull/6811))

PR #2324, opened 2022-11-07
Add JS Object.is()

NiedziolkaMichal:
Interactive example for [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).

image

### DIFF
--- a/live-examples/js-examples/object/meta.json
+++ b/live-examples/js-examples/object/meta.json
@@ -72,6 +72,12 @@
             "title": "JavaScript Demo: Object.getPrototypeOf()",
             "type": "js"
         },
+        "objectIs": {
+            "exampleCode": "./live-examples/js-examples/object/object-is.js",
+            "fileName": "object-is.html",
+            "title": "JavaScript Demo: Object.is()",
+            "type": "js"
+        },
         "objectHasOwn": {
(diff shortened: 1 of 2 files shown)


Josh-Cena (commented):
LGTM, but maybe add two object examples?

```js
console.log(Object.is({}, {}));
const obj = {};
const obj2 = obj;
console.log(Object.is(obj, obj2));
```

NiedziolkaMichal:
I am afraid it will cause the big version of JS example window to be displayed. Additionally, the output will not fit inside the console so scroll bar will appear:
image

Josh-Cena:
Okay; maybe get rid of the last one and keep `Object.is({}, {})`? I mostly want to tell people that it's still "compare-by-reference".

NiedziolkaMichal:
That's a good idea. I have changed the example a little bit, so it shows that:

- It doesn't do type coercion
- Treats NaN equal, unlike ===
- Treats -0 & 0 not equal, unlike ===
- Compares objects by reference, instead of shallow equal or sth else

Adding 1 more line would change the window height to tall.

image

Josh-Cena (approved):
LGTM, thanks!

PR #2329, opened 2022-11-10
Array.isArray()

NiedziolkaMichal:
Interactive example for [Array.isArray()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

image

### DIFF
--- a/live-examples/js-examples/array/array-isarray.js
+++ b/live-examples/js-examples/array/array-isarray.js
@@ -1,0 +1,11 @@
+console.log(Array.isArray([1, 3, 5]));
+// expected output: true
+
+console.log(Array.isArray('[]'));
+// expected output: false
+
+console.log(Array.isArray(new Array(5)));
+// expected output: true
+
+console.log(Array.isArray(new Int16Array([15, 33])));
+// expected output: false
--- a/live-examples/js-examples/array/meta.json
+++ b/live-examples/js-examples/array/meta.json
@@ -108,6 +108,12 @@
             "title": "JavaScript Demo: Array.indexOf()",
             "type": "js"
         },
+        "arrayIsArray": {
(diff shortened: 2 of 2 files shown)


bsmth:
Thanks @NiedziolkaMichal , it looks like there's no reviewers automatically assigned for these PRs. Do you need eyes on these additions? Should we set up round-robin reviews?

NiedziolkaMichal:
If you wish, you can freely review those PRs. I would appreciate that.

live-examples/js-examples/array/array-isarray.js
```diff
@@ -0,0 +1,11 @@
+console.log(Array.isArray([1, 3, 5]));
+// expected output: true
+
+console.log(Array.isArray('[]'));
+// expected output: false
+
```
bsmth, line 6: Should we add one or two more with different types? I would like to see some strings in with numeric types. We shouldn't cover every case, though. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#examples

bsmth (approved):
Looks good, see my comment about adding some more types (string, boolean, object maybe?).

live-examples/js-examples/array/array-isarray.js
NiedziolkaMichal: In case I add 1 more example, size of the widget would change to big and it would take up most of the viewport like this:

image

Don't you think it's an overkill considering how simple function `isArray` is? Even `Array.reduce` uses standard size.

bsmth: > size of the widget would change to big

Oh I see. That's overkill, I agree. Looks good to me, then 👍🏻

NiedziolkaMichal: Alright, thank you

PR #2332, opened 2022-11-16
Add timezone to create the same output for all users.

lemnis:
Currently  the output doesn't match the expected output when a user is in a different timezone, e.g. for me it outputs

> "Sunday, 20 December 2020 at 04:23:16 CET"


NiedziolkaMichal:
That looks good, thank you for this fix.

PR #2334, opened 2022-11-17
Add JS Object.setPrototypeOf

NiedziolkaMichal:
Interactive example for [Object.setPrototypeOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf).

image

### DIFF
--- a/live-examples/js-examples/object/meta.json
+++ b/live-examples/js-examples/object/meta.json
@@ -150,6 +150,12 @@
             "title": "JavaScript Demo: Object.seal()",
             "type": "js"
         },
+        "objectSetPrototypeOf": {
+            "exampleCode": "./live-examples/js-examples/object/object-setprototypeof.js",
+            "fileName": "object-setprototypeof.html",
+            "title": "JavaScript Demo: Object.setPrototypeOf()",
+            "type": "js"
+        },
         "objectValues": {
             "exampleCode": "./live-examples/js-examples/object/object-values.js",
             "fileName": "object-values.html",
--- a/live-examples/js-examples/object/object-setprototypeof.js
+++ b/live-examples/js-examples/object/object-setprototypeof.js
@@ -1,0 +1,12 @@
+function Foo() {}
+function Bar() {}
+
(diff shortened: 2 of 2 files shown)


Josh-Cena (commented):
I would suggest not including `new` on pages about prototypes; it unnecessarily complicates the matter. Especially when the property in discussion is not a property that everyone knows to live on `Bar.prototype`. I suggest creating prototype chains with plain objects, like:

```js
const parent1 = { inherited: "from parent1" };
const parent2 = { inherited: "from parent2" };

const obj = { __proto__: parent1 };
console.log(obj.inherited);
Object.setPrototypeOf(obj, parent2);
console.log(obj.inherited);
```

That's just my opinion, though; feel free to disagree.

NiedziolkaMichal:
@Josh-Cena I don't think the usage of `__proto__` in here is proper, because its deprecated, and many readers might not be familiar with its effect. How about this?
```js
const obj = {};
const parent = { foo: 'bar' }

console.log(obj.foo);
// expected output: undefined

Object.setPrototypeOf(obj, parent);

console.log(obj.foo);
// expected output: "bar"
```

Josh-Cena:
`__proto__` in object literals is not deprecated; it as a _getter_ property is deprecated. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#prototype_setter. And it's what we consistently use in prototype-related tutorials, so IMO there won't be much problem. Your proposed code also looks good to me.

NiedziolkaMichal:
I wasn't aware that using __proto__ inside an object literal is a good practice, it's good to know. I still think my example is simpler to understand so I committed it. If you feel any more improvements could be made or you wish your proposal should be committed, that's ok with me.

Josh-Cena (approved):
LGTM. Thanks!

PR #2336, opened 2022-11-18
Add example for JavaScript 'super' keyword

NiedziolkaMichal:
Interactive example for keyword [super](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super). It shows how super can be used in constructor and also to access methods from superclass. I hope it's not too complex.

image

### DIFF
--- a/live-examples/js-examples/expressions/expressions-super.js
+++ b/live-examples/js-examples/expressions/expressions-super.js
@@ -1,0 +1,28 @@
+class Foo {
+  constructor(name) {
+    this.name = name;
+  }
+
+  getNameSeparator() {
+    return '-';
+  }
+}
+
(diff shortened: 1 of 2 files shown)


bsmth:
Hi @NiedziolkaMichal thanks a lot! 🙌🏻  Side note: the title is really useful to get an idea of the PR contents. I've updated this one so the reviewer can see from the open PRs what the contents are from a format like this:

```
[Added/Edited/Removed] the [technology type] [specific property]
```

NiedziolkaMichal:
@bsmth Thank you

bsmth (approved):
LGTM 👍🏻

NiedziolkaMichal:
Thanks for review

PR #2341, opened 2022-11-25
Add JS GeneratorFunction examples

NiedziolkaMichal:
Interactive examples for [GeneratorFunction](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/GeneratorFunction) and [AsyncGeneratorFunction](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGeneratorFunction).

image

image

### DIFF
--- a/live-examples/js-examples/functionasterisk/async-functionasterisk-function.js
+++ b/live-examples/js-examples/functionasterisk/async-functionasterisk-function.js
@@ -1,4 +1,4 @@
-const AsyncGeneratorFunction = (async function* () {}).constructor;
+const AsyncGeneratorFunction = async function* () {}.constructor;
 
 const foo = new AsyncGeneratorFunction(`
   yield await Promise.resolve('a');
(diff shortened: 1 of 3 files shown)


live-examples/js-examples/functionasterisk/async-functionasterisk-function.js
```diff
@@ -0,0 +1,19 @@
+const AsyncGeneratorFunction = (async function* () {}).constructor;
```
Josh-Cena:
```suggestion
const AsyncGeneratorFunction = async function* () {}.constructor;
```

Prettier-style

Josh-Cena (approved):
LGTM; you may want to format it so the style is aligned with content (e.g. also use double quotes). These pages are very obscure so I won't worry much. We recently split the constructor pages from the main class pages—I'm almost sure it doesn't matter for interactive examples, but just as a heads-up

live-examples/js-examples/functionasterisk/async-functionasterisk-function.js
NiedziolkaMichal: Thank you, that looks better

bsmth:
Thanks, both, merging shortly 👍🏻

PR #2343, opened 2022-11-27
Add JS example: Symbol.asyncIterator

NiedziolkaMichal:
Interactive example for [Symbol.asyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator).

image


Josh-Cena (approved):
That's a great example!

PR #2344, opened 2022-11-27
fix(ci): Fix start-watch command

NiedziolkaMichal:
This PR is meant to fix `start-watch` command, which builds all interactive examples and then automatically triggers next build when any file has changed.
I have simply replaced all `'` with `"`. Command with `'` doesn't produce any output on my device(Windows 11). It would be great if someone could test this PR on linux.


package.json
```diff
@@ -21,7 +21,7 @@
         "lint:css": "stylelint \"**/*.css\"",
         "lint:js": "eslint ./live-examples/js-examples",
         "start-server": "http-server -p 9090 ./docs",
-        "start-watch": "chokidar '**/{*.js,*.css,*.html,*.json}' -i 'package.json' -i 'docs/**' -i 'node_modules/**' -i 'js/editor-*.js' -c 'npm run build:pages' --initial --silent",
+        "start-watch": "chokidar \"**/{*.js,*.css,*.html,*.json}\" -i \"package.json\" -i \"docs/**\" -i \"node_modules/**\" -i \"js/editor-*.js\" -c \"npm run build:pages\" --initial --silent",
```
bsmth: It feels like this should be inside a `./scripts/` dir because we're doing too much to get this working in the `package.json` but maybe that's good for a follow-up.

bsmth (approved):
This works fine on macOS, yet to try on a linux box

package.json
NiedziolkaMichal: Actually, how about changing this script to: `chokidar "./live-examples/**/{*.js,*.css,*.html,*.json}" -c "npm run build" --initial --silent`?
Instead of excluding a few files, we could just look for changes in `live-examples`.

NiedziolkaMichal: We could also do that: `chokidar "./live-examples/**" -c "npm run build" --initial --silent`.

bsmth: > We could also do that: `chokidar "./live-examples/**" -c "npm run build" --initial --silent`.

This looks a lot better

NiedziolkaMichal: Great. I have changed the script accordingly.

bsmth (approved):
Looks good to me 👍🏻

NiedziolkaMichal:
@bsmth Thanks for the review.

PR #2345, opened 2022-11-28
Add JS example: Symbol.prototype.valueOf()

NiedziolkaMichal:
Interactive example for [Symbol.prototype.valueOf()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/valueOf).

image


Josh-Cena (approved):
LGTM

bsmth:
Thanks, both. I'm making notes to add the corresponding content additions to include these.

Issue #2306, opened 2022-10-10
[Question] Add interactive example for web API(Streams API)?


zizifn:
I want to add an interactive example for https://developer.mozilla.org/en-US/docs/Web/API/Streams_API. 

I noticed there is a [`type: webapi-tabbed` ](https://github.com/mdn/interactive-examples/blob/main/live-examples/webapi-examples/document/meta.json#L10), but this is not in the `CONTRIBUTING` doc, I tested this in my local, this type is working fine.  
And the only example inside folder `live-examples/webapi-examples` is also working fine in PROD env https://interactive-examples.mdn.mozilla.net/pages/webapi-tabbed/document-queryselector.html

So is ok for me to add [Streams_API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) interactive example inside folder `live-examples/webapi-examples`?

NiedziolkaMichal:
Lately application used for displaying interactive examples is undergoing some changes. Current version displays output in Shadow DOM, because of which simple DOM queries like `document.getElementById` doesn't work and code needs to refer to Shadow Root instead:
```js
    let documentRoot = getShadowRoot();
    return documentRoot.querySelector('header');
```

This is going to change in new version of [BOB](https://github.com/mdn/bob) and linked example will need to be rewritten. Most likely new version will be released when all issues from [#881](https://github.com/mdn/bob/issues/881) are resolved. After that, we will be able to publish Web API examples. I can leave this issue open and ping you when that happens.

PR #2347, opened 2022-11-30
Add an example for Navigator.userAgent

MrBrain295:


MrBrain295:
@NiedziolkaMichal

NiedziolkaMichal:
Thank you for the PR. Unfortunately `webapi-tabbed` type is not yet ready to be used in production. Please refer to mdn/interactive-examples#2306 and [#15166](https://github.com/mdn/content/pull/15166) for an explanation.

MrBrain295:
Hmmm, should I close this or mark it as a draft?

NiedziolkaMichal:
Draft will do fine. I will notify you when something changes.

MrBrain295:
@NiedziolkaMichal I don’t know what to do about meta.json but from what I saw in mdn/mdn-community#305 this should be good now.

NiedziolkaMichal:
Since your example requires just a JS tab, you can use the JS editor. Your `meta.json` content should be something like that:
```json
        "navigatorPrototypeUserAgent": {
            "exampleCode": "./live-examples/webapi-examples/navigator/navigator-prototype-useragent.js",
            "fileName": "navigator-prototype-useragent.html",
            "title": "Web API Demo: Navigator.prototype.userAgent",
            "type": "js"
        }
```
You would need to move and rename `js` file to fit `exampleCode` value. Still, there are a few more problems with it:
- JS editor breaks when output is too long. I think it is fixed in the next BOB version, so we would need to wait for it to be deployed.
image
- Pretty much every usage of `console.log` in JS examples, has comment about expected output directly below. For example `// expected output: [6, 6, 6, 6]`. Here output depends on the browser of a user, so maybe `example output` would be more fitting. Still, any output is pretty big and it would show a scrollbar in the editor:
image
- You might have read that @wbamberg doesn't see `Navigator.userAgent` as a good candidate for an interactive example. You would need to convince him otherwise.

wbamberg:
> You might have read that @wbamberg doesn't see Navigator.userAgent as a good candidate for an interactive example. You would need to convince him otherwise.

The point about interactive examples is that they are interactive: the user can edit them and see something different. So for example [the one for `slice()` is nice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice): it's actually quite a fiddly API, and the example gives you a way to try it out with different parameters and see the effect. If a user is actually trying to figure out how to use it with a particular array they can paste that in too and try out different params until they get the answer they want.

But this example isn't interactive - there's nothing you can do with the editor: all you can do is press "Run". You could implement this as a normal live sample. And all it will tell you is your browser's UA string. It would be much more helpful to users to list several common UA strings.

When we started the interactive examples project we were conscious of the fact that they are quite rude: they occupy quite a lot of the most precious space on the page. So we always felt that there would be features where it wasn't appropriate to add examples, because they didn't add enough value to justify their existence - and I would say this is one of these cases. So I think it's only worth adding examples like this if we want to change that policy, and instead want to add examples for just about everything.

PR #2351, opened 2022-12-05
add hyphenate-limit-chars example

chrisdavidmills:
As part of my work on the [CSS `hyphenate-limit-chars` property](https://docs.google.com/document/d/1aLdSuLdZEMjdnMpeyBRdwwZyLr1JMbQ39Nk0ZWumLBU/edit#) (added to Chrome in version 109), I have added an interactive example for it.


chrisdavidmills:
Hi there @NiedziolkaMichal ! Are you able to review this PR this week? If not, no problem. Let me know and I can try to find someone else.

NiedziolkaMichal:
Hey! Unfortunately, it's not supported on Chrome and Firefox yet, so it cannot be published. There is a new version of BOB(component showing the examples) that provides an informative fallback when an example is not supported on user's browser. When it gets into production and a stable version of Chrome will support this feature, I will be very happy to review and merge `hyphenate-limit-chars`.

chrisdavidmills:
@NiedziolkaMichal It does work in Chrome Canary, but not Chrome Stable as yet. Is the policy to not publish examples until they work in a stable browser version? If so, that sounds reasonable to me.

NiedziolkaMichal:
Yeah, for a while now we try to publish only those examples which work well on firefox and chrome. Currently, when feature is not supported by the users browser, he sees something like that:
image

The newer version is more informative, so I think it will be ok to publish also those examples, which are supported only by a single browser:
image

chrisdavidmills:
Yup, that definitely looks better.

NiedziolkaMichal:
I will ping you when the new BOB gets deployed, so we can merge this example.

queengooborg (approved):
It looks like this is only blocked on a new BoB release -- of which has landed since the last comment.  As such, I'm approving and merging this!

PR #2352, opened 2022-12-05
Add an example for `autocapitalize`

MrBrain295:


NiedziolkaMichal:
Unfortunately, it is not supported on Firefox yet. The current version of the editor doesn't inform users about that in any way, so it might be the cause of confusion.
Currently, we ship examples that are supported on Firefox and Chrome.

bsmth:
Thanks @NiedziolkaMichal - going to close it for the moment. Please reopen if you think this is a mistake 👍🏻

PR #2357, opened 2022-12-10
chore: remove direct ajv dep

nschonni:
I included this in the ESLint PR, but I thought I'd split it out. AJV can be used to validate JSON files, but it doesn't appear to be directly used in this project right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "devDependencies": {
         "@babel/core": "^7.20.5",
         "@babel/eslint-parser": "^7.19.1",
-        "ajv": "^8.11.2",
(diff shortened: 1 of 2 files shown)


bsmth:
Did some git archaeology and it looks like it was added in https://github.com/mdn/interactive-examples/pull/1142. No objections to removing if we don't use it. Tagging @schalkneethling for a heads up.

bsmth (approved):
Going to approve, but let's wait for Schalk to chime in before merging

schalkneethling (approved):
Thanks, @nschonni 🌮

PR #2358, opened 2022-12-14
Add an example for :root

MrBrain295:
This one looks like it should be valid its supported in Chrome and Firefox and its a pseudo-class not a Web API.

### DIFF
--- a/live-examples/css-examples/pseudo-class/meta.json
+++ b/live-examples/css-examples/pseudo-class/meta.json
@@ -306,6 +306,15 @@
             "defaultTab": "css",
             "height": "tabbed-standard"
         },
+        "root": {
+            "cssExampleSrc": "./live-examples/css-examples/pseudo-class/root.css",
+            "exampleCode": "./live-examples/css-examples/pseudo-class/root.html",
+            "fileName": "pseudo-class-root.html",
+            "title": "CSS Demo: :root",
+            "type": "tabbed",
+            "defaultTab": "css",
+            "height": "tabbed-shorter"
+        },
         "valid": {
             "cssExampleSrc": "./live-examples/css-examples/pseudo-class/valid.css",
             "exampleCode": "./live-examples/css-examples/pseudo-class/valid.html",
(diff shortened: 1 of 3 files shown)


NiedziolkaMichal:
Unfortunately, your example doesn't work, at least in the newest version of BOB.
`html` and `body` elements that you add here are generally ignored by the browser because they already exist in the iframe that shows the example. That's generally not a problem and I guess it can stay.
The problem is that `font-family` property that you apply, is overridden by styles applied by BOB. It adds the following declarations to the body:
```css
    background-color: var(--background-primary);
    color: var(--text-primary);
    font: var(--type-body-l);
    height: 100%;
    margin: 0;
    padding: 0;
```
Additionally, those declarations are added in the `div` that contains the actual output:
```css
    color: #15141a;
    font-size: .9rem;
    height: min-content;
    line-height: 1.5;
    overflow: auto;
    padding: 2rem 1rem 1rem;
    width: fit-content;
```

MrBrain295:
So should I close this pr?

NiedziolkaMichal:
How about creating a custom variable inside :root and then using it in a paragraph or some other element? That's a common usage of this pseudo-class.

MrBrain295:
@NiedziolkaMichal I made some changes and I think its good now.

NiedziolkaMichal:
`#917413` doesn't meet the proper contrast requirements. I have changed it to `rebeccapurple` and added one empty line, is that ok?

MrBrain295:
Yep

NiedziolkaMichal:
Great, thank you for this example. It looks good.

bsmth:
It might be good to label these if we need a content addition. I've added `needs content update` for this so we don't forget closed PRs that add something new. We can remove the label after it's landed in `mdn/content` for at least `en-us`.

NiedziolkaMichal:
It would be hard to keep track of those, as typically individual contributors are creating a second PR in which an example gets published, and we don't get informed about it. Currently, when I see that quite a few examples were merged in here, I go through [commits](https://github.com/mdn/interactive-examples/commits/main) and look for all unpublished examples and create one big PR. The last time I did that, was Nov 4, so currently there are just 9 examples to add to the content repo.

bsmth:
I understand, maybe this helps in the meantime? https://github.com/mdn/interactive-examples/pulls?q=is%3Apr+label%3A%22needs+content+update%22+is%3Aclosed

PR #2359, opened 2022-12-14
Update interactive examples for comparison operators

yurj26:


NiedziolkaMichal:
That doesn't seem like a good change. The [strict inequality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality) page shows the behavior of `!==`, so what's the reasoning of changing this example?

yurj26:
see https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Reference/Operators/Strict_inequality

NiedziolkaMichal:
It's a problem with the content repository. Could you go to [strict_inequality](https://github.com/mdn/translated-content/edit/main/files/zh-cn/web/javascript/reference/operators/strict_inequality/index.md) and replace `{{EmbedInteractiveExample("pages/js/expressions-strict-equality.html")}}` with `{{EmbedInteractiveExample("pages/js/expressions-strict-inequality.html")}}`?

yurj26:
ok, thanks, I misread it.

NiedziolkaMichal:
That's ok, I will close this one.

PR #2363, opened 2022-12-15
Refreshing CONTRIBUTING page

### LINKED ISSUE #2360: CONTRIBUTING.md is outdated
From a quick glance I can see a couple issues. One part references adding an example in HTML and it’s all markdown now, two is that it like to issues with the example needed label but the last one of those was made in 2018.

NiedziolkaMichal:
Removed unnecessary html tags, in the example, showing how to publish an interactive example. I have also corrected the explanation of how to add tabbed examples and written how to publish CSS examples.

I have also added more details, as to what exactly people can contribute.

Fixes #2360


bsmth (approved):
Thanks @NiedziolkaMichal 👍🏻

PR #2365, opened 2022-12-16
Add JS example: async function expression

NiedziolkaMichal:
Interactive example for [async function expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/async_function).

image


live-examples/js-examples/expressions/expressions-async-function.js
```diff
@@ -0,0 +1,18 @@
...
+  return new Promise(resolve => {
+    console.log(`getting user id for ${userName}`);
+    setTimeout(() => {
+      // returning fixed id for now
+      resolve(512);
+    }, 2000);
+  });
+}
+
+const getUserId = async function(userName) {
```
Josh-Cena:
```suggestion
const getUserId = async function (userName) {
```

Josh-Cena:
Frankly the example is quite artificial anyway. What about using the same function as `Statements/async_function`?

queengooborg:
I think that using the existing example for `statement/statement-async.js` makes a lot of sense here; I don't see much of a reason to have a separate example that practically demonstrates the same thing.  I'm going to close this PR accordingly.

PR #2370, opened 2022-12-19
fix(stage-deploy): build main with local bob

caugner:
1. Fixes the `stage-deploy` by using `mdn-bob` as specified in the `package.json` / `yarn.lock`.
2. Run the workflow on pushes to `main` (continuous deployment).
3. Format the workflow file with prettier (`npx prettier -w`).

~~Depends on https://github.com/mdn/interactive-examples/pull/2388.~~

### DIFF
--- a/.github/workflows/stage-deploy.yml
+++ b/.github/workflows/stage-deploy.yml
@@ -14,14 +14,6 @@ jobs:
 
         steps:
             - uses: actions/checkout@v3
-              with:
-                  path: examples
-
-            - uses: actions/checkout@v3
-              with:
-                  repository: mdn/bob
-                  ref: stage
-                  path: bob
 
             - name: Setup Node.js environment
               uses: actions/setup-node@v3
@@ -30,15 +22,10 @@ jobs:
                   cache: yarn
 
             - name: Install all yarn packages
-              run: yarn --production --frozen-lockfile
-              working-directory: bob
+              run: yarn --frozen-lockfile
 
             - name: Build all
-              # TODO: bob uses relative paths in bundles. For now we just symlink.
-              run: |
-                  ln -s ../bob/node_modules .
-                  node ../bob/lib/mdn-bob.js
-              working-directory: examples
+              run: yarn build
 
             - name: Configure AWS credentials for stage
               uses: aws-actions/configure-aws-credentials@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/stage-deploy.yml
```diff
@@ -20,17 +23,16 @@ jobs:
...
                   repository: mdn/bob
-                  ref: stage
+                  ref: main
                   path: bob
 
             - name: Setup Node.js environment
               uses: actions/setup-node@v3
               with:
                   node-version: '16'
-                  cache: yarn
```
nschonni, line 27:
```suggestion
           cache: npm
           cache-dependency-path: bob/package-lock.json

```

caugner: Didn't work, but I'll try again. :)

caugner: Looking good, thanks, must have had a typo somewhere.

.github/workflows/stage-deploy.yml
```diff
@@ -20,25 +28,26 @@ jobs:
...
               working-directory: bob
 
             - name: Build all
               # TODO: bob uses relative paths in bundles. For now we just symlink.
               run:  |
-                  ln -s ../bob/node_modules .
-                  node ../bob/lib/mdn-bob.js
-              working-directory: examples
+                  rm -rf ./docs
+                  npm run build ../examples
```
caugner: This doesn't actually build the examples in `../examples`, just bob's own (test) examples.

caugner:
@NiedziolkaMichal @nschonni Feel free to merge this PR after approving. This will not only fix the `stage-deploy` workflow, which was broken for months, but also makes sure the interactive-examples on stage are built on every push to `main`.

Once this works, we can replicate this for prod, except that we would probably want to build prod only once per day.

NiedziolkaMichal:
This looks good to me, but I am not into GitHub workflows, so it would be great if @nschonni could check it too.

caugner:
Alternatively, @schalkneethling might be able to have a look next week. 🙏

nschonni (approved):
I guess this goes from using the live bob to the NPM published version, so i think it looks fine

caugner:
Thank you @nschonni! 🙏

PR #2372, opened 2022-12-21
Add idle workflow

MrBrain295:


NiedziolkaMichal:
Honestly, I don't see any value in adding `idle` tag to all of those old PRs, that are awaiting BOB update or better browser support.

MrBrain295:
See: https://github.com/mdn/mdn/issues/300

NiedziolkaMichal:
@MrBrain295 Alright, I wasn't aware.

bsmth:
Thanks both, it looks like this has already landed. These are being added across all MDN repos for consistency. Let us know your thoughts on the interval used to determine `idle` etc.

Going to close now 👍🏻

bsmth:
BTW, the defaults for the workflow are here: https://github.com/mdn/workflows/blob/main/.github/workflows/idle.yml

There's probably a way to override the inputs for "number days stale" per repo to adjust to how we want it here.

https://github.com/mdn/workflows/blob/main/.github/workflows/idle.yml#L20-L24

PR #2373, opened 2022-12-21
Normalize JS comments

NiedziolkaMichal:
This PR fixes plenty of minor issues in JS comments:
- There is now always an empty space immediately after `//`
- String output is now always wrapped into quotes
- Array output now always starts with the word `Array`
- Comment message starts with a lowercase
- Fixed a few instances of `expected output:` text missing or malformed
- Removed comments about bigint not being supported by all browsers

### DIFF
--- a/live-examples/js-examples/array/array-fill.js
+++ b/live-examples/js-examples/array/array-fill.js
@@ -2,11 +2,11 @@ const array1 = [1, 2, 3, 4];
 
 // fill with 0 from position 2 until position 4
 console.log(array1.fill(0, 2, 4));
-// expected output: [1, 2, 0, 0]
+// expected output: Array [1, 2, 0, 0]
 
 // fill with 5 from position 1
 console.log(array1.fill(5, 1));
-// expected output: [1, 5, 5, 5]
+// expected output: Array [1, 5, 5, 5]
 
 console.log(array1.fill(6));
-// expected output: [6, 6, 6, 6]
+// expected output: Array [6, 6, 6, 6]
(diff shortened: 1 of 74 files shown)


Josh-Cena:
> Comment message starts with a lowercase

That's unfortunately against the code example guideline. I don't know what the convention here is but just want to point that out.

Also it's unfortunate that we add the `Array`/`Object` prefix to every console output—that's a lot of noise, especially for nested arrays.

NiedziolkaMichal:
In such case, I think it might be better to alter comments so they start with an upper case letter. There are about 160 ordinary comments and 1147 `expected output` comments.
@wbamberg @bsmth do you think all of those comments should be changed so they fit the guidelines of examples in the content repository?

It's important to put `Array` and `Object` keywords, so they fit exactly with the output. For `Array.flat` it shows:
image

Josh-Cena:
Yeah, for `Array`/`Object` I mean that the runner's console output includes the keywords.

wbamberg:
> In such case, I think it might be better to alter comments so they start with an upper case letter. There are about 160 ordinary comments and 1147 `expected output` comments. @wbamberg @bsmth do you think all of those comments should be changed so they fit the guidelines of examples in the content repository?

I think that would be a good idea.

> It's important to put `Array` and `Object` keywords, so they fit exactly with the output. For `Array.flat` it shows: image

+1

This is a nice update, @NiedziolkaMichal ! Do you want to merge this and then fix initial letters in a follow-up?

wbamberg:
> Yeah, for `Array`/`Object` I mean that the runner's console output includes the keywords.

I guess it was written to align with the Firefox console, but the Chrome console omits them. I like them FWIW but 
I would be OK to remove them if people feel strongly about it.

Josh-Cena:
Yeah, but Firefox omits the keyword for nested arrays, making the `flat` example's output look less cluttered.

NiedziolkaMichal:
I reverted the change of letter case from this PR. Later I will make the next one, changing the first letters of most of the comments to upper case.

Josh-Cena (approved):

PR #2375, opened 2022-12-21
Deploy workflow requirements

### LINKED ISSUE #2364: [CI] We don't need to perform a prod deploy if meta docs are updated
This is a minor efficiency issue where we're triggering deploys based on any commit to main. We should at least exclude some paths:

```bash
# exclude markdown in root as this gets updated frequently
./*.md

# alternatively, only watch the following?
./bin/**/*
./live-examples/**/*
./media/**/*
Jenkinsfile
yarn.lock
package.json
```

NiedziolkaMichal:
Makes `prod-deploy` workflow run automatically, only when one of the following paths has changed:
- `./bin/**`
- `./live-examples/**`
- `./media/**`
- `Jenkinsfile`
- `package.json`

Fixes #2364

### DIFF
--- a/.github/workflows/prod-deploy.yml
+++ b/.github/workflows/prod-deploy.yml
@@ -4,6 +4,12 @@ on:
     push:
         branches:
             - 'main'
+        paths:
+            - './bin/**'
+            - './live-examples/**'
+            - './media/**'
+            - 'Jenkinsfile'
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thanks @NiedziolkaMichal 👍🏻

Issue #2374, opened 2022-12-21
Missing examples


NiedziolkaMichal:
**Missing HTML examples**
- [ ] [for](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for) #2376 
- [ ] [crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) https://github.com/mdn/interactive-examples/pull/2314
- [ ] [is](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is) https://github.com/mdn/interactive-examples/pull/2234
- [ ] [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) https://github.com/mdn/interactive-examples/pull/2226
- [ ] [slot element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) https://github.com/mdn/interactive-examples/pull/2177
- [ ] [slot attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot) https://github.com/mdn/interactive-examples/pull/2177
- [ ] [part](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/part) https://github.com/mdn/interactive-examples/pull/2177
- [ ] [exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts) https://github.com/mdn/interactive-examples/pull/2224
- [ ] [draggable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable) https://github.com/mdn/interactive-examples/pull/2220
- [ ] [template](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) https://github.com/mdn/interactive-examples/pull/2219
- [ ] [output](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output) https://github.com/mdn/interactive-examples/pull/2214
- [ ] [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) https://github.com/mdn/interactive-examples/pull/2212

**Missing CSS examples**
- [ ] [contain-intrinsic-size](https://developer.mozilla.org/en-US/docs/Web/CSS/contain-intrinsic-size)
- [ ] [contain-intrinsic-width](https://developer.mozilla.org/en-US/docs/Web/CSS/contain-intrinsic-width)
[...]

PR #2376, opened 2022-12-21
Add an example for `for`

MrBrain295:
#2374


NiedziolkaMichal:
Currently, this example doesn't show how attribute `for` is useful. You have also set its content to the value of attribute `id`, while `name` should be used instead. I think that the value of property `height` is currently invalid for this amount of content.

MrBrain295:
@NiedziolkaMichal I have made the requested changes.

NiedziolkaMichal:
@MrBrain295 Sorry but I don't think that you have. The value of attribute `for` is still pointing to `name` instead of `id` and the example doesn't show how this attribute is useful. You should also get rid of `form` element since we don't use it for similar examples.

MrBrain295:
I’ve applied the feedback except for demonstrating it’s usefulness which I don’t know how I should do.

NiedziolkaMichal:
It would be great if your CSS would have an immediate visual impact. Additionally, the attribute selector might be confusing to the user when the name of the attribute is `for`.
How about something like that?
image
When the reader will click on the labels, he/she might notice that label without the attribute `for` is not setting the focus state to the control field. Additionally, usage its usage in the `output` element is shown. This example should also have JS code though because `output` is supposed to be an interactive element.

estelle:
i added your suggestions. Now we just need to add the JS file.

queengooborg:
Thanks everyone -- because this has been sitting around for a while, I've decided to go ahead and make the requested changes.  This is now ready to go!

queengooborg (approved):

PR #2377, opened 2022-12-21
Clarify error message

### LINKED ISSUE #23112: Change to a more clear message
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target

try_it

I'm not a native English speaker. Unfortunately, I erroneously translated the message "Foo() must **be** called with new" in my head to just the opposite: "Foo() must **have been** called with the new operator". Hence I get confused about the behavior of the invocation. Please consider changing the message to a more clear version in active tense, e.g., "You must call Foo() with 'new'. Abort. ". 

just the opposite behavior

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/javascript/reference/operators/new.target`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/operators/new.target/index.md
* Last commit: https://github.com/mdn/content/commit/0f248adcab759bdad247a5dbfb7da12dc32bce59
* Document last modified: 2022-12-13T20:26:24.000Z

</details>

dawei-wang:
Fix mdn/content#23112

### DIFF
--- a/live-examples/js-examples/expressions/expressions-newtarget.js
+++ b/live-examples/js-examples/expressions/expressions-newtarget.js
@@ -1,10 +1,10 @@
 function Foo() {
-  if (!new.target) { throw 'TypeError: calling Map constructor without new is invalid'; }
+  if (!new.target) { throw 'TypeError: calling Foo constructor without new is invalid'; }
 }
 
 try {
   Foo();
 } catch (e) {
   console.log(e);
-  // expected output: "TypeError: calling Map constructor without new is invalid"
+  // expected output: "TypeError: calling Foo constructor without new is invalid"
 }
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal (changes requested):
Thanks for making this PR. Word `Map` should be `Foo` in both messages. Could you change it, please?

NiedziolkaMichal (approved):
Everything looks good now, thanks

live-examples/js-examples/expressions/expressions-newtarget.js
```diff
@@ -1,10 +1,10 @@
 function Foo() {
-  if (!new.target) { throw 'Foo() must be called with new'; }
+  if (!new.target) { throw 'TypeError: calling Foo constructor without new is invalid'; }
```
Josh-Cena, line 2: Oops, this should really be: `throw new TypeError ("calling Foo constructor without new is invalid")`

But when we one day enable ESLint, this would be caught by `no-throw-literals` anyway.

NiedziolkaMichal: You are right. I see that string is thrown also in a few other examples, so I will change them all at once. Thanks

PR #2378, opened 2022-12-21
Only throw Error objects in examples

NiedziolkaMichal:
Changes 3 examples in which string is thrown, instead of an object of exception.


Josh-Cena (approved):
LGTM

NiedziolkaMichal:
Thank you for the review

PR #2384, opened 2022-12-22
Add issue templates

bsmth:
This PR adds issue templates with links to common resources and prompts that readers can contribute themselves.


dipikabh (commented):
Thanks for taking up this one, @bsmth! 🙌  

Only one question regarding `.prettierrc` - okay to go with the custom configuration for this repo?

bsmth:
> okay to go with the custom configuration for this repo

Good question, it might be good to align with content configs for JS, CSS, HTML formatting, but there might be some special cases for the editor. We should check in a follow-up, I think.

dipikabh:
> Good question, it might be good to align with content configs for JS, CSS, HTML formatting, but there might be some special cases for the editor. We should check in a follow-up, I think.

Sounds good

dipikabh (approved):

PR #2385, opened 2022-12-22
Change `expected output` to `Expected output`

NiedziolkaMichal:
This PR changes all `// expected output` to `// Expected output`, so comments follow the guidelines of examples from the content repository. I have decided to split this change into 2 separate PRs, making them easier to review.

PR #2386, opened 2022-12-22
Make comments start with upper case letter

NiedziolkaMichal:
This PR changes the first letter of comments to upper case, making them follow the guidelines of examples from the content repository. I have changed all comments except for:
- Comments that start with `// expected output`, which were placed in PR #2385
- Comments that start with a digit
- `// eslint-disable-next-line use-isnan`
- Comments referring to an element in JavaScript code, like `console.log`
- Comments containing the second part of a sentence

### DIFF
--- a/live-examples/js-examples/array/array-findlastindex.js
+++ b/live-examples/js-examples/array/array-findlastindex.js
@@ -4,4 +4,4 @@ const isLargeNumber = (element) => element > 45;
 
 console.log(array1.findLastIndex(isLargeNumber));
 // expected output: 3
-// (index of element with value: 130)
+// (Index of element with value: 130)
--- a/live-examples/js-examples/array/array-findlastindex.js
+++ b/live-examples/js-examples/array/array-findlastindex.js
@@ -4,4 +4,4 @@ const isLargeNumber = (element) => element > 45;
 
 console.log(array1.findLastIndex(isLargeNumber));
 // expected output: 3
-// (Index of element with value: 130)
+// Index of element with value: 130
--- a/live-examples/js-examples/date/date-gettimezoneoffset.js
+++ b/live-examples/js-examples/date/date-gettimezoneoffset.js
@@ -3,7 +3,7 @@ const date2 = new Date('August 19, 1975 23:15:30 GMT-02:00');
 
 console.log(date1.getTimezoneOffset());
 // expected output: your local timezone offset in minutes
-// (eg -120). NOT the timezone offset of the date object.
+// (Eg -120). NOT the timezone offset of the date object.
 
 console.log(date1.getTimezoneOffset() === date2.getTimezoneOffset());
 // expected output: true
--- a/live-examples/js-examples/date/date-gettimezoneoffset.js
+++ b/live-examples/js-examples/date/date-gettimezoneoffset.js
@@ -3,7 +3,7 @@ const date2 = new Date('August 19, 1975 23:15:30 GMT-02:00');
 
 console.log(date1.getTimezoneOffset());
 // expected output: your local timezone offset in minutes
-// (Eg -120). NOT the timezone offset of the date object.
+// (e.g., -120). NOT the timezone offset of the date object.
 
 console.log(date1.getTimezoneOffset() === date2.getTimezoneOffset());
 // expected output: true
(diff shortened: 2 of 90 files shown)


live-examples/js-examples/typedarray/typedarray-fill.js
```diff
@@ -1,5 +1,5 @@
 const uint8 = new Uint8Array([0, 0, 0, 0]);
-// (value, start position, end position);
+// (Value, start position, end position);
```
dipikabh:
```suggestion
// Value, start position, end position
```

live-examples/js-examples/typedarray/typedarray-copywithin.js
```diff
@@ -1,6 +1,6 @@
 const uint8 = new Uint8Array([ 1, 2, 3, 4, 5, 6, 7, 8 ]);
 
-// (insert position, start position, end position)
+// (Insert position, start position, end position)
```
dipikabh:
```suggestion
// Insert position, start position, end position
```

live-examples/js-examples/array/array-findlastindex.js
```diff
@@ -4,4 +4,4 @@ const isLargeNumber = (element) => element > 45;
 
 console.log(array1.findLastIndex(isLargeNumber));
 // expected output: 3
-// (index of element with value: 130)
+// (Index of element with value: 130)
```
dipikabh: Do we need the brackets here?

live-examples/js-examples/date/date-gettimezoneoffset.js
```diff
@@ -3,7 +3,7 @@ const date2 = new Date('August 19, 1975 23:15:30 GMT-02:00');
 
 console.log(date1.getTimezoneOffset());
 // expected output: your local timezone offset in minutes
-// (eg -120). NOT the timezone offset of the date object.
+// (Eg -120). NOT the timezone offset of the date object.
```
dipikabh:
```suggestion
// (e.g., -120). NOT the timezone offset of the date object.
```
- the "e.g.," contraction should begin in lower case
- Should "NOT" be "Note"?

live-examples/js-examples/date/date-sethours.js
```diff
@@ -3,7 +3,7 @@ event.setHours(20);
 
 console.log(event);
 // expected output: "Tue Aug 19 1975 20:15:30 GMT+0200 (CEST)"
-// (note: your timezone may vary)
+// (Note: your timezone may vary)
```
dipikabh: Do we need the brackets here? and all the similar instances of this "Note" in other files

live-examples/js-examples/regexp/regexp-prototype-source.js
```diff
@@ -8,4 +8,4 @@ console.log(new RegExp().source);
 
 console.log(new RegExp('\n').source === '\\n');
 // expected output: true (starting with ES5)
-// (due to escaping)
+// (Due to escaping)
```
dipikabh: Do we need the brackets here?

dipikabh (changes requested):
Thanks for these updates, @NiedziolkaMichal! I've suggested a few edits for your consideration.

NiedziolkaMichal:
We could also remove brackets from all instances of `// (Max signed 8-bit integer)`. @dipikabh Do you agree?

dipikabh:
> We could also remove brackets from all instances of `// (Max signed 8-bit integer)`. @dipikabh Do you agree?

Agree!

live-examples/js-examples/date/date-gettimezoneoffset.js
NiedziolkaMichal: I have corrected `e.g.,`. Word `NOT` is fine in this context, because `getTimezoneOffset` always returns the timezone offset of a user(considering his localization). Values inside the Date object might matter, because of Dayling Saving Time, but it's explained by an article in detail, so I think it's fine to keep it simple in here.

NiedziolkaMichal:
@dipikabh Great! I have applied all of your suggestions.

dipikabh (approved):
Looks good, thank you!

NiedziolkaMichal:
Thank you for the review.

PR #2389, opened 2022-12-23
Fix automatic deployment

NiedziolkaMichal:
This PR fixes the automatic deployment of interactive examples, broken by PR #2375 
It looks like `./` prefix made workflow not work anymore. I have tested it in my fork and everything should be fine now.


bsmth (approved):
Good catch, thanks. Let's give it a shot

NiedziolkaMichal:
It works fine :)

PR #2391, opened 2022-12-26
Fix: correct return type for function

MendyBerger:
### Description
Correct return type for function

### DIFF
--- a/live-examples/wat-examples/statements/select.wat
+++ b/live-examples/wat-examples/statements/select.wat
@@ -8,7 +8,7 @@
     i32.const 0
     select
   )
-  (func (export "select_externref") (param $value externref) (param $condition i32) (result i32)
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal (changes requested):
This example, with or without your changes, produces the following exception:
image

MendyBerger:
@NiedziolkaMichal should be fixed with https://github.com/mdn/bob/pull/1046

NiedziolkaMichal:
You are right, the error is gone in the new BOB version. Is this the expected result?
image

MendyBerger:
Yes

NiedziolkaMichal:
Alright, so everything looks good. Thanks

NiedziolkaMichal (approved):

PR #2397, opened 2023-01-01
Add lang attribute that should be present

sposhe:
The [usage notes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i#usage_notes) section of the `<i>` page in the MDN HTML Reference states that this element can be used to mark up—

> Idiomatic terms from another language (such as "et cetera"); these should include the lang attribute to identify the language

There are two terms in Latin in this example, but they have `class="latin"` (which is unused in the CSS) instead of the `lang="la"` that should be present.

This change replaces that `class` attribute with the `lang` attribute that should be present.

### Description

Make add `lang` to relevant tags to example better conform to what is suggested in the documentation.

### Motivation

Apply best practices to example.

### DIFF
--- a/live-examples/html-examples/inline-text-semantics/i.html
+++ b/live-examples/html-examples/inline-text-semantics/i.html
@@ -1,5 +1,5 @@
 <p>I looked at it and thought <i>This can't be real!</i></p>
 
-<p><i class="latin">Musa</i> is one of two or three genera in the family <i class="latin">Musaceae</i>; it includes bananas and plantains.</p>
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal (approved):
Thank you, that looks good.

sposhe:
@NiedziolkaMichal Was this the right place to edit the interactive example that is iframed onto the [`<i> Element Reference`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i#try_it) page? I see this has been merged, but the example on that page hasn't updated.

NiedziolkaMichal:
It's now published. Your changes are now visible by every developer visiting [&lt;i> page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i)

PR #2403, opened 2023-01-06
Add examples for WeakRef

NiedziolkaMichal:
### Description

Those are examples for [WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef), its [constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef/WeakRef) and [deref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef/deref) method.

image
In this example, object is garbage collected usually after few seconds.

image

image


Josh-Cena (approved):
Just curious: is it better to duplicate two examples with the same code, or is it better to just link the same code to two pages (like in #2404)?

PR #2404, opened 2023-01-06
Add examples for FinalizationRegistry

NiedziolkaMichal:
### Description

Those are interactive examples for [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), its [constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry/FinalizationRegistry), [register()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry/register) and [unregister()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry/unregister).

image
Examples for constructor and register() are the same.

image


live-examples/js-examples/finalizationregistry/finalizationregistry.js
```diff
@@ -0,0 +1,20 @@
...
+
+console.log('Triggering garbage collection, which you should never do in production code');
+
+(function allocateMemory() {
+  Array.from({ length: 50000 }, () => () => {});
+
+  if (!arrayCollected) {
+    setTimeout(allocateMemory, 2);
+  }
+})();
```
Josh-Cena, line 20: This looks really... Eye opening 😄

Josh-Cena (approved):
I'm not sure if we show the same example on three pages in other places, but apart from that, it's a really nice example. I'll approve the code.

PR #2406, opened 2023-01-11
Rewrite `string.repeat()` example

NiedziolkaMichal:
### Description

This PR changes [string.repeat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) example, so it doesn't show any scrollbars.
**OLD:**
image

**NEW:**
image


Josh-Cena (approved):
LGTM, thanks

Issue #2054, opened 2022-03-21
Improve contrast of interactive example


schalkneethling:
There are probably other situations where too little contrast makes things hard to see. Compare these white letters on bright yellow background atop this page,

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

I'll stop reporting these, anyone could find them easily.

_Originally posted by @iDave2 in https://github.com/mdn/yari/issues/5387#issuecomment-1055782535_

vedant-z:
![Uploading r4{bfc025b} (290f7d3c-25a8-48b2-b308-0d7245079b1a id=0x7f0a0400 tag=BaseCommentFragment) …]()

@schalkneethling Are you referring to this part. If so I need to change the BG color for better visibility right?

bsmth:
Has this already been fixed in the following PR?

- [x] https://github.com/mdn/css-examples/pull/75

NiedziolkaMichal:
Unfortunately there are still plenty of examples with too low contrast.

vedant-z:
@NiedziolkaMichal Can you point out me some so that I can get started contributing to this repository

NiedziolkaMichal:
Those are all CSS examples that don't meet proper contrast guidelines in dark theme according to the firefox accessibility tool:
- [x] [all](https://developer.mozilla.org/en-US/docs/Web/CSS/all) #2460
- [x] [backdrop-filter](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter) #2460
- [x] [backface-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/backface-visibility) #2460
- [x] [background-attachment](https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment) #2460
- [x] [background-clip](https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip) #2460
- [x] [background-origin](https://developer.mozilla.org/en-US/docs/Web/CSS/background-origin) #2460
- [x] [bottom](https://developer.mozilla.org/en-US/docs/Web/CSS/bottom) #2419
- [x] [box-decoration-break](https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break) #2460
- [x] [box-sizing](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) #2460
- [x] [clear](https://developer.mozilla.org/en-US/docs/Web/CSS/clear) #2460
[...]

PR #2414, opened 2023-01-22
text-emphasis-position CSS contrast

rohini-ranjanR:
### Description
Change color in text demo of https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position#examples to have a good contrast. issue #2054


NiedziolkaMichal:
I will close it as a duplicate of https://github.com/mdn/interactive-examples/pull/2419.

PR #2422, opened 2023-02-09
fix(ci): skip deploy/release on dependabot merges

caugner:
### Description

Avoids failing `prod-deploy`, `stage-deploy` and `publish-release` workflow runs, if they are triggered by a Dependabot merge (which run without permissions).

### Additional details

See:
image


bsmth (approved):
Nice, thanks 👍🏻

PR #2434, opened 2023-02-13
Add resizable ArrayBuffer examples

chrisdavidmills:
### Description

This PR adds interactive examples for the new features related to the [Resizable ArrayBuffer and growable SharedArrayBuffer](https://tc39.es/proposal-resizablearraybuffer/) Stage 3 draft, which was implemented in Chrome 111+.

Note that this just includes the new features related to [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/ArrayBuffer), not [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer). I didn't add examples for those, as SAB is only available in cross-origin isolated browsing contexts. In the SAB pages that currently do have interactive examples (for example [`SharedArrayBuffer.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/slice)), the examples don't run successfully because of this, and I am concerned that it makes the pages look broken.

### DIFF
--- a/live-examples/js-examples/arraybuffer/arraybuffer-maxbytelength.js
+++ b/live-examples/js-examples/arraybuffer/arraybuffer-maxbytelength.js
@@ -1,0 +1,8 @@
+// Create an ArrayBuffer with a size and max length in bytes
+const buffer = new ArrayBuffer(8, { maxByteLength: 16 } );
+
+console.log(buffer.byteLength);
+// Expected output: 8
+
+console.log(buffer.maxByteLength);
+// Expected output: 16
--- a/live-examples/js-examples/arraybuffer/arraybuffer-maxbytelength.js
+++ b/live-examples/js-examples/arraybuffer/arraybuffer-maxbytelength.js
@@ -1,4 +1,3 @@
-// Create an ArrayBuffer with a size and max length in bytes
 const buffer = new ArrayBuffer(8, { maxByteLength: 16 } );
 
 console.log(buffer.byteLength);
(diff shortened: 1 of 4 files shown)


wbamberg:
> SAB is only available in cross-origin isolated browsing contexts. In the SAB pages that currently do have interactive examples (for example [`SharedArrayBuffer.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/slice)), the examples don't run successfully because of this, and I am concerned that it makes the pages look broken.

You're right, we should remove the examples from these pages IMO.

NiedziolkaMichal:
@chrisdavidmills We are already taking a look into SharedArrayBuffer in issue mdn/fred#971. I am pretty confident that we can make cross-origin isolation work fine, so those examples don't need to be removed.

chrisdavidmills:
Thanks for the heads-up @NiedziolkaMichal. I'll not call for the removal of what's already there, in that case.

@wbamberg does what I've added here for the ArrayBuffer pages look OK to merge?

wbamberg (commented):
We don't have explanatory comments in interactive examples, they're "show don't tell". Otherwise these look great.

chrisdavidmills:
@wbamberg ok, gotcha. I've removed the comments

wbamberg (approved):
👍 thank you Chris. It always feels weird asking people to remove helpful comments, and maybe it's a bad policy, but I think it matters to be consistent one way or the other.

PR #2437, opened 2023-02-15
fix(css): Improve contrast for angle

rohini-ranjanR:
### Description
improve contrast by changing color for angle issue #2054 

Relates #2054

### DIFF
--- a/live-examples/css-examples/values-and-units/angle.css
+++ b/live-examples/css-examples/values-and-units/angle.css
@@ -1,5 +1,5 @@
 #example-element {
-    background-color: #ff5b9b;
+    background-color: #0454aa;
     padding: 0.75em;
     width: 180px;
     height: 120px;
--- a/live-examples/css-examples/values-and-units/angle.css
+++ b/live-examples/css-examples/values-and-units/angle.css
@@ -1,5 +1,5 @@
 #example-element {
-    background-color: lightsalmon;
+    background-color: #ff5b9b;
     padding: 0.75em;
     width: 180px;
     height: 120px;
--- a/live-examples/css-examples/values-and-units/angle.css
+++ b/live-examples/css-examples/values-and-units/angle.css
@@ -3,5 +3,5 @@
     padding: 0.75em;
     width: 180px;
     height: 120px;
-    color: white
+    color: white;
 }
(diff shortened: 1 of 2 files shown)


live-examples/css-examples/values-and-units/angle.css
```diff
@@ -1,5 +1,5 @@
 #example-element {
-    background-color: lightsalmon;
+    background-color: #ff5b9b;
```
NiedziolkaMichal: This color still doesn't meet the minimum contrast guidelines, as the text to background contrast ratio is `2.91`.
How about the #0454aa - which meets [AAA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast#related_wcag_success_criteria) requirement with score `7.36`?

NiedziolkaMichal (changes requested):

live-examples/css-examples/values-and-units/angle.css
rohini-ranjanR: I think this color looks good in dark theme but when we switch to light the text in black will little hard to look. What about #ffb703 it looks good in both dark and light theme.

NiedziolkaMichal: This color has a contrast ratio of `1.75` and even for me it's difficult to read the content:
image

rohini-ranjanR: ok suggested color applied

NiedziolkaMichal: @rohini-ranjanR We could still go with something else that you think would be better for both light and dark themes. To check the contrast ratio, first, inspect `#example-element` in dev tools, then add `color: white;` declaration in it(this shouldn't have any effect), and now click on the `white` dot:
image

The contrast ratio will be displayed at the bottom, comparing color with `background-color`.

rohini-ranjanR: Thanks @NiedziolkaMichal . I am using #0118f3 here that is giving a contrast of `8.65`

image

looks like this in **dark theme** 

image

and this in **light theme** 
image

NiedziolkaMichal: Thanks for noticing. How about also changing the color to white, so the example looks good in both themes?

rohini-ranjanR: okay, thanks for your suggestions.

NiedziolkaMichal:
@rohini-ranjanR Could you fix the ESLint error, please?

NiedziolkaMichal (approved):
It looks good. Thank you for your contribution.

PR #2411, opened 2023-01-17
Add an example for BigInt64Array()

MrBrain295:
### Description
Add an example for BigInt64Array().

### Motivation
There isn't already an example.

PR #2439, opened 2023-02-16
Update ESLint ECMA version

NiedziolkaMichal:
### Description

I have changed the ECMAScript version against which ESLint is performing an analysis, so we remove the `globals` property and skip updating it whenever we wish to use some of the new features. Unfortunately, there is no `ESNext` value that we can use, so whenever a new ES version comes up and we get the error `NEW_FEATURE is not defined`, we need to update `env` again.

This also makes it easier to merge #2411 and other Typed Array examples, because we don't need to add `BigInt64Array` and other types to `globals`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,7 @@
 {
     "env": {
         "browser": true,
-        "es6": true,
+        "es2022": true,
         "node": true
     },
     "extends": "eslint:recommended",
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thanks @NiedziolkaMichal, going to merge shortly 👍🏻

PR #2441, opened 2023-02-16
Remove `lang` from scientific names

### LINKED ISSUE #2440: Scientific names of organisms should not have a lang="la" attribute
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

Try it > taxonomic name example

The example given for scientific names uses a `lang="la"` attribute.

```html
<p><i lang="la">Musa</i> is one of two or three genera in the family <i lang="la">Musaceae</i>; it includes bananas and plantains.</p>
```

While it is appropriate to use the `<i>` element for scientific names, the `lang` attribute is inappropriate because scientific names are not necessarily Latin. Many are derived from Latin, but many are also derived from Ancient Greek, and [several other sources](https://en.wikipedia.org/wiki/Binomial_nomenclature#Derivation_of_binomial_names).

> A complete binomial name is always treated grammatically as if it were a phrase in the Latin language (hence the common use of the term "Latin name" for a binomial name). However, the two parts of a binomial name can each be derived from a number of sources, of which Latin is only one.

So, scientific names are _Latinised_ but they are not, strictly speaking, _Latin_.

Wikipedia's manual of style states the following:

> Although often derived from Latin or Ancient Greek, scientific names are never marked up with {{[lang](https://en.wikipedia.org/wiki/Template:Lang)}} or related templates.

Nigel Chaffey, Senior Lecturer in Botany at Bath Spa University argues:

> I acknowledge that in formulating such binomials [the words must follow the rules of Latin grammar](http://www.scienceprofonline.com/biology-general/biological-classification-binomial-nomenclature.html) (and are therefore Latinised), which is emphasised by the words being given emphasis [in italic script](http://www.dailywritingtips.com/italicizing-foreign-words/) (as is also the case in UK and US English [for unfamiliar Latin words and foreign phrases](https://grammarpartyblog.com/2012/02/23/when-to-italicize-foreign-words-and-phrases/)). But, I also recognise that the words used in those names are often derived from languages other than Latin, such as [Greek](http://tomclothier.hort.net/page36.html), Japanese (e.g. the genera [Tsuga](http://scholar.lib.vt.edu/ejournals/JARS/v18n4/v18n4-kruckeberg.htm), and [Ginkgo](http://imaginatorium.org/sano/ginkgo2.htm)), slang/conversational English (e.g. the genus [Hebejeebie](http://www.theplantlist.org/1.1/browse/A/Scrophulariaceae/Hebejeebie/)), etc. For the latter reason in particular I think that binomials are best described as scientific names rather than Latin names (and which nicely contrasts with the term ‘[common names](https://extension.umass.edu/floriculture/plant-identification)’ by which most plants are also known in [everyday conversation](http://www.flowers.org.uk/plants/plant-common-names/)).

([Source](https://botany.one/2017/02/latin-name-scientific-name-stand/))

I personally worked for over 12 years in biodiversity informatics and although we colloquially refer to scientific names as _Latin names_, we all knew the correct generic term is _scientific name_ or, more specifically, _binomial name_ or _binomials_. These are what we would use in formal writing, reports, and in our specialised databases.

Bearing all of the above in mind, if the intention of the `lang="la"` attribute is to express _[Latinisation](https://en.wikipedia.org/wiki/Latinisation_of_names)_ (the practice of rendering non-Latin names in a Latin style) in the more loose sense, then this attribute remains appropriate.

### MDN metadata

<details>

* Folder: `en-us/web/html/element/i`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/i/index.md
* Last commit: https://github.com/mdn/content/commit/3c925962d641d83660a3c28d3e9a0627a7996183
* Document last modified: 2023-01-19T12:03:58.000Z

</details>

NiedziolkaMichal:
### Description
Removes `lang="la"` attribute from words `Musa` and `Musaceae`, as those are scientific names and shouldn't be marked as Latin.

Fixes #2440

### DIFF
--- a/live-examples/html-examples/inline-text-semantics/i.html
+++ b/live-examples/html-examples/inline-text-semantics/i.html
@@ -1,5 +1,5 @@
 <p>I looked at it and thought <i>This can't be real!</i></p>
 
-<p><i lang="la">Musa</i> is one of two or three genera in the family <i lang="la">Musaceae</i>; it includes bananas and plantains.</p>
+<p><i>Musa</i> is one of two or three genera in the family <i>Musaceae</i>; it includes bananas and plantains.</p>
 
 <p>The term <i>bandwidth</i> describes the measure of how much information can pass through a data connection in a given amount of time.</p>


wbamberg (approved):
👍 thanks for the quick fix @NiedziolkaMichal !

PR #2452, opened 2023-02-24
fix(css): Fix 2 contrast issues

rohini-ranjanR:
### Description

- [transform style](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-style)
- [text-emphasis-position ](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-position)

Relates #2054

### DIFF
--- a/live-examples/css-examples/text-decoration/text-emphasis-position.css
+++ b/live-examples/css-examples/text-decoration/text-emphasis-position.css
@@ -3,5 +3,5 @@ p {
 }
 
 #example-element {
-    text-emphasis: filled double-circle blue;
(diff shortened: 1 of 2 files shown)


NiedziolkaMichal (changes requested):
Thanks for the pull request.
Text parent has too low contrast of `1,71` which should be fixed. Except for that, everything looks fine.

Oh, you also need to fix ESLint issues. Simply run the script `fix:css` before your next commit.

NiedziolkaMichal (approved):
Everything looks good, thank you for your contribution

PR #2458, opened 2023-02-25
build: 10x faster rebuild of pages

NiedziolkaMichal:
### Description

This PR adds `build:pages` script, which builds only pages of the project, without running webpack. This makes `start-watch` script very responsive, as `build:pages` can refresh all examples in around 2 seconds.


bsmth (approved):
Looks good, going to merge shortly 👍🏻

Issue #2463, opened 2023-02-27
Add MathML examples


NiedziolkaMichal:
It is now possible to add MathML interactive examples. Anyone interested is welcome to contribute. Ideally, we would have examples for all of those elements:

- [ ] [&lt;math&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math)
- [ ] [&lt;maction&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/maction)
- [ ] [&lt;annotation&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics)
- [ ] [&lt;annotation-xml&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/semantics)
- [ ] [&lt;menclose&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose)
- [ ] [&lt;merror&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/merror)
- [ ] [&lt;mfenced&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mfenced)
- [ ] [&lt;mfrac&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mfrac)
- [ ] [&lt;mi&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mi)
- [ ] [&lt;mmultiscripts&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mmultiscripts)
- [ ] [&lt;mn&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mn)
- [ ] [&lt;mo&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mo)
- [ ] [&lt;mover&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mover)
- [ ] [&lt;mpadded&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mpadded)
- [ ] [&lt;mphantom&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mphantom)
- [ ] [&lt;mprescripts&gt;](https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mmultiscripts)
[...]

PR #2464, opened 2023-02-27
Semantic exmple

jdavidleon:
### Description
Semantic exmple


NiedziolkaMichal:
Thank you for trying to contribute, but this is not a good improvement:
- Your example throws `Error: people1 is not defined`
- `(4)` shouldn't be added after `Array`, because it will not match the actual output
- Empty line after `work` property is not needed
- Property names `a`, `b`, `c` are perfectly clear, there is no need to change them

jdavidleon:
Hi. Can you tell me how to find examples that are not in the MDN documentation. In order for me to collaborate, I was reading the ECMA 2022 documentation but they are already done.

Or how can I collaborate on something. I learned to program but I have never worked in a team and I would like to do it.

NiedziolkaMichal:
I am glad to hear that. You can find missing examples in #2374 and #2463

PR #2466, opened 2023-02-27
Interactive Example

jdavidleon:
### Description
Descriptive suffixes


NiedziolkaMichal:
Sorry but as I have already told you in #2464 there is no need for this change. We use short names like `a` or `x` in multiple simple examples, where there is no need for bulky and descriptive naming.

PR #2469, opened 2023-02-28
fix(css): fix contrast of 9 css examples

rohini-ranjanR:
### Description
- [pointer-events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events)
- [margin-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline)
- [margin-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start)
- [margin-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end)
- [margin-bottom](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-bottom)
- [margin-block](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block)
- [margin-block-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-start)
- [margin-block-end](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-end)
- [list-style-type](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type)

Relates #2054


NiedziolkaMichal (changes requested):
Very good job, thank you. There are still 3 small things that need to be changed:
- In `list-style-type` example, the text `@counter-style` has a poor contrast in the dark mode.
- In `pointer-events` example, the text `example link` also has poor contrast.
- In the `pointer-events`, could you change the white color of the text `SVG` to be set with `fill` rather than `stroke`? It will be more legible.

NiedziolkaMichal:
@rohini-ranjanR It's not a good idea to use `style` attribute. In both cases, it's best to use a separate class or use tag attribute. Could you also keep those colors consistent across the examples?

NiedziolkaMichal (changes requested):
The shade of green you have chosen doesn't meet contrast requirements now, `#009e5f` will be better. Can you also run `fix:css`, so the empty space after `}` is deleted?

NiedziolkaMichal (approved):

PR #1845, opened 2021-06-09
Add interactive example for Array.prototype.flatMap

tannerdolby:
Fixes #1587

ping @wbamberg

PR #2470, opened 2023-03-02
improve `array-flatmap`

WofWca:
### Description

Show the difference from `flat()`, emphasize that `flat` is done after `map`

### Motivation

The current example is confusing. It makes it look like `flat` is performed before `map`.

Relates to #1845

### DIFF
--- a/live-examples/js-examples/array/array-flatmap.js
+++ b/live-examples/js-examples/array/array-flatmap.js
@@ -1,10 +1,6 @@
 const arr1 = [1, 2, 1];
 
-const result = arr1.flatMap(num => (
-  num === 2
-    ? [2, 2]
-    : 1
-));
+const result = arr1.flatMap((num) => (num === 2 ? [2, 2] : 1));
 
 console.log(result);
 // Expected output: Array [1, 2, 2, 1]
--- a/live-examples/js-examples/array/array-flatmap.js
+++ b/live-examples/js-examples/array/array-flatmap.js
@@ -1,6 +1,10 @@
-const arr1 = [1, 2, [3], [4, 5], 6, []];
+const arr1 = [1, 2, 1];
 
-const flattened = arr1.flatMap(num => num);
+const result = arr1.flatMap(num => (
+  num === 2
+    ? [2, 2]
+    : 1
+));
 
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
Thank you for your contribution. How about keeping the `arr1` multi-dimensional, so users can see that input is also flattened? I suggest making the value of `arr1` be `[1, 2, [3]]` and the result `Array [1, 2, 2, 3]`.

WofWca:
I thought about it, and decided that the initial array doesn't matter. What matters is what we return from the closure. And I have shown that it can either be an array or just a number.

NiedziolkaMichal:
You have, but since the height of the example is now going to be automatically increased, it would be better to show more than that. If you just want to show that flattening is performed only after the callback, I think something like that would be better:
```js
const arr1 = [1, 2, [3, 6]];

const result = arr1.flatMap(e => [e, e * 2])

console.log(result);
// Expected output: Array [1, 2, 2, 4, Array [3, 6], NaN]
```
Because:
- Nested array shows that `e` is an array element that is not affected prior to the callback
- Result of the callback is flattened
- The height of the example is still `shorter`

live-examples/js-examples/array/array-flatmap.js
```diff
@@ -1,6 +1,10 @@
-const arr1 = [1, 2, [3], [4, 5], 6, []];
+const arr1 = [1, 2, 1];
 
-const flattened = arr1.flatMap(num => num);
+const result = arr1.flatMap(num => (
+  num === 2
+    ? [2, 2]
+    : 1
+));
```
Josh-Cena:
```suggestion
const result = arr1.flatMap((num) => (num === 2 ? [2, 2] : 1));
```

Josh-Cena:
The example is now the same height as before :) This is the formatting suggested by Prettier anyway.

Josh-Cena (approved):
This LGTM. @NiedziolkaMichal WDYT?

NiedziolkaMichal (approved):

NiedziolkaMichal:
@Josh-Cena Looks good, thank you both.

PR #2471, opened 2023-03-03
fix(css): Remove invalid CSS comment

NiedziolkaMichal:
### Description

Removes `// thumbs up sign` from `list-style-type`, because it makes the CSS invalid.

Before:
image

After:
image


teoli2003 (approved):
`/* thumbs up sign */` would have been valid, but I'm fine with the removal of the comment.

PR #2473, opened 2023-03-04
fix(css): fix contrast of 9 CSS examples

rohini-ranjanR:
### Description

- [counter-set](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-set)
- [float](https://developer.mozilla.org/en-US/docs/Web/CSS/float)
- [inset-block-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-end)
- [inset-block-start](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-start)
- [inset-block](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block)
- [inset-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end)
- [inset-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-start)
- [inset-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline)
- [inset](https://developer.mozilla.org/en-US/docs/Web/CSS/inset)

Relates to #2054

### DIFF
--- a/live-examples/css-examples/basic-box-model/float.css
+++ b/live-examples/css-examples/basic-box-model/float.css
@@ -7,8 +7,9 @@
 }
 
 #example-element {
-    border: solid 10px #ffc129;
-    background-color: rgba(255, 244, 219, 0.6);
+    border: solid 10px #efac09;
+    background-color: #040d46;
+    color: white;
     padding: 1em;
     width: 40%;
 }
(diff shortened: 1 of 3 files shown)


NiedziolkaMichal (approved):

NiedziolkaMichal:
Everything seems to be ok. Great job!

PR #2474, opened 2023-03-07
feat(css): Add example of `overflow-clip-margin`

NiedziolkaMichal:
This PR adds an example of [overflow-clip-margin](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin).

image


estelle (approved):
Thanks! 🎉

PR #2475, opened 2023-03-08
fix(css): fix contrast of 4 css examples

rohini-ranjanR:
### Description

- [overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior)
- [scroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior)
- [text-emphasis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis)
- [top](https://developer.mozilla.org/en-US/docs/Web/CSS/top)

Relates to #2054

### DIFF
--- a/live-examples/css-examples/basic-box-model/overscroll-behavior.css
+++ b/live-examples/css-examples/basic-box-model/overscroll-behavior.css
@@ -15,7 +15,7 @@
 #example-element {
     width: 50%;
     height: 12em;
-    border: medium dotted #1b76c4;
+    border: medium dotted #009e5f;
     padding: 0.3em;
     margin: 0 0.3em;
     text-align: left;
--- a/live-examples/css-examples/basic-box-model/overscroll-behavior.css
+++ b/live-examples/css-examples/basic-box-model/overscroll-behavior.css
@@ -15,7 +15,7 @@
 #example-element {
     width: 50%;
     height: 12em;
-    border: medium dotted #009e5f;
+    border: medium dotted #1b76c4;
     padding: 0.3em;
     margin: 0 0.3em;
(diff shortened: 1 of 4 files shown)


live-examples/css-examples/basic-box-model/overscroll-behavior.css
```diff
@@ -15,7 +15,7 @@
 #example-element {
     width: 50%;
     height: 12em;
-    border: medium dotted #1b76c4;
+    border: medium dotted #009e5f;
```
NiedziolkaMichal: Sorry, I made a mistake in the list of contrast issues. This shouldn't be changed, because contrast looks good and also `overscroll-behavior-x` and `overscroll-behavior-y` use the same color.

live-examples/css-examples/text-decoration/text-emphasis.html
```diff
@@ -7,7 +7,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">text-emphasis: filled red;</code></pre>
+        <pre><code class="language-css">text-emphasis: filled #ffb703;</code></pre>
```
NiedziolkaMichal: Could we please keep the red color in here? Contrast for it is fine and it's good to have a different color in the 2 choice boxes.

NiedziolkaMichal (changes requested):
That's a good change, thank you. I have added 2 small suggestions.

NiedziolkaMichal (approved):

NiedziolkaMichal:
Everything looks great, thank you!

PR #2476, opened 2023-03-08
ci(stage-deploy): sync to GCP as well as AWS

bkochendorfer:
### Description

Syncing to GCP as well as AWS

### DIFF
--- a/.github/workflows/stage-deploy.yml
+++ b/.github/workflows/stage-deploy.yml
@@ -65,4 +65,4 @@ jobs:
 
       - name: Deploy pages to GCP
         run: |-
-          gsutil -m -h "Cache-Control:public, max-age=86400" rsync -d -r docs/ gs://interactive-examples-stage-mdn/
+          gsutil -m rsync -d -r docs/ gs://interactive-examples-stage-mdn/
--- a/.github/workflows/stage-deploy.yml
+++ b/.github/workflows/stage-deploy.yml
@@ -57,7 +57,7 @@ jobs:
         uses: google-github-actions/auth@v0
         with:
           token_format: access_token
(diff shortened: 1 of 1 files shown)


.github/workflows/stage-deploy.yml
```diff
@@ -41,3 +52,17 @@ jobs:
...
+          token_format: access_token
+          service_account:  deploy-stage-interactive-examp@{{ secrets.GCP_PROJECT_NAME }}.iam.gserviceaccount.com
+          workload_identity_provider: projects/${{ secrets.WIP_PROJECT_ID }}/locations/global/workloadIdentityPools/github-actions/providers/github-actions
+
+      - name: Setup gcloud
+        uses: google-github-actions/setup-gcloud@v1
+
+      - name: Deploy pages to GCP
+        run: |-
+          gsutil -m -h "Cache-Control:public, max-age=86400" rsync -d -r docs/ gs://interactive-examples-stage-mdn/
```
caugner: Currently, we deploy `interactive-examples` as soon as a PR gets merged. If I understand this correctly, this would prevent changes from a PR that is merged within 24h of another PR to be immediately visible. Do we need another step to invalidate the CDN?

caugner (changes requested):
LGTM, but the action fails (possibly due to whitespace in the secrets) and I'm not sure about the `max-age`.

.github/workflows/stage-deploy.yml
bkochendorfer: This sets the cache-control on the object, so it would be set once the CDN hits it. I can remove though and we can always set again if needed.

caugner (approved):

PR #2478, opened 2023-03-09
Fix comment about variable scope

jlpettersson:
### Description
The variable below the comment is a `local` variable and not a `global` variable. This patch fixes the comment.

### Motivation
It is confusing that the comment says one thing but the code says a different thing.

### Additional details

Syntax for Global variables is documented on https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Variables/Global

Syntax for Local variables is documented on https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Variables/Local

### DIFF
--- a/live-examples/wat-examples/statements/loop.wat
+++ b/live-examples/wat-examples/statements/loop.wat
@@ -3,7 +3,7 @@
   (import "console" "log" (func $log (param i32)))
 
   (func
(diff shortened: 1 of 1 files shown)


MendyBerger (approved):
Good catch @jlpettersson, thanks!

teoli2003 (approved):

PR #2481, opened 2023-03-12
feat(css): Add example of `@container`

NiedziolkaMichal:
This PR adds example of [@container](https://developer.mozilla.org/en-US/docs/Web/CSS/@container).

image

I have also moved the `contain` property to a new directory called `container`. I think we could put in here also container properties and contain-visibility properties in the future.


estelle (approved):
thank you

PR #2494, opened 2023-03-25
update: Added closing tags to the remaining input fields

AurelianSpodarec:
Added closing tags to the other remaining input fields so it's consistent.

### DIFF
--- a/live-examples/html-examples/forms/fieldset.html
+++ b/live-examples/html-examples/forms/fieldset.html
@@ -2,10 +2,10 @@
   <fieldset>
     <legend>Choose your favorite monster</legend>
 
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
Could you please remove the closing `/` from the `#mothman` instead? We don't close void elements in this repository. Exactly the same problem is present in the `legend` example. If you could remove the slash from that example too, it would be great.

AurelianSpodarec:
Sure, I've removed the `/` from the input and from the legend example as well in a new PR. Though perhaps all of that should have been one PR, not sure.

NiedziolkaMichal (approved):

PR #2498, opened 2023-03-31
docs(fix): Add clip keyword value to overflow examples

dipikabh:
### Description

The syntax on the [`overflow`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#formal_syntax) page states five possible values for the `overflow` shorthand property and so do the syntax on the pages [`overflow-x`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x#formal_syntax) and [`overflow-y`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y#formal_syntax) - `visible`, `hidden`, `clip`, `scroll`, and `auto`.

While all other values are demo'ed in the [interactive example](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y#try_it), `clip` is missing.

### Motivation

All this pull request does is to add the `clip` keyword to the interactive examples for the sake of completeness.

Related content update: https://github.com/mdn/content/pull/25749
Related doc issue tracker: https://github.com/mdn/content/issues/25355

### DIFF
--- a/live-examples/css-examples/basic-box-model/overflow-block.html
+++ b/live-examples/css-examples/basic-box-model/overflow-block.html
@@ -15,6 +15,13 @@
     </div>
 
     <div class="example-choice">
+        <pre><code class="language-css">overflow-x: clip;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
         <pre><code class="language-css">overflow-block: scroll;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
--- a/live-examples/css-examples/basic-box-model/overflow-block.html
+++ b/live-examples/css-examples/basic-box-model/overflow-block.html
@@ -15,7 +15,7 @@
     </div>
 
     <div class="example-choice">
-        <pre><code class="language-css">overflow-x: clip;</code></pre>
+        <pre><code class="language-css">overflow-block: clip;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
         </button>
--- a/live-examples/css-examples/basic-box-model/overflow-inline.html
+++ b/live-examples/css-examples/basic-box-model/overflow-inline.html
@@ -15,6 +15,13 @@
     </div>
 
     <div class="example-choice">
+        <pre><code class="language-css">overflow-x: clip;</code></pre>
+        <button type="button" class="copy hidden" aria-hidden="true">
+            <span class="visually-hidden">Copy to Clipboard</span>
+        </button>
+    </div>
+
+    <div class="example-choice">
         <pre><code class="language-css">overflow-inline: scroll;</code></pre>
         <button type="button" class="copy hidden" aria-hidden="true">
             <span class="visually-hidden">Copy to Clipboard</span>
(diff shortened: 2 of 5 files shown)


NiedziolkaMichal (changes requested):
It looks good. Could you please also add this value to the `overflow-inline` and `overflow-block` examples? They are not currently displayed, because of a lack of support on chrome, but when we finally add them, we won't have to make another pull request.

dipikabh:
Thanks for reviwing, @NiedziolkaMichal!

Nice catch - that seems correct. To confirm, `overflow-block`and `overflow-inline` in the spec's [Property Index](https://w3c.github.io/csswg-drafts/css-overflow/#property-index) do have the `clip` value. Interestingly, from `overflow` property's [bcd table](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#browser_compatibility), `clip` is supported.

> Could you please also add this value to the overflow-inline and overflow-block examples?

Both [`overflow-block`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-block) and [`overflow-inline`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-inline) pages do not have interactive examples right now. Did you mean adding interactive examples to those pages?

There are other sections - "Syntax", "Values", "Examples" - where `clip` will need to be added, all of which are in mdn/content. I'll open a separate PR for those updates.

NiedziolkaMichal:
Both of those properties have interactive examples, but they are not added to MDN because of poor browser support. Please take a look at those 2 links: [overflow-block](https://github.com/mdn/interactive-examples/blob/main/live-examples/css-examples/basic-box-model/overflow-block.html), [overflow-inline](https://github.com/mdn/interactive-examples/blob/main/live-examples/css-examples/basic-box-model/overflow-inline.html)

dipikabh:
Done. Added `clip` to `overflow-block` and `overflow-inline`.

Sorry for the delay on this.
Could you take another look at this @NiedziolkaMichal?

live-examples/css-examples/basic-box-model/overflow-inline.html
```diff
@@ -14,6 +14,13 @@
         </button>
     </div>
 
+    <div class="example-choice">
+        <pre><code class="language-css">overflow-x: clip;</code></pre>
```
NiedziolkaMichal: This needs to be `overflow-inline: clip;`

live-examples/css-examples/basic-box-model/overflow-block.html
```diff
@@ -14,6 +14,13 @@
         </button>
     </div>
 
+    <div class="example-choice">
+        <pre><code class="language-css">overflow-x: clip;</code></pre>
```
NiedziolkaMichal: This needs to be `overflow-block: clip;`

NiedziolkaMichal (changes requested):

live-examples/css-examples/basic-box-model/overflow-block.html
dipikabh:
```suggestion
        <pre><code class="language-css">overflow-block: clip;</code></pre>
```

live-examples/css-examples/basic-box-model/overflow-inline.html
dipikabh:
```suggestion
        <pre><code class="language-css">overflow-inline: clip;</code></pre>
```

dipikabh:
Oops, silly copy-paste mistake!
Fixed now @NiedziolkaMichal

NiedziolkaMichal (approved):

NiedziolkaMichal:
Great, thank you!

PR #2519, opened 2023-05-04
fix CSS/@layer try-it: make text-align effect visible

leeyc0:
### Description

built in #html-output: fit-content limits the text box width, this PR unset it to make text-align of try-it example visible

### DIFF
--- a/live-examples/css-examples/cascading-and-inheritance/at-rule-layer.css
+++ b/live-examples/css-examples/cascading-and-inheritance/at-rule-layer.css
@@ -1,21 +1,17 @@
 @layer module, state;
 
-#html-output {
-  width: unset;
-}
-
 @layer state {
     .alert {
         background-color: brown;
     }
     p {
-        text-align: right;
+        border: medium solid limegreen;
     }
 }
 
 @layer module {
     .alert {
-        text-align: left;
+        border: medium solid violet;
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
@leeyc0 Could you tell me what's the purpose of this change? The text inside the brown box is perfectly visible and the CSS of the example is as minimal as possible.
image

leeyc0:
As the title says, It's not about fixing the colour. I just realized text align left/right text-align effect is not visible because the div width was limited with the built-in #html-output CSS class. This PR fixes this by unsetting #html-output width

Screenshot 2023-05-05 at 07-07-17 @layer - CSS Cascading Style Sheets MDN
Screenshot 2023-05-05 at 07-07-41 @layer - CSS Cascading Style Sheets MDN

NiedziolkaMichal:
@leeyc0 How about replacing `text-align` with something else, so we don't have the scrollbar? For example, we could use `border: medium solid coral;` in the p rule.

leeyc0:
pushed a new commit, but replaced the border colour with high contrast.

NiedziolkaMichal (approved):

NiedziolkaMichal:
Looks great. Thanks a lot @leeyc0

PR #2530, opened 2023-06-08
Update attribute-autocomplete.html

Phil-G-94:
Updated attribute-autocomplete.html to include id's corresponding to the respective labels' for attributes. Resolving issue #27126.

### Description

Added id's within `<input>` that correspond to the `<label>`'s for attribute. 

### Motivation

Previous source code did not allow the user to focus the inputs by clicking on the corresponding label, which is the whole point of the interactive demo. 

### Additional details

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

To explicitly associate a <label> element with an <input> element, you first need to add the id attribute to the <input> element. Next, you add the for attribute to the <label> element, where the value of for is the same as the id in the <input> element.

Fixes issue #27126.

### DIFF
--- a/live-examples/html-examples/form-attributes/attribute-autocomplete.html
+++ b/live-examples/html-examples/form-attributes/attribute-autocomplete.html
@@ -1,8 +1,8 @@
 <label for="firstName">First Name:</label>
-<input name="firstName" type="text" autocomplete="given-name">
+<input name="firstName" id="firstName" type="text" autocomplete="given-name">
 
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal (approved):

Phil-G-94:
@NiedziolkaMichal Hello 👋🏼 thank you for the review and approval. Anything else for me to do here or shall I just close down the pull request?

NiedziolkaMichal:
Everything looks good, thank you for this fix.

PR #2553, opened 2023-07-10
Better examples for globalprops-parseint.js

lionel-rowe:
### Description

This change swaps out the current example with simply calling the `parseInt` function directly and showing the results given various inputs.

### Motivation

The current example, `roughScale`, doesn't seem to have any real-world use case and just adds unnecessary confusion.

### Additional details

N/A

N/A


queengooborg (approved):
Agreed, the function isn't helpful and only adds more noise to the example -- LGTM, thank you!

PR #2561, opened 2023-07-22
Run eslint on all .js files

OnkarRuikar:
There is a lot of JavaScript in CSS samples but it wasn't getting linted.

The PR enables linting of all the example JS files and lints all the JS files.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 js/lib/*
 js/mode/*
+live-examples/wat-examples/*
--- a/live-examples/css-examples/animation/animation-state.js
+++ b/live-examples/css-examples/animation/animation-state.js
@@ -1,35 +1,35 @@
 'use strict';
 
-window.addEventListener('load', function() {
-    const el = document.getElementById('example-element');
-    const status = document.getElementById('playstatus');
+window.addEventListener('load', () => {
+  const el = document.getElementById('example-element');
+  const status = document.getElementById('playstatus');
 
-    function update() {
-        status.textContent = 'delaying';
-        el.className = '';
-        window.requestAnimationFrame(function() {
-            window.requestAnimationFrame(function() {
-                el.className = 'animating';
-            });
-        });
-    }
-
-    el.addEventListener('animationstart', function() {
-        status.textContent = 'playing';
+  function update() {
+    status.textContent = 'delaying';
+    el.className = '';
+    window.requestAnimationFrame(() => {
+      window.requestAnimationFrame(() => {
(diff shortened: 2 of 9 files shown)


wbamberg (commented):
Thanks for your PR, @OnkarRuikar !

The idea about these commands is that they can be run by contributors, who are generally making a specific type of contribution - a CSS example, a JS example, an HTML example or a WASM example. So `lint:css` doesn't mean "Lint the CSS" it means "Lint my CSS example". If my CSS example includes JS, which some of them do these days, then `lint:css` should lint the JS too.

So I think the scope of these commands, in this PR, is a little off.

OnkarRuikar:
> So I think the scope of these commands, in this PR, is a little off.

It was already off :O\
In package.json `"lint:css": "stylelint \"**/*.css\"",` lints CSS files under `html-examples` and `webapi-examples` as well. Also, no CONTIRBUTING docs ask HTML example contributors to lint CSS file they introduce. They have to find it the hard way when GitHub CI fails. :confused : 

> So lint:css doesn't mean "Lint the CSS" it means "Lint my CSS example". If my CSS example includes JS, which some of them do these days, then lint:css should lint the JS too.

This is a bad idea. By this logic, in package.json, we need to implement `lint:html`, `lint:wat`, `lint:webapi`, and `lint:some-future-technology` as well. And we'll be duplicating same `stylelint` and `eslint` commands in all the `lint:xx` and `fix:xx` commands twice. :roll_eyes: \
We are here in this PR because current `"fix:js": "eslint --fix ./live-examples/js-examples",` selfishly linted only the JS examples.\
Why not expose only `lint:all` and `fix:all` commands to the contributors?

wbamberg:
I'm just saying what the expectation of these commands is. It's not sensible to have multiple different commands that contributors have to run, in order to lint a single example that includes multiple languages. The fact that the repo is deficient in various other ways doesn't change that.

> Why not expose only `lint:all` and `fix:all` commands to the contributors?

That sounds like a much better idea.

OnkarRuikar:
> That sounds like a much better idea.

I've reverted the changes in `package.json`. I'll create a separate PR for `lint:all` and `fix:all` commands because we'll have to update "CONTRIBUTING" docs as well.

OnkarRuikar:
- Closing this PR in favor of the anticipated followup fix PR of https://github.com/mdn/interactive-examples/pull/2566

OnkarRuikar:
- Closing this PR in favor of the anticipated followup fix PR of https://github.com/mdn/interactive-examples/pull/2566

PR #2562, opened 2023-07-23
Remove figure/figcaption from blockquote.html

### LINKED ISSUE #2563: Blockquote example does not need figure/figcaption
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote

Try it

The "Try it" section wraps the blockquote text in a figure with a figure caption, but does not explain why the figure is used nor explain the impact on screen reader announcements.

The code:
```
<figure>
    <blockquote cite="https://www.huxley.net/bnw/four.html">
        <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
    </blockquote>
    <figcaption>—Aldous Huxley, <cite>Brave New World</cite></figcaption>
</figure>
```

Either:
- no `<figure>` element, or
- an explanation why it is used in this section but not in the "Examples" section later in the page.

Fire up a screen reader and navigate it.

This is the output navigating with the virtual cursor in NVDA/Firefox:

`↓`
> em dash Aldous Huxley, Brave New World, figure, block quote, left quote Words can be like X dash rays, if you use them properly em dash they’ll go through anything. You read and you’re pierced. right quote

`↓`
> out of block quote, caption, em dash Aldous Huxley, Brave New World

The `<figure>` causes an announcement of the figure, which is fine but adds unnecessary noise. The `<figcaption>` causes announcement of the text it contains at the _start_ of the content, prior to announcing the presence of the figure or blockquote Then the contents of the `<figcaption>` are announced again as the user moves the virtual cursor through the DOM.

As a user, it is unclear why I am hearing this information twice (other than the code chosen). As a developer, there is no guidance on what this code will output.

This is the output navigating with the virtual cursor in JAWS/Chrome:

`↓`
> block quote cited from https://www.huxley.net/bnw/four.html

`↓`
> has details, left quote Words can be like X dash rays, if you use them properly em dash they’ll go through anything. You read and you’re pierced. right quote
block quote end

`↓`
> has details, em dash Aldous Huxley, Brave New World

Here, the "has details" announcement is a function of the `<figcaption>`.

I cannot edit this example in GitHub because it comes in from an embed with no immediately apparent documentation on where to find it:
`{{EmbedInteractiveExample("pages/tabbed/blockquote.html","tabbed-standard")}}`

### MDN metadata

<details>

* Folder: `en-us/web/html/element/blockquote`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/blockquote/index.md
* Last commit: https://github.com/mdn/content/commit/e04d8d2766c468f149445c0bf438d09f9b2d188c
* Document last modified: 2023-06-13T08:10:47.000Z

</details>

aardrian:
### Description

Removes the `<figure>` / `<figcaption>` construct that wraps the `<blockquote>` but is not explained in the page nor matches the example later in the page.

### Motivation

In NVDA, the `<figure>` causes an announcement of the figure, which is fine but adds unnecessary noise. The `<figcaption>` causes announcement of the text it contains at the start of the content, prior to announcing the presence of the figure or blockquote Then the contents of the `<figcaption>` are announced again as the user moves the virtual cursor through the DOM.

As a user, it is unclear why I am hearing this information twice (other than the code chosen). As a developer, there is no guidance on what this code will output.

In JAWS, the "has details" announcement is a function of the `<figcaption>`.

### Additional details

See [#28133 Blockquote example does not need figure/figcaption](https://github.com/mdn/interactive-examples/issues/2563)

Fixes [#28133](https://github.com/mdn/interactive-examples/issues/2563)

### DIFF
--- a/live-examples/html-examples/text-content/blockquote.html
+++ b/live-examples/html-examples/text-content/blockquote.html
@@ -1,6 +1,4 @@
-<figure>
-    <blockquote cite="https://www.huxley.net/bnw/four.html">
-        <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
-    </blockquote>
-    <figcaption>—Aldous Huxley, <cite>Brave New World</cite></figcaption>
-</figure>
+<blockquote cite="https://www.huxley.net/bnw/four.html">
+    <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
+    <footer>—Aldous Huxley, <cite>Brave New World</cite></footer>
+</blockquote>


estelle (approved):
Thanks! 🎉

PR #2572, opened 2023-08-02
Format files with Prettier

queengooborg:
This PR formats two files in the repository with Prettier.


hamirmahal:
Do you think it'd be useful to have a pre-commit formatting check, and/or a GitHub Action to check for formatting errors?

queengooborg:
We actually already have the pre-commit hook!  A GitHub Action is also supposed to run, but it doesn't look like it's configured properly at the moment -- I'll fix that up.

live-examples/css-examples/pseudo-class/defined.html
```diff
@@ -1,9 +1,9 @@
 <div class="all-elements">
-    <p>This is an ordinary <b>&;lt;p&gt;</b> element.</p>
+  <p>This is an ordinary <b>&lt;p&gt;</b> element.</p>
```
bsmth, line 2: I'm confused by the original here, I guess it was just a typo.

bsmth (approved):
Looks good, thanks 👍🏻

PR #2576, opened 2023-08-02
feat(ci): run all linting actions

queengooborg:
This PR updates the GitHub Action to test a PR to ensure it also checks Prettier formatting.

PR #2592, opened 2023-08-17
chore(prettier): Ignore implicit closing bracket in rtc element

bsmth:
Slightly ugly workaround to get tests to pass in https://github.com/mdn/interactive-examples/pull/2576

This is the only case of a magic comment necessary, I think we can allow one exception for this.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 docs/
+live-examples/html-examples/inline-text-semantics/rtc.html
--- a/live-examples/html-examples/inline-text-semantics/rtc.html
+++ b/live-examples/html-examples/inline-text-semantics/rtc.html
@@ -1,4 +1,3 @@
-<!-- prettier-ignore -->
 <ruby xml:lang="zh-Hant" style="ruby-position: under;">
     <rbc>
         <rb>馬</rb><rp>(</rp><rt>mǎ</rt><rp>)</rp>
@@ -8,4 +7,5 @@
     </rbc>
     <rtc xml:lang="en" style="ruby-position: over;">
         <rp>(</rp><rt>Malaysia</rt><rp>)</rp>
+    </rtc>
 </ruby>
--- a/live-examples/html-examples/inline-text-semantics/rtc.html
+++ b/live-examples/html-examples/inline-text-semantics/rtc.html
@@ -1,3 +1,4 @@
+<!-- prettier-ignore -->
 <ruby xml:lang="zh-Hant" style="ruby-position: under;">
     <rbc>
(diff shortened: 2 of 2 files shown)


queengooborg:
Is the point of the test to demonstrate that there's an implicit closing bracket...?

(Side note, I can't find where this example is referenced in content...)

bsmth:
> (Side note, I can't find where this example is referenced in content...)

It's used here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rtc

> Is the point of the test to demonstrate that there's an implicit closing bracket...?

No, Prettier throws an error if there's a closing `</rtc>` regardless of the inclusion of an ignore comment:

```log
["ERROR" - 10:29:11 AM] Unexpected closing tag "rtc". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (12:5)
  10 |     <rtc xml:lang="en" style="ruby-position: over;">
  11 |         <rp>(</rp><rt>Malaysia</rt><rp>)</rp>
> 12 |     </rtc>
     |     ^^^^^^
  13 | </ruby>
  14 |
```

Note this is also valid and passes `npm test`:

```html
<rtc xml:lang="en" style="ruby-position: over;"/>
        <rp>(</rp><rt>Malaysia</rt><rp>)</rp>
```

If we remove the Prettier ignore and let the formatter do it's thing it looks like this:

```html
<ruby xml:lang="zh-Hant" style="ruby-position: under">
  <rbc>
    <rb>馬</rb><rp>(</rp><rt>mǎ</rt><rp>)</rp> <rb>來</rb><rp>(</rp><rt>lái</rt><rp>)</rp> <rb>西</rb><rp>(</rp
    ><rt>xī</rt><rp>)</rp> <rb>亞</rb><rp>(</rp><rt>yà</rt><rp>)</rp>
  </rbc>

  <rtc xml:lang="en" style="ruby-position: over" />
  <rp>(</rp><rt>Malaysia</rt><rp>)</rp>
</ruby>
```

Note that this element is considered deprecated so I don't think it's necessary to worry over adding special handling here for tests to pass.

caugner:
@bsmth Maybe add `live-examples/html-examples/inline-text-semantics/rtc.html` to the `.prettierignore` file instead to keep it simple?

bsmth:
> @bsmth Maybe add `live-examples/html-examples/inline-text-semantics/rtc.html` to the `.prettierignore` file instead to keep it simple?

Yeah that's fair, done in 43e6cc6ff55c461e56d0a061463cd044bfdb849d

hirehamir (approved):

caugner (approved):

PR #2597, opened 2023-08-22
Avoid using h24 in examples

hsivonen:
### Description

Avoid suggesting the use of "h24" anywhere.

### Motivation

The hour cycle h24 is specified for completeness, as its existence in the mathematical sense is implied by h11, h12, and h23. However, h24 isn't used anywhere in the world (as far as CLDR is aware).

Superficially, h24 looks like the designator for "the 24-hour clock", but the correct designator for "the 24-hour clock" is h23.

### Additional details

See https://github.com/unicode-org/cldr-json/blob/80a94b0f6c3a34d6e2dc0dca8639a54babc87f94/cldr-json/cldr-core/supplemental/timeData.json (h means h12, H means h23, K means h11. h24 would be k, which doesn't appear there.)

https://github.com/mdn/content/pull/28701


live-examples/js-examples/intl/intl-locale.js
```diff
@@ -1,7 +1,7 @@
 const korean = new Intl.Locale('ko', {
   script: 'Kore',
   region: 'KR',
-  hourCycle: 'h24',
+  hourCycle: 'h12',
```
Josh-Cena: Maybe use `h23`? Otherwise we have two `h12`s.

hsivonen: Done.

Josh-Cena (approved):
LGTM; @bsmth could you merge? I used to have the perms but not anymore.

bsmth:
Thanks both, merging shortly

PR #2604, opened 2023-09-04
fix(css): for font examples use 'rem' instead of 'em'

OnkarRuikar:
- part of fix for https://github.com/mdn/content/issues/14282

There is no change in output because we are not updating font on parent elements in these cases.


chrisdavidmills (approved):
Yup, makes more sense! Thanks @OnkarRuikar.

PR #2611, opened 2023-09-12
Fix double allocation in typedarray-of.js

codefrau:
### Description

The current example for [TypedArray.of()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/of) allocates an empty array first, before invoking the static `of()` method. My fix removes that unnecessary and misleading initial allocation.

### Motivation

A novice user might look at the current example code and infer that they need to create a new typed array first before they can use the `of()` method. That is not the case. Also, simpler is better. 

### Additional details

I replaced

    let int16array = new Int16Array();
    int16array = Int16Array.of('10', '20', '30', '40', '50');

with

    const int16array = Int16Array.of('10', '20', '30', '40', '50');

to avoid the unnecessary allocation of the empty array.

n/a


Josh-Cena (approved):
LGTM, thanks!

PR #2628, opened 2023-10-09
fix(ci): block workflows on forks

### LINKED ISSUE #2586: [meta] CI runs on forks
We should be ignoring forks for the following types of actions:

- https://github.com/bsmth/interactive-examples/actions/workflows/stage-deploy.yml
- https://github.com/bsmth/interactive-examples/actions/workflows/prod-deploy.yml
- CodeQL?

StyleShit:
Closes #2586


.github/workflows/codeql-analysis.yml
```diff
@@ -24,6 +24,7 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    if: startsWith( github.repository, 'mdn/' )
```
caugner: Let's be consistent with [workflows in other repos](https://github.com/mdn/content/blob/6fa75c683af3831c0e988a9270e3b870852dc9a1/.github/workflows/markdown-lint-fix.yml#L10):

```suggestion
    if: github.repository == 'mdn/interactive-examples'
```

.github/workflows/prod-deploy.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       id-token: write
 
     runs-on: ubuntu-latest
-    if: github.triggering_actor != 'dependabot[bot]'
+    if: github.triggering_actor != 'dependabot[bot]' && startsWith( github.repository, 'mdn/' )
```
caugner: dito

.github/workflows/stage-deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       contents: read
       id-token: write
     runs-on: ubuntu-latest
-    if: github.triggering_actor != 'dependabot[bot]'
+    if: github.triggering_actor != 'dependabot[bot]' && startsWith( github.repository, 'mdn/' )
```
caugner: dito

caugner (changes requested):
Thank you! 👍 

LGTM, but let's use the same condition as in other repos, to be consistent.

I can see the advantage of `startsWith`, but we don't have plans to rename the repos.

.github/workflows/codeql-analysis.yml
StyleShit: sure, done :)

.github/workflows/prod-deploy.yml
StyleShit: Done

.github/workflows/stage-deploy.yml
StyleShit: Done

caugner (approved):

bsmth:
Thanks, both! 👍🏻

PR #2630, opened 2023-10-11
Remove slashes from void elements

bennypowers:
https://html.spec.whatwg.org/multipage/syntax.html#start-tags

### Description

Removed slashes from void `<co.>` elements on the col element page example

### Motivation

The [spec](https://html.spec.whatwg.org/multipage/syntax.html#start-tags) discourages use of slash:

>

### Additional details

> On [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), [a slash] does not mark the start tag as self-closing but instead is unnecessary and has no effect of any kind. For such void elements, **it should be used only with caution** — especially since, if directly preceded by an [unquoted attribute value](https://html.spec.whatwg.org/multipage/syntax.html#unquoted), it becomes part of the attribute value rather than being discarded by the parser.

(emphasis mine)

### DIFF
--- a/live-examples/html-examples/table-content/col.html
+++ b/live-examples/html-examples/table-content/col.html
@@ -3,9 +3,9 @@
     Superheros and sidekicks
   </caption>
   <colgroup>
-    <col />
-    <col span="2" class="batman" />
-    <col span="2" class="flash" />
+    <col>
+    <col span="2" class="batman">
+    <col span="2" class="flash">
   </colgroup>
   <tr>
     <td></td>


dipikabh (approved):
Thanks for providing this fix @bennypowers!

We've documented the same guidelines for void elements on MDN. 🙂 ([glossary entry](https://developer.mozilla.org/en-US/docs/Glossary/Void_element))

dipikabh:
Please excuse my previous approval. I looked into the lint errors flagged here - prettier is adding back the slashes that were removed.

Please see this thread https://github.com/orgs/mdn/discussions/242, which captures the discussion and the decision to keep using Prettier and accepting its behavior. This is why we cannot accept the changes and will have to close this PR.

We appreciate your effort to keep MDN accurate. Thank you!

bennypowers:
It's a sad condemnation of the state of the art that the once-proud MDN has capitulated to javascript-tooling's arbitrary opinions, which run counter to both the specifications and your own guidelines.

PR #2633, opened 2023-10-11
fix: `array-reduce` requiring horizontal scrolling

### LINKED ISSUE #2632: `Array.prototype.reduce()` demo requires horizontal scrolling

I think it would be more user-friendly if readers didn't have to horizontally scroll to see the `reduce()` example.

The entire `reduce()` example is visible without any scrolling.

_No response_

_No response_

hirehamir:
fixes #2632.


hamirmahal:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#try_it

hamirmahal:
# Without this change

1680x1050

hamirmahal:
# With this change (presumably)

printWidth100

hamirmahal:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#try_it

live-examples/js-examples/array/.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+// We use `.prettierrc.js` instead of `.prettierrc.json`
+// so we can extend this repository's default `prettier` configuration.
+module.exports = {
+  ...require('../../../.prettierrc.json'),
+  printWidth: 100,
```
hirehamir, line 5:
```suggestion
  // This prevented at least one example from requiring horizontal scrolling.
  printWidth: 100,
```

hirehamir: It might be worth documenting why we're overriding the base `prettier` config.

bsmth: So it appears I set the [print width to 120](https://github.com/mdn/interactive-examples/pull/2274#issuecomment-1268508109) which might not even be desirable overall. It looks like the original objection to `80` as a print width was for changes in an `.md` file (which we can maybe handle with overrides?), not a code example. It might be worth considering getting rid of the `printWidth` key altogether and using the default which is [what we use in content](https://github.com/mdn/content/blob/main/.prettierrc.json) with overrides for html/md. We should check how many files this changes though

__edit:__ It might be good to change the base config only. This is a neat approach, adding an overriding config in a subdir, but it might get confusing with separate config files dotted through the source like this.

yishengjiang99:
i approve

bsmth:
Thanks @hamirmahal - should be fixed now in https://github.com/mdn/interactive-examples/pull/2648 🙌🏻

hamirmahal:
You're welcome @bsmth! Thanks for the fix.

hamirmahal:
You're welcome, @bsmth! Thanks for the fix.

PR #2650, opened 2023-11-03
chore: use reserved values for phone numbers and domains where possible

bsmth:
We should be using example domains instead of real URLs where possible.

### DIFF
--- a/live-examples/css-examples/pseudo-element/after.html
+++ b/live-examples/css-examples/pseudo-element/after.html
@@ -1,6 +1,6 @@
 <p>
   The sailfish is named for its sail-like dorsal fin and is widely considered the fastest fish in the ocean.
-  <a href="https://example.com/wiki/Sailfish">You can read more about it here</a>.
+  <a href="https://en.wikipedia.org/wiki/Sailfish">You can read more about it here</a>.
 </p>
 
 <p>
--- a/live-examples/css-examples/pseudo-element/after.html
+++ b/live-examples/css-examples/pseudo-element/after.html
@@ -1,6 +1,6 @@
 <p>
   The sailfish is named for its sail-like dorsal fin and is widely considered the fastest fish in the ocean.
-  <a href="https://en.wikipedia.org/wiki/Sailfish">You can read more about it here</a>.
+  <a href="https://example.com/wiki/Sailfish">You can read more about it here</a>.
 </p>
 
 <p>
--- a/live-examples/html-examples/inline-text-semantics/q.html
+++ b/live-examples/html-examples/inline-text-semantics/q.html
@@ -1,6 +1,6 @@
 <p>
   When Dave asks HAL to open the pod bay door, HAL answers:
-  <q cite="https://www.imdb.com/title/tt0062622/quotes/?item=qt0396921&ref_=ext_shr_lnk">
+  <q cite="https://www.example.com/movies/2001-space-odyssey/quotes/?item=2">
     I'm sorry, Dave. I'm afraid I can't do that.
   </q>
 </p>
--- a/live-examples/html-examples/inline-text-semantics/q.html
+++ b/live-examples/html-examples/inline-text-semantics/q.html
@@ -1,6 +1,6 @@
 <p>
   When Dave asks HAL to open the pod bay door, HAL answers:
-  <q cite="https://www.example.com/movies/2001-space-odyssey/quotes/?item=2">
+  <q cite="https://www.imdb.com/title/tt0062622/quotes/?item=qt0396921&ref_=ext_shr_lnk">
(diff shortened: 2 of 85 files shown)


pepelsbey (approved):
LGTM 👍

live-examples/webapi-examples/url/url.js
```diff
@@ -1,9 +1,9 @@
-const url = new URL('https://en.wikipedia.org/wiki/Mozilla#Software');
+const url = new URL('https://example.com/wiki/Mozilla#Software');
 console.log(url.protocol);
 // expected output: "https:"
 
 console.log(url.hostname);
-// expected output: "en.wikipedia.org"
+// expected output: "example.com"
```
estelle: here, the original is more informative than the edit.

live-examples/html-examples/inline-text-semantics/q.html
```diff
@@ -1,6 +1,6 @@
 <p>
   When Dave asks HAL to open the pod bay door, HAL answers:
-  <q cite="https://www.imdb.com/title/tt0062622/quotes/?item=qt0396921&ref_=ext_shr_lnk">
+  <q cite="https://www.example.com/movies/2001-space-odyssey/quotes/?item=2">
```
estelle: this removes a little bit of entertainment from MDN. I am not sure I agree with removing actual links to  replace with fake `example.com` links in all cases. Is there a discussion I missed?

live-examples/css-examples/pseudo-element/after.html
```diff
@@ -1,6 +1,6 @@
 <p>
   The sailfish is named for its sail-like dorsal fin and is widely considered the fastest fish in the ocean.
-  <a href="https://en.wikipedia.org/wiki/Sailfish">You can read more about it here</a>.
+  <a href="https://example.com/wiki/Sailfish">You can read more about it here</a>.
```
estelle: people might actually want to learn about sailfish. We should keep the original here.

estelle (commented):
Using `example.com`, `http://example.com/`, `@example.com` for examples is generally a good idea, especially in the case of emails. That said, URLs that give contextual support (or provide entertainment without being promotional) shouldn’t be replaced.

Also, good catch on the phone number. TIL: "The North American Numbering Plan reserves 555-01 numbers for fictitious purposes. "

bsmth:
> Using `example.com`, `http://example.com/`, `@example.com` for examples is generally a good idea, especially in the case of emails. That said, URLs that give contextual support (or provide entertainment without being promotional) shouldn’t be replaced.


Yeah I think this makes sense, I've reverted most of the wiki links, but I agree, with emails and telephone numbers, we should use reserved values. 

I presume it's also important what the example is doing ([fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) or others that are actually making requests should only use example.com, I guess.)

bsmth:
I'm going ahead to merge this one after incorporating the feedback! TY 👍🏻

live-examples/html-examples/inline-text-semantics/q.html
Jxck: @estelle https://github.com/orgs/mdn/discussions/450
If live-example causes real traffic to the site, it should be avoided since MDN has a large number of users.
Example domains are standardized for this purpose in [RFC6761](https://datatracker.ietf.org/doc/html/rfc6761).
MDN should respect it.

live-examples/webapi-examples/url/url.js
Jxck: @estelle @bsmth It perfectly illustrates its api with an example domain.

live-examples/html-examples/inline-text-semantics/q.html
Jxck: cc @bsmth

live-examples/css-examples/pseudo-element/after.html
Jxck: @estelle @bsmth That's not the purpose of this example and MDN.

Jxck:
> Using example.com, http://example.com/, @example.com for examples is generally a good idea, especially in the case of emails. That said, URLs that give contextual support (or provide entertainment without being promotional) shouldn’t be replaced.

If real traffic happens from document, testing, demo etc will suffer domain owners. This is especially true for sites like MDN, which are used by many users.
Example domains are standardized for this purpose in [RFC6761](https://datatracker.ietf.org/doc/html/rfc6761).
MDN should respect it.

When you require some context, `*.example` can be used like `wiki.example`, `movie.example` etc.
I'll re-revert some changes above in separate PR.
If you require more discussion, https://github.com/orgs/mdn/discussions/450 is the best place.

PR #2652, opened 2023-11-05
Fix: expected output comments for indexOf example

tvxxd:
### Description
The comments stated that the first occurrence of the search term "dog" should be at index 7, and the second occurrence at index 41. However, the correct values should be 15 for the first occurrence and 38 for the second occurrence.

incorrect expected output comments in the indexOf example [Fixes mdn/interactive-examples#2654](https://github.com/mdn/interactive-examples/issues/2654)


bsmth:
Thanks a lot for the proposed fix @tvxxd - I've just merged in a similar PR which addresses this in https://github.com/mdn/interactive-examples/pull/2653 so I will close for now :)

PR #2663, opened 2023-11-16
Create an example for content-visibility

dkdk225:
### Description
Create an interactive example for [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility).

### Motivation
To improve understandability of  [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility).

#2374


estelle (approved):
Thanks! 🎉

PR #2666, opened 2023-11-18
fix(css): improve contrast of margin-block

dkdk225:
### Description
Make the text inside the rows at [margin-block-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-start) and [margin-block-end](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-end) examples white.

### Motivation
To prevent the text from becoming black color at light mode causing contrast issues

#2054


estelle (approved):
Thanks! 🎉

PR #2668, opened 2023-11-21
[CSS] Add interactive examples for text-indent values `each-line` and `hanging`

zfox23:
#### Summary

Updates the `text-indent` interactive example to show usage of `each-line` and `hanging`. Also makes the indentation amount for all examples clearer with the addition of a background color behind the text.

#### Related issues

Parent issue: https://github.com/mdn/content/issues/30337

The three relevant PRs:
1. https://github.com/mdn/browser-compat-data/pull/21309
2. https://github.com/mdn/interactive-examples/pull/2668
3. https://github.com/mdn/content/pull/30433

### DIFF
--- a/live-examples/css-examples/text/text-indent.html
+++ b/live-examples/css-examples/text/text-indent.html
@@ -45,8 +45,11 @@
 <div id="output" class="output hidden">
   <section id="default-example">
     <div id="example-element">
-      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam ac dolor fermentum.</p>
-      <p>Curabitur sed eleifend libero. Donec iaculis dignissim egestas.</p>
+      <p>This text is contained within a single paragraph. This paragraph is two sentences long.</p>
+      <p>
+        This text starts a new paragraph. Within this paragraph and after this sentence, there will be a
+        <code>&lt;br&gt;</code> line break element.<br />There it was! How did that affect this paragraph's indentation?
+      </p>
     </div>
   </section>
 </div>
--- a/live-examples/css-examples/text/text-indent.html
+++ b/live-examples/css-examples/text/text-indent.html
@@ -38,12 +38,8 @@
 <div id="output" class="output hidden">
   <section id="default-example">
     <div id="example-element">
-      <p>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam ac dolor fermentum.
-      </p>
-      <p>
-        Curabitur sed eleifend libero. Donec iaculis dignissim egestas.
-      </p>
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam ac dolor fermentum.</p>
+      <p>Curabitur sed eleifend libero. Donec iaculis dignissim egestas.</p>
     </div>
   </section>
 </div>
--- a/live-examples/css-examples/text/text-indent.html
+++ b/live-examples/css-examples/text/text-indent.html
@@ -47,8 +47,8 @@
     <div id="example-element">
       <p>This text is contained within a single paragraph. This paragraph is two sentences long.</p>
       <p>
-        This text starts a new paragraph. Within this paragraph and after this sentence, there will be a
-        <code>&lt;br&gt;</code> line break element.<br />There it was! How did that affect this paragraph's indentation?
+        Within this new paragraph and after this sentence, there is a
+        <code>&lt;br&gt;</code> line break element.<br />There it was! How did that affect the indentation?
       </p>
     </div>
   </section>
--- a/live-examples/css-examples/text/text-indent.html
+++ b/live-examples/css-examples/text/text-indent.html
@@ -33,6 +33,13 @@
       <span class="visually-hidden">Copy to Clipboard</span>
     </button>
   </div>
+
+  <div class="example-choice">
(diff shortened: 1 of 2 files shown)


dipikabh:
Hi @zfox23, I tested the changes in Safari 17.1 and Firefox Nightly (122.01a). 

- The addition of background color is great for visually demonstrating the indentation. Thanks!
- I could not make out whats expected to happen when both `each-line` and `hanging` are present. The "Values" and "Examples" section on the [text-indent](https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent) page don't provide any hint either.
- I noticed that in Safari, the values are `-webkit-each-line` and `-webkit-hanging` and so the three new additions to the interactive example don't produce any indentation (although `text-indent: 3em each-line;` does seem to be working). (Side note to check if BCD for Safari should be updated)
- If the two values `each-line` and `hanging` are currently fully supported only in Firefox, I am wondering if we should hold off publishing the interactive example with those two values until there is broader support across more browsers.

zfox23:
> * The addition of background color is great for visually demonstrating the indentation. Thanks!

yay! :)

> * I could not make out whats expected to happen when both `each-line` and `hanging` are present. The "Values" and "Examples" section on the [text-indent](https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent) page don't provide any hint either.

now that i am looking again, i can't make that out either. lemme think some more about this. i thought i noticed a difference here last week, but maybe i'm imagining that. i'll take that example out if i can't make it relevant.
 
> * I noticed that in Safari, the values are `-webkit-each-line` and `-webkit-hanging` and so the three new additions to the interactive example don't produce any indentation (although `text-indent: 3em each-line;` does seem to be working). (Side note to check if BCD for Safari should be updated)

hmm! i don't think i am seeing the same thing on Safari 17.1. here's what my test looks like there:
CleanShot 2023-11-27 at 11 56 31

is that what you are seeing?

(also, i did change the BCD for Safari for this :) )

> * If the two values `each-line` and `hanging` are currently fully supported only in Firefox, I am wondering if we should hold off publishing the interactive example with those two values until there is broader support across more browsers.

yeah, maybe! what do y'all usually do with interactive examples in cases like this, when baseline browser support isn't yet "high"?

dipikabh:
> i'll take that example out if i can't make it relevant.

Syntactically it is correct though, both can be used together:
```
text-indent = 
 [<length-percentage>] &&
  hanging? &&
  each-line?
```

Example 30 in spec seems to have some hint: https://drafts.csswg.org/css-text/#example-f29e1f90. How about we explain the use case of both values in content? Guess it is to do with hard and soft line breaks

> is that what you are seeing?

Oh yeah, `each-line` and `hanging` are working for me now in Safari, as in your demo. I was trying to update values in the inspector to see the difference and the values `-webkit-each-line` and `-webkit-hanging` in the drop-down were getting selected. But they're not valid! `each-line` and `hanging` work as expected.

> yeah, maybe! what do y'all usually do with interactive examples in cases like this, when baseline browser support isn't yet "high"?

I guess we should be okay here because there's support in 2 out of 3 popular browsers.

dipikabh (approved):
These are good updates, thanks @zfox23!

You can take a call to remove `text-indent: 3em hanging each-line;`.
If you'd like, I can take a look again if anything else in the example changes to accommodate `hanging each-line`. Leaving my approval.

zfox23:
@dipikabh I appreciate that you nudged me to better understand `hanging each-line`! I have made some changes to the example to better show what that means, and I also made a `background-color` change to ensure visible text no matter the page theme. Could you please take another look when you have a moment? Thank you!

2023-11-27_15-19-41_firefox

live-examples/css-examples/text/text-indent.html
```diff
@@ -19,13 +19,36 @@
...
 <div id="output" class="output hidden">
   <section id="default-example">
-    <div>
-      <p id="example-element">
-        Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt.
+    <div id="example-element">
+      <p>This text is contained within a single paragraph. This paragraph is two sentences long.</p>
+      <p>
+        Within this new paragraph and after this sentence, there is a
+        <code>&lt;br&gt;</code> line break element.<br />There it was! How did that affect the indentation?
```
dipikabh: How about:
```suggestion
      <p>
        This is a new paragraph. There is a line break element <code>&lt;br&gt;</code>
        after this sentence.<br />There it is! Notice how it affects the indentation.
```

dipikabh (approved):
You cracked it! Great working example, well done!

> I also made a background-color change to ensure visible text no matter the page theme.

This is great; I did not know how to convey this in my last review :). However, I see the two shades of blue in Safari but not in Firefox 🤔

live-examples/css-examples/text/text-indent.html
zfox23: Works for me!

zfox23:
> You cracked it! Great working example, well done!
> 
> > I also made a background-color change to ensure visible text no matter the page theme.
> 
> This is great; I did not know how to convey this in my last review :). However, I see the two shades of blue in Safari but not in Firefox 🤔

Thank you!

Weird...I changed the named CSS colors to hex values in an attempt to fix what you're seeing. *Shouldn't* matter, but...

zfox23:
i'm gonna merge this now!

PR #2669, opened 2023-11-24
chore(ci): Switch to weekly cron for idle issues

bsmth:
I'm not sure if we need this workflow running daily, maybe a weekly schedule is sufficient:

* `0 10 * * 1` -> https://crontab.guru/#0_10_*_*_1

### DIFF
--- a/.github/workflows/idle.yml
+++ b/.github/workflows/idle.yml
@@ -2,7 +2,7 @@ name: 'Label idle issues'
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


caugner:
> I'm not sure if we need this workflow running daily

The advantage of a daily run is that issues get labeled stale consistently after 30 days, otherwise it will be after 30-36 days.

bsmth:
Thanks, Claas, I'm going to close for now

PR #2675, opened 2023-12-01
Removed comma in try it example on Web/JavaScript/Reference/Global_Objects/Map/groupBy 

fALKENdk:
Removed comma in the [Try it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy#try_it) example callbackFn.

Updated the [Using Map.groupBy()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy#using_map.groupby) example in this [PR](https://github.com/mdn/content/pull/30726).


Josh-Cena (commented):
Thanks, this is the style enforced by our formatter.

PR #2678, opened 2023-12-04
fix(html): provide corresponding 'id's for 'for' attributes

OnkarRuikar:
Many `<label for=` do not have corresponding element with `id`s.

[Without the `id`s the `for` attributes are redundant.](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for)


estelle (approved):
Thank you for doing this!!!

PR #2679, opened 2023-12-04
fix: various code issues

OnkarRuikar:
See commit comments for descriptions for simple changes.


live-examples/html-examples/form-attributes/attribute-capture.html
```diff
@@ -1,7 +1,7 @@
...
+<label for="selfie">Take a picture of your face:</label>
 
-<input type="file" name="picture" accept="image/*" capture="user" />
+<input type="file" id="selfie" name="selfie" accept="image/*" capture="user" />
 
-<label for="voice">Record a sample of your voice:</label>
+<label for="picture">Take a picture using back facing camera:</label>
 
-<input type="file" name="voice" accept="audio/*" capture />
+<input type="file" id="picture" name="picture" accept="image/*" capture="environment" />
```
OnkarRuikar, line 7: The `capture` attribute is [no longer a boolean attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture). The change tries to demonstrate the other `environment` value.
Tested this on iPad for front and back camera uses.

Josh-Cena:
JS changes LGTM

pepelsbey (approved):
Nice cleanup, thanks :) LGTM!

PR #2680, opened 2023-12-04
fix(html): use `<textarea>` in spellcheck example

OnkarRuikar:
- related to https://github.com/mdn/content/pull/30772

`<textarea>` is the most suitable for this demo as spellchecking on it [is supported on more devices than any other element](https://github.com/mdn/content/issues/19285).


teoli2003 (approved):
👍

PR #2693, opened 2023-12-15
Add return_call example for WASM

pepelsbey:
### Description

Adds example for `return_call`, tail call version of the `call`. I took it [from the explainer](https://github.com/WebAssembly/tail-call/blob/main/proposals/tail-call/Overview.md#examples).

But there’s a catch: the playground shows an error when I try to run the code. There might be two reasons for that:

- I don’t fully understand what I’m doing (first time looking at WASM)
- Playground doesn’t support `return_call`, see the error below

### Motivation

To support enabled tail calls in Firefox 121.

### Additional details

```
> Error: parseWat failed:
3:6: error: opcode not allowed: return_call
    (return_call $fac-aux (local.get $x) (i32.const 1))
     ^^^^^^^^^^^
10:10: error: opcode not allowed: return_call
        (return_call $fac-aux
         ^^^^^^^^^^^
```

- https://github.com/mdn/content/issues/30333
- https://github.com/mdn/content/pull/36288

### DIFF
--- a/live-examples/wat-examples/statements/return_call.js
+++ b/live-examples/wat-examples/statements/return_call.js
@@ -1,0 +1,11 @@
+const url = '{%wasm-url%}';
+await WebAssembly.instantiateStreaming(fetch(url), {
+  env: {
+    fac: function (x) {
+      if (x === 0) {
+        return 1;
+      }
+      return x * this.fac(x - 1);
+    },
+  },
+});
--- a/live-examples/wat-examples/statements/return_call.js
+++ b/live-examples/wat-examples/statements/return_call.js
@@ -3,3 +3,4 @@ const { instance } = await WebAssembly.instantiateStreaming(fetch(url));
 const result = instance.exports.fac(5n);
 
 console.log(result);
+// Expected output: 120n
--- a/live-examples/wat-examples/statements/return_call.js
+++ b/live-examples/wat-examples/statements/return_call.js
@@ -1,4 +1,5 @@
 const url = '{%wasm-url%}';
 const { instance } = await WebAssembly.instantiateStreaming(fetch(url));
 const result = instance.exports.fac(5n);
+
 console.log(result);
--- a/live-examples/wat-examples/statements/return_call.js
+++ b/live-examples/wat-examples/statements/return_call.js
@@ -1,11 +1,4 @@
 const url = '{%wasm-url%}';
-await WebAssembly.instantiateStreaming(fetch(url), {
-  env: {
-    fac: function (x) {
-      if (x === 0) {
-        return 1;
-      }
-      return x * this.fac(x - 1);
-    },
-  },
-});
+const { instance } = await WebAssembly.instantiateStreaming(fetch(url));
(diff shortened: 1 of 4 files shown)


pepelsbey:
@eqrion could you please have a look?

live-examples/wat-examples/statements/return_call.wat
```diff
@@ -0,0 +1,17 @@
+(module
+  (func $fac (export "fac") (param $x i64) (result i64)
+    (return_call $fac-aux (local.get $x) (i64.const 1))
+  )
+
+  (func $fac-aux (param $x i64) (param $r i64) (result i64)
+    (if (i64.eqz (local.get $x))
```
yurydelendik: There is a mistake in the example. Replace this line with
```suggestion
    (if (result i64) (i64.eqz (local.get $x))
```

live-examples/wat-examples/statements/return_call.js
```diff
@@ -0,0 +1,11 @@
...
+await WebAssembly.instantiateStreaming(fetch(url), {
+  env: {
+    fac: function (x) {
+      if (x === 0) {
+        return 1;
+      }
+      return x * this.fac(x - 1);
+    },
+  },
+});
```
yurydelendik:
```suggestion
const { instance } = await WebAssembly.instantiateStreaming(fetch(url));
const result = instance.exports.fac(5n);
console.log(result);
```

pepelsbey:
@yurydelendik thank you! However, the code still won’t run if pasted into the playground we have [on the call page](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Control_flow/call), for example:

```
> Error: parseWat failed:
3:6: error: opcode not allowed: return_call
    (return_call $fac-aux (local.get $x) (i64.const 1))
     ^^^^^^^^^^^
10:10: error: opcode not allowed: return_call
        (return_call $fac-aux
         ^^^^^^^^^^^
```

Do you have an idea if it’s the problem with the playground?

yurydelendik:
> Do you have an idea if it’s the problem with the playground?

I have a little insight into how playground works. The WebAssembly engine does not know about WAT and text representation. The transformation to bytecode is happening somewhere else. That something perhaps needs to be updated with new wasm specifications support.

yurydelendik:
If not mistaken, `parseWat` is API function from wabt JS library. Is https://github.com/mdn/bob/blob/main/editor/js/editable-wat.ts executed on the main page?

There is a reference to `"wabt": "^1.0.29"`, which is https://github.com/AssemblyScript/wabt.js . The package is not updated for long time and probably missing new specifications.

pepelsbey:
> The package is not updated for long time and probably missing new specifications.

Thank you for investigating! I’ll reach out to platform engineers about this.

fiji-flo:
once https://github.com/mdn/interactive-examples/pull/2845 landed we can get this out.

yurydelendik (commented):
Looks good, approved. (I don't have "Approve" permission)

Rumyra (approved):

PR #2696, opened 2023-12-19
fix(js): ignore formatting empty statement example with code comment

yg-i:
moved the comment to a new position which (i) makes more sense and (ii) is where the identical comment is located in the other code snippet on the same MDN page (see: https://github.com/mdn/content/blob/63a26b042a4d27c1038060c4ea8b605e188a5787/files/en-us/web/javascript/reference/statements/empty/index.md?plain=1#L41C17-L41C17 )

image

image

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 CHANGELOG.md
 docs/
 live-examples/html-examples/inline-text-semantics/rtc.html
+live-examples/js-examples/statement/statement-empty.js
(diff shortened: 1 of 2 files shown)


bsmth:
Thanks @yg-i - the linter doesn't like this. Maybe we can manage it with an ignore rule? We already have one: https://github.com/mdn/interactive-examples/blob/main/.prettierignore#L3

bsmth (approved):
Thank you

PR #2709, opened 2024-01-10
respect example domains in RFC 6761

Jxck:
### Description

Replace some real domains in documentation with example domains.

### Motivation

Respect "example domain" defined in [RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761).

### Additional details

https://github.com/orgs/mdn/discussions/450

#2650


estelle (commented):
Thank you for suggesting edits for these live examples. While it makes sense to use "example.com" in syntax explanations, these are examples, not syntax boxes. Thank you for the suggestion, but the originals are better suited for the purpose.

PR #2710, opened 2024-01-10
fix(css): update 'display' interactive demo

### LINKED ISSUE #31437: Wrong Visulation of Display Property
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/display

First section - Try it! 

In the display section, one of the properties is Display: block;

The corresponding visual effect is not what it should be. 
Instead of being a display block, the boxes are being displayed inline. 
Screenshot 2024-01-01 at 12 27 38 AM

The inner boxes- one, two, three- should be displayed top to bottom, not left to right!

_No response_

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/css/display`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/display
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/display/index.md
* Last commit: https://github.com/mdn/content/commit/b9db4e51b6f1cddba3af708643fc9804849d61c2
* Document last modified: 2023-12-26T21:19:48.000Z

</details>

OnkarRuikar:
- fixes https://github.com/mdn/content/issues/31437

### DIFF
--- a/live-examples/css-examples/basic-box-model/display.css
+++ b/live-examples/css-examples/basic-box-model/display.css
@@ -3,8 +3,12 @@
   height: 100%;
 }
 
+code {
+  background: #8888;
+}
+
 #example-element {
-  border: 3px solid #00f;
+  border: 3px solid orange;
 }
 
 .child {
--- a/live-examples/css-examples/basic-box-model/display.css
+++ b/live-examples/css-examples/basic-box-model/display.css
@@ -8,7 +8,7 @@ code {
 }
 
 #example-element {
-  border: 3px solid orange;
+  border: 3px dashed orange;
 }
 
 .child {
--- a/live-examples/css-examples/basic-box-model/display.html
+++ b/live-examples/css-examples/basic-box-model/display.html
@@ -38,13 +38,13 @@
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
     <div class="example-container">
-      Change <code>display</code> property of this
+      Some text A.
       <div id="example-element">
         <div class="child">Child 1</div>
         <div class="child">Child 2</div>
         <div class="child">Child 3</div>
       </div>
-      <code>div</code> element.
+      Some text B.
     </div>
   </section>
 </div>
--- a/live-examples/css-examples/basic-box-model/display.html
+++ b/live-examples/css-examples/basic-box-model/display.html
@@ -36,6 +36,10 @@
 </section>
 
 <div id="output" class="output large hidden">
+  <p>
+    Apply different <code>display</code> values on the dashed orange-bordered <code>div</code>, which contains three
+    child elements.
+  </p>
   <section id="default-example" class="default-example">
     <div class="example-container">
       Some text A.
--- a/live-examples/css-examples/basic-box-model/display.html
+++ b/live-examples/css-examples/basic-box-model/display.html
@@ -38,11 +38,13 @@
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
     <div class="example-container">
+      Change <code>display</code> property of this
       <div id="example-element">
-        <div class="child">One</div>
-        <div class="child">Two</div>
-        <div class="child">Three</div>
+        <div class="child">Child 1</div>
+        <div class="child">Child 2</div>
+        <div class="child">Child 3</div>
       </div>
+      <code>div</code> element.
     </div>
   </section>
 </div>


dipikabh:
This is close. But I think the additional text in the example does not fully clarify the element on which `display` is being set, especially considering that the OP in the [issue](https://github.com/mdn/content/issues/31437) expected each child element to have its own `display` setting.

Brainstorming an idea here. How about we add a description outside the `div` to explain the example, something like "The orange element represents the `div` on which different `display` values are set. It contains three child elements." This can be followed by:

```
    <div class="example-container">
      some text A
      <div id="example-element">
        <div class="child">Child 1</div>
        <div class="child">Child 2</div>
        <div class="child">Child 3</div>
      </div>
      some text B
    </div>
```

With `display: none`, the current text in the proposed PR reads: "Change `display` property of this `div` element.". It refers to the `div` element that gets hidden when `display: none`. Instead of using a complete sentence, using placeholders like "some text A" and "some text B" could demonstrate the absence of the `div` more effectively. With `display: none`, the output would read "some text A some text B"", indicating that the `div` is not displayed.

Let me know what you think.

live-examples/css-examples/basic-box-model/display.html
```diff
@@ -38,11 +38,13 @@
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
     <div class="example-container">
+      Change <code>display</code> property of this
```
OnkarRuikar:
```suggestion
      Some text A.
```

live-examples/css-examples/basic-box-model/display.html
```diff
@@ -38,11 +38,13 @@
...
       <div id="example-element">
-        <div class="child">One</div>
-        <div class="child">Two</div>
-        <div class="child">Three</div>
+        <div class="child">Child 1</div>
+        <div class="child">Child 2</div>
+        <div class="child">Child 3</div>
       </div>
+      <code>div</code> element.
```
OnkarRuikar:
```suggestion
      Some text B.
```

OnkarRuikar:
> How about we add a description outside the `div` to explain the example, something like "The orange element represents the `div` on which different `display` values are set. It contains three child elements." This can be followed by:

@dipikabh  Yes, we need to be more explicit.

live-examples/css-examples/basic-box-model/display.css
```diff
@@ -3,8 +3,12 @@
   height: 100%;
 }
 
+code {
+  background: #8888;
+}
+
 #example-element {
-  border: 3px solid #00f;
+  border: 3px solid orange;
```
estelle:
```suggestion
  border: 3px dashed orange;
```
this enables using more than color as a reference in the discussion in the content.

estelle (approved):
LGTM. @dipikabh I'll let you merge both PRs at the same time if this PR is ok with you.

live-examples/css-examples/basic-box-model/display.html
```diff
@@ -38,11 +38,13 @@
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
     <div class="example-container">
```
dipikabh: > Yes, we need to be more explicit.

@OnkarRuikar, should we add this then:

```suggestion
<div id="output" class="output large hidden">
  <p>The <code>display</code> property is set on the <code>div</code> element enclosed by the orange dashed border.
    You can apply different <code>display</code> values on this element to observe their effects on its layout
    in relation to adjacent elements.
    Note that this element contains three child <code>div</code> elements.
  <section id="default-example" class="default-example">
    <div class="example-container">
```



This is how it would look:

<kbd>display-inetractive</kbd>

dipikabh (changes requested):

live-examples/css-examples/basic-box-model/display.html
dipikabh: @OnkarRuikar I didn't realize you opened another PR to add the statement :)

dipikabh: Oh, that PR is in `mdn/content`.

I have not noticed other instances where we discuss or refer to the "Try it" demo examples in content.

It might be best to keep the interactive demo examples self-contained. Let me know what you think - we can close https://github.com/mdn/content/pull/31672 and instead make the explanatory updates (as suggested above) to the code within this repo.

OnkarRuikar: @dipikabh that's too much text in the output. On narrow screens, readers will have to scroll to see the actual target div.

> I have not noticed other instances where we discuss or refer to the "Try it" demo examples in content.

We do have some "Try it" sections with pros, e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/max#try_it

In the `mdn/content` PR we can put the text below the macro call. WDYT?

dipikabh: I am still trying to see if we can maintain this info in the example itself. How about using the text from your content PR:
<kbd>display-interactive-2</kbd>

or even a more shortened version:

<kbd>display-interactive-4</kbd>

Else the last option is to add text in content after the macro call.

OnkarRuikar: Let's pic the shortest version `Apply different <code>display</code> values on the dashed orange-bordered <code>div</code>, which contains three child elements.` https://github.com/mdn/interactive-examples/pull/2710/commits/bef7553c9c4915f23046f3ee6a8c08e4db00e90a

dipikabh (approved):
Thanks for the improvements, @OnkarRuikar!

OnkarRuikar:
@dipikabh you'll have to merge https://github.com/mdn/interactive-examples/pull/2716 as well to make it live.

dipikabh:
> you'll have to merge https://github.com/mdn/interactive-examples/pull/2716 as well to make it live.

Done 👍

wbamberg:
Thanks! FWIW I agree with @dipikabh that we should try very hard to keep interactive examples self-contained. And that really any kind of exposition in the examples is unfortunate, but sometimes as in this case it is unavoidable (as also for instance here: https://developer.mozilla.org/en-US/docs/Web/CSS/position). But I think what you have here is perfect :).

PR #2741, opened 2024-02-21
fix: `Intl.getCanonicalLocales` error message

btea:

### DIFF
--- a/live-examples/js-examples/intl/intl-getcanonicallocales.js
+++ b/live-examples/js-examples/intl/intl-getcanonicallocales.js
@@ -8,7 +8,7 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output: 
+  // Expected output:
   //   (Firefox/Safari) RangeError: invalid language tag: "EN_US"
   //   (Chrome) RangeError: Incorrect locale information provided
 }
--- a/live-examples/js-examples/intl/intl-getcanonicallocales.js
+++ b/live-examples/js-examples/intl/intl-getcanonicallocales.js
@@ -8,5 +8,5 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output: RangeError: invalid language tag: EN_US
+  // Expected output: RangeError: Incorrect locale information provided
 }
--- a/live-examples/js-examples/intl/intl-getcanonicallocales.js
+++ b/live-examples/js-examples/intl/intl-getcanonicallocales.js
@@ -8,5 +8,7 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output: RangeError: Incorrect locale information provided
+  // Expected output: 
+  //   (Firefox/Safari) RangeError: invalid language tag: "EN_US"
+  //   (Chrome) RangeError: Incorrect locale information provided
 }
--- a/live-examples/js-examples/intl/intl-getcanonicallocales.js
+++ b/live-examples/js-examples/intl/intl-getcanonicallocales.js
@@ -8,7 +8,6 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output:
-  //   (Firefox/Safari) RangeError: invalid language tag: "EN_US"
-  //   (Chrome) RangeError: Incorrect locale information provided
+  // Expected output (Firefox/Safari): RangeError: invalid language tag: "EN_US"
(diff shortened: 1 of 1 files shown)


bsmth:
Thanks a lot for raising this one. I'm going to close, however, because I get the following:

image

The error message here is browser-specific so it's safe enough to leave as-is.

Thanks!

btea:
@bsmth  Is it because of the browser difference? The error message I get is as follows: 

image

The browser version I am using is Edge 121.0.2277.128.

bsmth:
Yes, I can confirm [if you try it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/getCanonicalLocales) with Chrome, you get the following:

```
> Array ["en-US"]
> Array ["en-US", "fr"]
> "RangeError: Incorrect locale information provided"
```

And Safari:

```
> Array ["en-US"]
> Array ["en-US", "fr"]
> "RangeError: invalid language tag: EN_US"
```

btea:
The error messages of Firefox and Safari are consistent, and the output messages of Chrome and node are consistent. Maybe add both error messages and then mark the running environment?

bsmth:
We can add both to cover Chrome, Fx and Safari, why not. This can also be verified [using eshost](https://github.com/bterlson/eshost-cli)

```
'Intl.getCanonicalLocales("EN_US")'
┌──────────────┬───────────────────────────────────────────────────┐
│ SpiderMonkey │                                                   │
│              │ RangeError: invalid language tag: "EN_US"         │
├──────────────┼───────────────────────────────────────────────────┤
│ V8           │                                                   │
│              │ RangeError: Incorrect locale information provided │
└──────────────┴───────────────────────────────────────────────────┘
```

> mark the running environment

The running environment is the reader's browser in this case, so I don't think we need to do anything else there.

live-examples/js-examples/intl/intl-getcanonicallocales.js
```diff
@@ -8,5 +8,5 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output: RangeError: invalid language tag: EN_US
+  // Expected output: RangeError: Incorrect locale information provided
```
bsmth:
```suggestion
  // Expected output: 
  //   (Firefox/Safari) RangeError: invalid language tag: "EN_US"
  //   (Chrome) RangeError: Incorrect locale information provided
```

bsmth (approved):
Thank you!

Josh-Cena:
Do we do this elsewhere?

I don't care much for these examples but I also don't think it's helpful when the message has the same meaning anyway.

bsmth:
> Do we do this elsewhere?

We do have a few:

```
live-examples/js-examples/intl/intl-listformat-prototype-resolvedoptions.js:// Expected output (Chrome): "de"
live-examples/js-examples/intl/intl-numberformat-prototype-resolvedoptions.js:// Expected output (Chrome): "de"
live-examples/js-examples/weakset/weakset-prototype-add.js:  // Expected output (Chrome): TypeError: Invalid value used in weak set
live-examples/js-examples/symbol/symbol-match.js:// Expected output (Chrome): Error: First argument to String.prototype.startsWith must not be a regular expression
```

And actually I just spotted we have this in the style guide: https://github.com/mdn/interactive-examples/blob/main/contributing/javascript-style-guide.md#representing-browser-differences
 
> I also don't think it's helpful when the message has the same meaning anyway

I tend to agree somewhat, my initial decision was to close for this reason, but I can see why some readers might be confused. I would format it like the suggestion in the style guide and merge, unless you have objections, @Josh-Cena.

Josh-Cena:
That "browser difference" is actual runtime difference, but this one is just a representational difference. I don't have strong objections, though.

live-examples/js-examples/intl/intl-getcanonicallocales.js
```diff
@@ -8,5 +8,7 @@ try {
   Intl.getCanonicalLocales('EN_US');
 } catch (err) {
   console.log(err.toString());
-  // Expected output: RangeError: invalid language tag: EN_US
+  // Expected output:
+  //   (Firefox/Safari) RangeError: invalid language tag: "EN_US"
+  //   (Chrome) RangeError: Incorrect locale information provided
```
bsmth:
```suggestion
  // Expected output (Firefox/Safari): RangeError: invalid language tag: "EN_US"
  // Expected output (Chrome): RangeError: Incorrect locale information provided
```

bsmth (approved):
Thank you!

bsmth:
> That "browser difference" is actual runtime difference, but this one is just a representational difference. I don't have strong objections, though.

You're right. Let's go with this for now 👍🏻

PR #2745, opened 2024-03-03
Add example of `text-wrap`

gldaniil:
### Description

This PR adds an example of [text-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap).

### Motivation

There is not an existing example.

### Additional details

Image

### DIFF
--- a/live-examples/css-examples/text/text-wrap.html
+++ b/live-examples/css-examples/text/text-wrap.html
@@ -37,8 +37,13 @@
 
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
-    <div id="example-element" class="transition-all">
-      <p>Lorem ipsum dolor sit amet consectetur adipisicing elit. Voluptatem aut cum eum id quos est.</p>
+    <div class="whole-content-wrapper">
+      <p>Click here to edit the text:</p>
+      <div id="example-element" class="transition-all">
+        <p contenteditable>
+          Lorem ipsum dolor sit amet consectetur adipisicing elit. Voluptatem aut cum eum id quos est.
+        </p>
+      </div>
     </div>
   </section>
 </div>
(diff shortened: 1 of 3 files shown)


live-examples/css-examples/text/text-wrap.html
```diff
@@ -0,0 +1,44 @@
...
+    <button type="button" class="copy hidden" aria-hidden="true">
+      <span class="visually-hidden">Copy to Clipboard</span>
+    </button>
+  </div>
+</section>
+
+<div id="output" class="output large hidden">
+  <section id="default-example" class="default-example">
+    <div id="example-element" class="transition-all">
+      <p>Lorem ipsum dolor sit amet consectetur adipisicing elit. Voluptatem aut cum eum id quos est.</p>
```
chrisdavidmills: I think it would make sense to add `contenteditable` to the paragraph, and perhaps also add a little note to tell readers to experiment with editing the text. I say this because some values of `text-wrap` (most notably `stable`) only have an effect when editing user input.

chrisdavidmills (changes requested):
Hi there @gldaniil, and thank you so much for contributing to MDN! The example setup looks good. In terms of the content, I have one comment for you to consider. Let me know what you think.

Thanks again!

live-examples/css-examples/text/text-wrap.html
gldaniil: Hi @chrisdavidmills ! Thanks for the suggestions, I agree with you.
Also, I added a wrapper `<div>` for a better look:
Image

chrisdavidmills: Look great, thanks @gldaniil.

I do wonder whether "Click here..." might confuse people into thinking they have to click the text above the box rather than the box itself. How about "Edit the text in the box" or similar?

gldaniil: Yeah, you're right, I considered different versions of this phrase to make it short and informative, but I think your choice looks more appropriate. Thanks @chrisdavidmills!

chrisdavidmills (approved):
Perfect, that looks great — thanks @gldaniil! I look forward to seeing this live on MDN.

PR #2752, opened 2024-03-21
fix incorrect method parameter '...arguments' with actual parameters (target, prop, receiver)

meck93:
### Description

The [Proxy documentation page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) currently contains an example which is most likely outdated. The parameter `..arguments` no longer exists. At least, when I used TypeScript it complained that the arguments did not exist. On the other hand, the example itself in the browser still works.

Relates to https://github.com/mdn/content/pull/32790


Josh-Cena (commented):
See the content issue

PR #2759, opened 2024-04-12
WAT memory examples - fix await

hamishwillee:
This fixes two WASM examples for working for memory.

Specifically these are JavaScript files that use `await` to wait on a returned promise, but then chain off the result using `then` - there is no point doing both, so this removes the "then" and makes it await properly.

There are probably similar examples, but these are the ones I'm looking at for https://github.com/mdn/content/issues/32777

### DIFF
--- a/live-examples/wat-examples/memory/load.js
+++ b/live-examples/wat-examples/memory/load.js
@@ -1,12 +1,11 @@
 const url = '{%wasm-url%}';
-await WebAssembly.instantiateStreaming(fetch(url)).then((result) => {
-  const load_first_item_in_mem = result.instance.exports.load_first_item_in_mem;
-  const memory = result.instance.exports.memory;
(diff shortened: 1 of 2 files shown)


dipikabh:
Hi @bsmth, I'm not as familiar with WASM memory. Since you've recently reviewed similar content and examples, could you help by taking a look at this PR as well? Thanks!

hamishwillee:
This is actually a JavaScript async issue really. It makes no sense to await on a promise when you're going to handle the result in a `then`. Either you await and handle the result in the next line, or you don't await and you handle in a `then` call. Not both.

dipikabh (approved):
LGTM, makes sense to remove the redundancy (or rather to make the code cleaner). Thanks!

Maykon153:
Vcs vão me ajudar ?

PR #2761, opened 2024-04-19
css change

zhangenming:
### Description
image


estelle (approved):
Thank you. The original did NOT look good!. Thanks for catching that and making the suggested change.

I added an outline to the example element so the difference wasn't only the background color so the only indicator of which element was important isn't just color.

Congratulations on your first merged MDN content PR. Welcome to the team!

PR #2787, opened 2024-06-09
fix(css): add cue pseudo element with argument case

OnkarRuikar:
- part of fix for https://github.com/mdn/content/issues/23209


bsmth (approved):
Nice, thank you. Tested in Chrome `125.0.6422.113` and works as expected.

PR #2794, opened 2024-06-18
style example - add second css selector, bolden keyword in example

wesinator:
### Description

This PR adds a second CSS selector to the style example, which boldens the style keyword in the example HTML styling.

### Motivation

This serves two purposes: on the HTML code side of the example/demo, it gives an example of using multiple CSS selectors for different elements in a style element.
On the output side, it adds a little touch to bolden the keyword `style` in the description text (which is already expressed as a code element).
image

[NO_TRAIN]::

### DIFF
--- a/live-examples/html-examples/document-metadata/style.html
+++ b/live-examples/html-examples/document-metadata/style.html
@@ -2,6 +2,9 @@
   p {
     color: #26b72b;
   }
+  code {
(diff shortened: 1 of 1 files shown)


estelle (approved):
Thank you. Also, congratulations on your first merged MDN interactive examples PR. Welcome to the team!

PR #2796, opened 2024-06-22
Fix expected output in example for Logical OR assignment (||=) - expressions-logical-or-assignment.js

artemshar:
Missing dot in expected output

### Description

So as not to frustrate the reader, this operator is not a variation of the trim method.
When I first read about this method and this example confused me - “why did the dot disappear?”, but RUN put everything in its place.

### Motivation

Perhaps these seconds of frustration will save thousands of developer hours. ;)

### Additional details

Perhaps this will be my most significant contribution to open source in my entire life! (one dot in commit)

☺️


Josh-Cena (approved):
Thanks

PR #2806, opened 2024-07-12
move citation out of blockquote per spec

### LINKED ISSUE #28401: Example given on blockquote page is incorrect according to the HTML spec
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote

Try it

The example given places the attribution for the quotation inside `blockquote` element, which incorrect according to the HTML specification.

The example should place the quotation attribution outside the `blockquote` element, eg. in `figcaption` element:

```
<figure>
  <blockquote cite="https://www.huxley.net/bnw/four.html">
    <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
  </blockquote>
  <figcaption>—Aldous Huxley, <cite>Brave New World</cite></figcaption>
</figure>
```

The HTML specification (https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element) states: “Attribution for the quotation, if any, must be placed outside the `blockquote` element.” 

_No response_

### MDN metadata

<details>

* Folder: `en-us/web/html/element/blockquote`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/html/element/blockquote/index.md
* Last commit: https://github.com/mdn/content/commit/e04d8d2766c468f149445c0bf438d09f9b2d188c
* Document last modified: 2023-06-13T08:10:47.000Z

</details>

estelle:
fixes https://github.com/mdn/content/issues/28401

### DIFF
--- a/live-examples/html-examples/text-content/css/blockquote.css
+++ b/live-examples/html-examples/text-content/css/blockquote.css
@@ -1,10 +1,7 @@
-blockquote {
-  margin: 0;
-}
-
-blockquote p {
+div:has(> blockquote) {
+  background-color: ##ededed;
+  margin: 10px auto;
   padding: 15px;
-  background: #eee;
   border-radius: 5px;
 }
 
@@ -14,4 +11,8 @@ blockquote p::before {
 
 blockquote p::after {
   content: '\201D';
+}
+
+blockquote + p {
+  text-align: right;
 }
--- a/live-examples/html-examples/text-content/css/blockquote.css
+++ b/live-examples/html-examples/text-content/css/blockquote.css
@@ -1,5 +1,5 @@
 div:has(> blockquote) {
-  background-color: ##ededed;
+  background-color: #ededed;
   margin: 10px auto;
   padding: 15px;
   border-radius: 5px;
--- a/live-examples/html-examples/text-content/blockquote.html
+++ b/live-examples/html-examples/text-content/blockquote.html
@@ -1,4 +1,6 @@
-<blockquote cite="https://www.huxley.net/bnw/four.html">
-  <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
-</blockquote>
-<p>—Aldous Huxley, <cite>Brave New World</cite></p>
+<div>
+  <blockquote cite="https://www.huxley.net/bnw/four.html">
+    <p>Words can be like X-rays, if you use them properly—they’ll go through anything. You read and you’re pierced.</p>
+  </blockquote>
+  <p>—Aldous Huxley, <cite>Brave New World</cite></p>
+</div>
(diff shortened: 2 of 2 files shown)


Josh-Cena:
This significantly changes the layout of the quote and I'm not sure if anyone would be doing this. Usually, when rendering a quote, you would expect the attribution to be styled as if it's part of the quote (gray color, left vertical line, right aligned, etc.), and not as if it's any other paragraph. This was why, before https://github.com/mdn/interactive-examples/pull/2562, it was wrapped in another figure, so styles can be applied to the `figure`. `figure`/`figcaption` is also the spec-recommended structure when writing attribution. However based on https://github.com/mdn/content/issues/28401#issuecomment-1666396747, it seems this actually degrades screen reader performance? Maybe we should get the spec authors to clarify?

estelle:
> However based on [mdn/content#28401 (comment)](https://github.com/mdn/content/issues/28401#issuecomment-1666396747), it seems this actually degrades screen reader performance? 

That's why i didn't use figcaption.

> changes the layout of the quote

I updated the CSS and HTML to make it look like a blockquote without adding any semantics for a screenreader to find.

Josh-Cena:
OK, I'm +1 on the current form, but do we want to clarify in content that although the spec says to use `figure`, it's actually bad for some screen readers?

live-examples/html-examples/text-content/css/blockquote.css
```diff
@@ -1,10 +1,7 @@
-blockquote {
-  margin: 0;
-}
-
-blockquote p {
+div:has(> blockquote) {
+  background-color: ##ededed;
```
Josh-Cena:
```suggestion
  background-color: #ededed;
```

estelle:
> OK, I'm +1 on the current form, but do we want to clarify in content that although the spec says to use `figure`, it's actually bad for some screen readers?

Not in the interactive example, as I assume most people don't actually read the content in the example.

Josh-Cena:
@sideshowbarker Since you approved https://github.com/mdn/content/pull/34813, could you take a quick look at this too? (You are probably also more knowledgeable about HTML than me.)

sideshowbarker:
> _@sideshowbarker Since you approved [mdn/content#34813](https://github.com/mdn/content/pull/34813), could you take a quick look at this too? (You are probably also more knowledgeable about HTML than me.)_
>…
> _Maybe we should get the spec authors to clarify?_

1. Given that at https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element:the-blockquote-element-4 the spec text itself is unambiguous:

   > _Attribution for the quotation, if any, must be placed outside the [blockquote](https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element) element._

2. Over many years of discussions about the spec that I’ve followed (going back 17 years or more), I’ve seen this debated many times — with some people expressing very strong disagreement with it (including some prominent “standardista” influencers) — but the resolution has nevertheless always been to keep that attribution-must-be-placed-outside-the-blockquote requirement in there

…then I’m certain the answer from the spec authors would just be, _“That’s what the spec says”_. I think they’d probably even not suggest that anyone open a new issue if they disagree — because there have already been issues for it, and they’ve been resolved, and it doesn’t seem like there’s any new information which merits re-visiting those.

So I think as far as MDN goes, we have an obligation to conform to the spec requirements, and I’ll go ahead and merge this.

sideshowbarker (approved):

PR #2808, opened 2024-07-15
feat: Remove `@@notation` from user-facing example titles

Josh-Cena:
Part of https://github.com/mdn/mdn/issues/562. This PR only changes user-facing strings, without touching any file paths.

### DIFF
--- a/live-examples/js-examples/intl/intl-prototype-@@tostringtag.js
+++ b/live-examples/js-examples/intl/intl-prototype-@@tostringtag.js
@@ -1,4 +1,0 @@
-const intl1 = Object.prototype.toString.call(Intl);
-
-console.log(intl1);
-// Expected output: "[object Intl]"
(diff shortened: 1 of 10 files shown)


Josh-Cena:
I don't really know who to request for reviews, so I'm just randomly fetching @bsmth 😄

live-examples/js-examples/intl/intl-prototype-@@tostringtag.js
bsmth, line 1: do we reference this in content? Same as the other deletion below

Josh-Cena: No, we don't. We removed these two pages long ago, which is why I'm simply removing the examples.

bsmth (approved):

bsmth:
LGTM, thank you!

PR #2832, opened 2024-08-31
Demonstrate wrapping with asIntN/asUintN Bigint static methods

lionel-rowe:
### Description

Demonstrate wrapping with asIntN/asUintN Bigint static methods

### Motivation

Current examples fail to demonstrate the main use case for `asIntN` and `asUintN`, namely wrapping. The `number > max` check in the function means the number will never wrap, so there's no point in calling `asIntN`/`asUintN`. Also, the check is buggy, given that it only checks the max and not the min (`-1n` doesn't fit in a u64, and `-9223372036854775809n` doesn't fit in an i64, yet neither would trigger the "Number doesn't fit" path).

### Additional details

N/A

N/A

### DIFF
--- a/live-examples/js-examples/bigint/bigint-asintn.js
+++ b/live-examples/js-examples/bigint/bigint-asintn.js
@@ -1,12 +1,12 @@
-const max = 2n ** (64n - 1n) - 1n;
+const I64_CEIL = 2n ** 63n;
 
-function check64bit(number) {
-  number > max
-    ? console.log("Number doesn't fit in signed 64-bit integer!")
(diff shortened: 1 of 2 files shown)


chrisdavidmills (approved):
Hi @lionel-rowe, and thanks for your contribution to MDN.

I've tested the examples — they work as advertised, and are the right size to fit into the live example widget.

I'd also like to get a comment on this from someone who knows more than me about JS language mechanics, like @Josh-Cena.

Josh-Cena (approved):
Yeah, this makes sense to me. Thanks

PR #2836, opened 2024-09-06
Replace sample text for unicode-bidi example

coolreader18:
### Description

Replaces the sample hebrew text used in the unicode-bidi example with a less meaningful line.

### Motivation

The current text is a line from a [Jewish prayer](https://en.wikipedia.org/wiki/Shema#Baruch_Shem), and specifically a line that is usually said more quietly or to oneself - it startled me a bit to see it used as sample text. I replaced it with the first line from the Torah, "In the beginning God created the heavens and the earth", which is about as well-known as a bit of hebrew text can get.

### DIFF
--- a/live-examples/css-examples/writing-modes/unicode-bidi.html
+++ b/live-examples/css-examples/writing-modes/unicode-bidi.html
@@ -30,6 +30,6 @@
 
 <div id="output" class="output large hidden">
   <section id="default-example" class="default-example">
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
Hi @coolreader18! Thanks for this contribution to MDN. I really appreciate your thoughts here on respectful/appropriate use of language.

PR #2838, opened 2024-09-13
Use kbd to represent user input

dragunovartem99:
### Description

1. Changed the HTML element
2. Added CSS to achieve same visual effect that was created by `strong + u` combination

### Motivation

I think it's more appropriate, and may motivate readers to investigate into `<kbd>` semantics. Maybe in future this element can be included in corresponding page's "See also"

### Additional details

https://html.spec.whatwg.org/dev/text-level-semantics.html#the-kbd-element
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

### DIFF
--- a/live-examples/html-examples/global-attributes/attribute-accesskey.html
+++ b/live-examples/html-examples/global-attributes/attribute-accesskey.html
@@ -1,6 +1,5 @@
 <p>
-  If you need to relax, press the <strong><u>S</u></strong
-  >tress reliever!
+  If you need to relax, press the <kbd>S</kbd>tress reliever!
 </p>
 
 <button accesskey="s">Stress reliever</button>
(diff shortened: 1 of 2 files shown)


estelle:
Thank you for your contribution to MDN. 

I don't think using `<kbd>` for the `accesskey` character is a "best practice". If you use voice over to read the edit, it reads "press the ess tress reliever", but it does that for other elements as well. This page should likely be updated, but not with KBD

dragunovartem99:
@estelle thank you for the feedback, it was useful, as I never considered this a11y side so seriously. Probably I should learn about screen readers!

Maybe I can change it to single <b> element, preserving the styling?

According to [WHATWG](https://html.spec.whatwg.org/dev/text-level-semantics.html#the-b-element): The b element represents a span of text to which attention is being drawn for utilitarian purposes without conveying any extra importance

If you will be satisfied with this solution, I can add another commit or create new PR

estelle:
a `<b>` would work!

PR #2840, opened 2024-09-16
fix: improve color contrast in menu element example for accessibility

### LINKED ISSUE #2839: Inaccessible color contrast ratio in menu element example

The current color scheme of the [menu element example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#try_it) does not meet accessible contrast standards, making it difficult to read.

### Safari
(EDIT: as mentioned by @estelle, `color: buttontext;` causes the color of button almost the same as its background resulting in a poor color contrast ratio of **1.2**)

Screenshot 2024-09-16 at 11 11 48 AM
### Chrome
Screenshot 2024-09-16 at 11 12 12 AM
### Firefox
Screenshot 2024-09-16 at 11 12 44 AM
## Contrast ratio
Screenshot 2024-09-16 at 11 13 59 AM

I expected the example to use colors that meet the Web Content Accessibility Guidelines (WCAG) 2.1, specifically the contrast ratio of at least 4.5:1 for normal text.
  
This can be fixed using `background-color: black` and we can remove the black border (if we want).
This fixes both the color between button text and it's background and anchor (link) text and the background.
   
Screenshot 2024-09-16 at 11 17 01 AM
Screenshot 2024-09-16 at 11 17 37 AM

You can verify the contrast issue using accessibility testing tools like [Contrast Checker](https://webaim.org/resources/contrastchecker/) or by using browser developer tools to inspect the element's styles.

The [menu element example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#try_it) can be found here.

The issue is reproducible on macOS Monterey using Safari 17.5, Chrome 128.0.6613.138, and Firefox 130.0. It is not browser-specific, as the contrast issue appears consistently across all of them.

yashrajbharti:
### Description

I changed the background color of the `menu` element example from bisque to black to fix the inaccessible color contrast issue, improving readability and meeting WCAG 2.1 contrast ratio standards.

### Motivation

The previous background color did not meet the minimum contrast ratio requirement, making the text difficult to read for users with visual impairments (and even hard for normal users). This change improves accessibility across browsers.

### Additional details

- [MDN menu element example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#try_it)  
- The issue is reproducible on macOS Monterey using Safari 17.5, Chrome 128.0.6613.138, and Firefox 130.0.

Fixes #2839

### DIFF
--- a/live-examples/html-examples/interactive-elements/css/menu.css
+++ b/live-examples/html-examples/interactive-elements/css/menu.css
@@ -1,5 +1,5 @@
 .news {
-  background-color: bisque;
+  background-color: black;
   padding: 1em;
   border: solid thin black;
 }
--- a/live-examples/html-examples/interactive-elements/css/menu.css
+++ b/live-examples/html-examples/interactive-elements/css/menu.css
@@ -1,7 +1,11 @@
 .news {
-  background-color: black;
+  background-color: bisque;
   padding: 1em;
   border: solid thin black;
+
+  @media (prefers-color-scheme: dark) {
+    background-color: black;
+  }
 }
 
 menu {
@@ -10,4 +14,9 @@ menu {
   padding: 0;
   margin-bottom: 0;
   gap: 1em;
+}
+
+button {
+  color: black;
+  background: white;
 }


live-examples/html-examples/interactive-elements/css/menu.css
```diff
@@ -1,5 +1,5 @@
 .news {
-  background-color: bisque;
+  background-color: black;
```
estelle: this makes it more accessible in dark mode, but makes the link above impossible to read in light mode.
Screenshot 2024-09-16 at 3 04 04 AM

yashrajbharti: Okay, fixing it for both light and dark modes and checking in all the major browsers

yashrajbharti: Making the following changes fixes the color contrast issue as checked in all major browsers:

(But I couldn't find a way without using a `@media` query for color schemes, and I didn't wanted to re-color the states of anchor tag, which by default comes with accessibility out of the box.)

```css
.news {
  background-color: bisque;
  padding: 1em;
  border: solid thin black;

  @media (prefers-color-scheme: dark) {
    background-color: black;
  }
}

menu {
  list-style-type: none;
  display: flex;
  padding: 0;
  margin-bottom: 0;
  gap: 1em;
}

button {
  color: black;
  background: white;
}
```

Screenshot 2024-09-16 at 6 14 23 PM
Screenshot 2024-09-16 at 6 14 29 PM
Screenshot 2024-09-16 at 6 16 10 PM
Screenshot 2024-09-16 at 6 15 37 PM

@estelle Should I go for it?

yashrajbharti:
The latest commit fixes the color contrast issues that occurred in Safari due to its user agent style sheet setting the color `buttontext`. This can also be fixed using `color: unset`. 
The commit also cares about the light and dark modes where color contrast are now better for readability (greater than 5.6 in all cases). 

Added image below:
Light theme:
Screenshot 2024-09-18 at 7 39 51 PM
Dark theme:
Screenshot 2024-09-18 at 7 39 43 PM
Safari (17.5)
Screenshot 2024-09-18 at 7 40 33 PM

estelle:
I added a comment to https://github.com/mdn/interactive-examples/issues/2839#issuecomment-2357346985 which I thought I had added here. I am hesitant to merge anything targeting the issue with the Safari button as this should be a webkit fix.

yashrajbharti:
There is something more to it, the same code gives different results if pasted from interactive example in the docs to the mdn playground, thus I fail to reproduce the bug on MDN playground. However, I have added a report to [webkit bugzilla](https://bugs.webkit.org/show_bug.cgi?id=282132) with a link to this a11y button issue #2839 , and `<menu>` element example from [mdn web docs | menu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#try_it). 

Screenshot 2024-10-26 at 2 58 46 PM
Screenshot 2024-10-26 at 2 58 41 PM

estelle:
Thanks @yashrajbharti. Let's close this rather than commit it, as it is better to keep the examples as basic as possible (the fewest features needed).

PR #2842, opened 2024-09-17
Replace two HTML tags with one (with better semantics)

dragunovartem99:
## Description

- Simplified the HTML structure
- Added CSS to achieve same visual effect that was created by strong + u combination

Motivation

#2838 — agreed with @estelle that we can change this example

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey


estelle (approved):
Thank you. Also, congratulations on your first merged MDN interactive examples PR. Welcome to the team!

PR #2844, opened 2024-09-22
Remove unnecessary CSS property

dragunovartem99:
### Description

Removes useless style property

### Motivation

In addition to #2842

For some reason I've included `font-weight: bold` for the `<b>` element

It was unintentional


estelle (approved):
I was the one who added the bold style. I added it to get the linter to work, and adding it made it pass linting, so it seemed like the linter required it. This PR removed that, and is approved by the linter, so I am not sure what i did in the past PR to make the linter happy; but glad it's happy without the bold! Thank you.

PR #2848, opened 2024-10-09
CSS live examples: color contrast improvements

ferdnyc:
### Description

For the demonstration elements with `background-color:` set and text content, also set explicit foreground color.

### Motivation

Prevents contrast issues (unreadable text inside the element), with the dark skin.

### Additional details

I noticed at https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-start#try_it that, with the dark site palette enabled, the yellow absolutely-positioned element appeared to contain no text — because it was rendered in white over yellow, due to the element's background-color being set while leaving the default text color.

I'd be pretty surprised if this is the only instance of color clashes due to incomplete styles, across the examples — it'd probably make sense to tackle them all at once, or at least as many as possible. So, feel free to consider this a WIP / starting-point PR. I'm willing to scan (ar at least grep) through the rest of the demo CSS files to look for more styles that need to be fixed in a similar manner, if it'd be a useful project to undertake.

### DIFF
--- a/live-examples/css-examples/container/content-visibility.css
+++ b/live-examples/css-examples/container/content-visibility.css
@@ -11,6 +11,7 @@
 .child {
   border: 3px solid rgb(64, 28, 163);
   background-color: wheat;
+  color: black;
   width: 80%;
   height: 80%;
   display: flex;
--- a/live-examples/css-examples/filter-effects/filter.css
+++ b/live-examples/css-examples/filter-effects/filter.css
@@ -1,0 +1,13 @@
+.example-container {
+  background-color: #fff;
+  width: 260px;
+  height: 260px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+#example-element {
+  flex: 1;
+  padding: 30px;
+}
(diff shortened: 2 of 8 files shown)


chrisdavidmills (approved):
Hi @ferdnyc, and thanks for your contribution to MDN — this is really useful!

I have approved it and could merge, but I'm holding off for now in case you want to add more fixes to this PR. Let me know if you want me to go ahead and merge it.

ferdnyc:
Thanks @chrisdavidmills — so far I've found three more `inset-{block-inline}-*` examples with the same color scheme (which I fixed by adding to the same commit; sorry for rewriting history, but it's a very small change and seemed cleaner), and one more at:

https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility

Also fixed (in a separate commit). I'll keep looking through the examples for more.

ferdnyc:
And https://developer.mozilla.org/en-US/docs/Web/CSS/filter#try_it needed a container element added to make the background white, so that the default drop-shadow filter is visible in dark mode.

chrisdavidmills:
Nice work @ferdnyc; these make perfect sense. 

Tell you what, I'm going to merge this now; I think it'd be good for you to open new PRs for any more that you find. Maybe do them in blocks of about 8-10 per PR, so that improvements are published quickly and each one is nicely manageable?

Thanks again for the super-helpful contribution.

PR #2853, opened 2024-10-18
chore(column-fill): remove balance-all example

caugner:
### Description

Removes an example for a feature that has not been implemented by any browser.

See:
- https://github.com/mdn/content/pull/34561
- https://github.com/mdn/browser-compat-data/pull/23578


estelle (approved):
Thanks! 🎉

PR #2854, opened 2024-10-18
Disambiguate `super` from `this` inheritance

davidmcguiredesign:
### Description

If the child class inherits a function, calling it from `super` behaves identically to calling it from `this`. But by overriding the inherited definition, the example can now demonstrate that `super` bypasses the child’s definition.

### Motivation

Understanding `super` is easier when contrasted with `this` than when defined in (comparative) isolation.

### DIFF
--- a/live-examples/js-examples/expressions/expressions-super.js
+++ b/live-examples/js-examples/expressions/expressions-super.js
@@ -14,6 +14,7 @@ class FooBar extends Foo {
     this.index = index;
   }
 
+  // Does not get called
   getNameSeparator() {
     return '/';
   }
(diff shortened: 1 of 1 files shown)


live-examples/js-examples/expressions/expressions-super.js
```diff
@@ -14,6 +14,10 @@ class FooBar extends Foo {
     this.index = index;
   }
 
+  getNameSeparator() {
```
Josh-Cena, line 18:
```suggestion
  // Does not get called
  getNameSeparator() {
```

Josh-Cena (approved):
This makes sense to me. Thanks!

davidmcguiredesign:
Thanks Josh! Sorry about the early “resolution”—I’m still new to open-source contribution mechanisms and etiquette.

Josh-Cena:
You accepted my suggestion and automatically marked the conversation as resolved—that's exactly how it should go :)

PR #2876, opened 2024-12-10
fix: Fix JS for CSS transform-origin example (#2875)

### LINKED ISSUE #2875: transform-origin example is broken
The examples in the interactive example at https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin are supposed to rotate, but they don't any more. I don't know which other interactive examples might also be broken in this way.

lionel-rowe:
### Description

From linked issue:

> The examples in the interactive example at https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin are supposed to rotate, but they don't any more.

### Motivation

Fix examples.

### Additional details

Also added clarifying comment to 4th example, which uses a different transform.

Fixes #2875.


chrisdavidmills (approved):
I have tested this locally, and it works fine. Thanks a lot for the fix @lionel-rowe, let's get this merged!

PR #2879, opened 2024-12-14
Made JS Expected output comments consistent with actual output

atg:
### Description

Fix a few `// Expected output: ...` comments in JS examples. In some cases these comments weren't consistent with what JS engines output, in other cases, the formatting was inconsistent with how the interactive example console formats things.

Note that for currency formatting (e.g. bigint-tolocalestring.js), JS engines produce a non-breaking space U+00A0 instead of an ASCII space U+0032.


Josh-Cena:
Our CI does not like irregular whitespace. And it's fine to use regular spaces here; the goal is not to compare the output character by character. Could you revert those space changes and possibly add some text like `(Note: the space may be a non-breaking space)`?

Josh-Cena:
Hi! These files are no longer used by MDN. If you are still interested in contributing, please raise the PR in mdn/content instead. (Sorry for not getting to it sooner; it fell between my cracks)

PR #2886, opened 2025-01-11
Add keyword values for `white-space` property

skyclouds2001:
### Description

these keyword values have got supported in firefox for a long time, also is getting supported in preview chromium, so it's ok to add thses values to the example


estelle (approved):
Thanks! 🎉

PR #2900, opened 2025-02-13
chore: Add warning for new PRs

bsmth:
There is migration work in progress to inline the example code into content. We should put up a warning to avoid that anyone loses time and work due to rejected / closed PRs.


pepelsbey (approved):
Looks good! Thank you :)

